Issue #413, opened 2014-12-31
Generate JSON data files programmatically


shawnbot:
The data for the map is currently stored in [.geojson files](https://github.com/18F/doi-extractives-data/tree/gh-pages/static/data) which are *actually* JavaScript scripts that declare a global variable for each GeoJSON FeatureCollection.

This is bad practice, and makes it difficult for other people/sites to access the data. I'm proposing the following changes:

1. Regenerate standalone JSON files containing *just the geometry* for each layer, e.g. from the original Shapefiles.
2. Create CSV files containing the extractives data for each feature.
3. Use [geojoin](https://github.com/shawnbot/py-geojoin) or "join" the data onto the feature collections, producing a separate set of JSON files.
3. To reduce load times, we could combine all of the layers into a single [TopoJSON](https://github.com/mbostock/topojson/wiki) file.

These steps can all be automated so that future data updates should simply be a matter of modifying the CSV files and running a script. Then we would load the file(s) via `d3.json()` at runtime, rather than using `<script>` tags.

meiqimichelle:
@shawnbot Is this issue still relevant?

shawnbot:
Nope, we're already doing it on the data branch :)

PR #414, opened 2014-12-31
map accessibility, code cleanup and optimization

shawnbot:
This refactor addresses several issues with the map:

1. All of the regions are now keyboard accessible via the `tabindex` attribute, and most have `aria-label` attributes. One particular bit of code reorganizes the SVG DOM that Leaflet generates so that MultiPolygon features fall under a single `<g>` element, rather than one per polygon (which resulted in three separately focusable regions for Michigan, for instance).
1. The state draw order has been updated so that western states draw last (though not all states have data, so the order may not be final)
1. The map layers have been turned into a configuration object, `layers`, which specifies the data set, sort order and optional name property. This will ease the transition to asynchronous loading of GeoJSON per #413.
1. The 3D effect has been refactored using d3 data joins, which should *in theory* be faster. Zooming hides then re-renders the shapes, but does not recalculate heights.
1. JavaScript has been reformatted and some functions renamed for clarity and consistency.

A couple of things need to be done before this should be merged:

- [x] incorporate an ES5 shim (either [es5-shim](https://github.com/es-shims/es5-shim) or [aight](https://github.com/shawnbot/aight)) for IE9 (which lacks `Array.prototype.sort()`, among other methods)
- [ ] test it in IE9 and IE10 
- [ ] test it on mobile devices (side note: the `navigator.userAgent` sniffing could be re-worked to use feature and/or [form factor](http://www.html5rocks.com/en/tutorials/detection/#ff-detection) detection, or a more robust tool like [WhichBrowser](https://github.com/WhichBrowser/WhichBrowser))


meiqimichelle:
Also, recording here for The Future, I think we're pretty close. The things I would tweak:
- remove 3D because it is too visually complex with all the colors, and doesn't add much (the idea is awesome, but in implementation it might be too much noise per signal) (I think we-18F are coming to this conclusion, but we will have to talk with our DOI partners on this)
- add star or something to the negative value in the scale that goes directly to explanation of that value (footnote marker or somesuch)
- change the state rollover color to something distinct from the gray that now represents the null-value (and/or change the null-value gray)
- review the color scale one more time -- might be good, might need some tweaking still. Also get feedback on switching to a scale that doesn't change when you select a commodity (I like...see what others think)

meiqimichelle:
As per discussion in Slack, we will merge for now to get partner feedback and continue working on the remaining tasks.

PR #443, opened 2015-01-19
Update DAP snippet location

konklone:
I moved the DAP snippet earlier today, as part of a [big DAP reorg](https://github.com/GSA/analytics.usa.gov/pull/105) as we prepare to launch the dashboard. 

I forgot that this site was pulling from my temporary CloudFront URL, and so I have caused you a several-hour outage (in your DAP-reported data only, not in EITI's own GA account). My apologies for this. I have replaced a copy of the DAP at the old URL, so that it will continue to work until this pull request is merged and deployed.

This PR updates the DAP location to pull from the (far nicer) new URL for our cloud-hosted copy of the DAP. (This is also a version bump of the DAP snippet, from 1.0.0 to 1.0.1.) I also added a note in the comments above, linking to the soon-to-be-launched dashboard.

Finally, I did some marginal tightening of the HTTPS settings for GA, as copied from the GA snippet we use on 18f.gsa.gov.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -41,8 +41,8 @@
     });
   });
   $(document).ready(function() {
-    Tipped.create('.term-tooltip', { 
-      position: 'topright', 
+    Tipped.create('.term-tooltip', {
+      position: 'topright',
       radius: false,
       maxWidth: 200,
       offset: { y: 5 },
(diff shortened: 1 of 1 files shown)


meiqimichelle:
Got it. I will merge here into our dev site. We have a review scheduled with our client today and if all goes well, we will merge into staging/production later this evening or tomorrow. Will let you know when that happens.

PR #521, opened 2015-07-10
Refactor server routes and data

shawnbot:
This PR moves all of the server routes to a standalone file, `routes.yml`, and moves the data logic to `data/decorator.js`. The main goal was to make the routes easier to declar, introduce automatic sorting of route URLs (reducing the potential for issues of ordering in `server.js`), and reduce the overall size of `server.js` for maintainability.

The `routes.yml` file is structured like this:

```yaml
/url:
  view: basename

/other/url:
  data: dataSourceName
  view: other-basename
```

Only the following properties of a route URL are recognized:

* `view`: the basename of the view file. If you have `view: foo`, the view will render `views/foo.html`.
* `data`: "decorate" the request with one or more data sources from `data/decorator.js`. This can be specified as a single string (the "data key"), a list of keys (inline: `['source1', 'source2']` or with a hyphenated list), or a map (`target: source`).
* `redirect`: redirect to the named route.
* `api`: expose the data as a JSON or CSV data endpoint (needs documentation)

### DIFF
--- a/data/decorator.js
+++ b/data/decorator.js
@@ -1,0 +1,106 @@
+var Festoon = require('festoon');
+var path = __dirname;
+
+console.warn('data/decorate path:', path);
+
+module.exports = new Festoon({
+  path: path,
+  sources: {
+    // master resources (commodities) list, plus other metadata
+    resources:  'commodities.json',
+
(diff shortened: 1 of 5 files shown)


meiqimichelle:
Thanks @shawnbot , this is great!

PR #524, opened 2015-07-13
Adds self-counting footnote styles

meiqimichelle:
- Uses [this technique](http://www.sitepoint.com/accessible-footnotes-css/). Pretty sweet :)


views/base.html
```diff
@@ -55,6 +55,9 @@
     </div> -->
 
     <main>
+      {% block communities %}
```
shawnbot: I think the block name "communities" is too specific. Can we call it "hero" or something?

meiqimichelle: As it stands, it includes the communities selector. It wouldn't be used anywhere else.

shawnbot: Maybe do it with an include and drop that into each of the communities pages then?

meiqimichelle: I couldn't get includes working :(

shawnbot: I just tested includes, and they appear to work. You should be able to do this:

1. Put the selector markup in, e.g. `views/includes/communities_selector.html` (probably good to put includes into their own directory)
2. Replace the `{% macro %}` template tag with `{% include 'includes/communities_selector.html' %}` (paths are relative to `views/`)

meiqimichelle: I'm still doing something wrong -- still fail on the includes. Do I need to tell node that the new include exists somehow? It "can't find the template."

meiqimichelle: Nevermind. Spelling.

meiqimichelle: Will be addressed when https://github.com/18F/doi-extractives-data/commit/8fbdbc246b4e68e11dec3c225e8790b125990327 merges.

PR #539, opened 2015-07-23
Resource maps

shawnbot:
This is a big PR that mostly consists of changes to the data, namely:

- I did a big rename of some of the data files to remove the redundant `-yearly` suffix (because all data is broken down by year) and replaced all instances of `volumes` with `production`.
- I created two new *regional* revenue data files, which combine state and offshore area revenues into single files:
  1. `data/output/regional/revenues.tsv`, which breaks down by year, region (2-letter state code or 3-letter offshore area code), commodity, revenue type, and revenue in dollars.
  2. `data/output/regional/resource-revenues.tsv`, which is just like the previous file but swaps out commodity for the broader "resource groups" that are used throughout the site.

  These files make it much more straightforward to load revenue figures on resource pages and color the maps accordingly.
- I added an `<svg is="eiti-map">` element to the revenue pages, along with some JavaScript that loads the `regional/resource-revenues.tsv` file and colorizes the map features according to their sum revenues.

### TODO
- [x] get a slider on the page
- [x] make the slider affect the map
- [ ] ~~change the resource page map coloration to be binary, as in [the prototype](http://18f.github.io/eiti-data/#/commodities/oilgas)~~ (not doing this for now)


meiqimichelle:
`Files changed: 101` woooooooo!

PR #541, opened 2015-07-24
Map HTML/CSS templates

meiqimichelle:
- Changes resources to revenues url to remove conflict with map work.
- Gets basic map template up and running

@shawnbot I thought more about how to implement the data types given what we talked about, and what we know about the data that is in reality available (versus theoretically available), and I think an implementation like this will work better than a table with stats. Instead, we get a dropdown with a splash area for related numbers to go.

This image shows two stats because the real data looks like:

- Revenues: one number
- Production: one number
- Jobs: two numbers (wage & salaried job number + self-employed job number)
- Exports: two numbers (absolute value + percentage of total exports)

With this implementation, we can flexibly display however many numbers per data type we need to. I can also envision turning the stats themselves into buttons if we want to give people the option to choose which of the two-number data gets displayed on the map, but I feel like we can leave that decision for another day and see how this goes at a basic level.

I think this also works for the 'combined' display we talked about -- [this prototype page](http://18f.github.io/eiti-data/#/revenue) with a map. The commodity type selector could say "compare resources" and you still have to choose a data type, say, revenues. The stats space then loads like:

- $900,900,540 total natural resource revenues in the U.S. in 2013 (in black)
- $389,900,540 coal revenues in the U.S. in 2013 (in the proper color)
- $235,000 hard mineral revenues in the U.S. in 2013 (in the proper color)
- ... etc ...

Otherwise, the biggest difference between this and the mocks is the styling on the selects for commodity and location. I will do that in a separate PR. Later.

screenshot 2015-07-23 21 21 00


meiqimichelle:
@shawnbot v1 map template ready for review. I tried to write out my thought on resolving the design issues we were talking about today, might not be clear enough. Ping me and we can talk.

PR #546, opened 2015-07-28
Add HTTP basic auth

shawnbot:
Hey @dlapiduz, can you check my work on this? I've brought in [http-auth](https://www.npmjs.com/package/http-auth) and am using it if `process.env.HTTP_BASIC_AUTH` is set. It's assumed to take the form `username:password`.

### DIFF
--- a/lib/server-helpers.js
+++ b/lib/server-helpers.js
@@ -119,10 +119,18 @@ var addRoutes = function(app, routes, data) {
   return _routes;
 };
 
+var httpAuth = require('http-auth');
+var auth = function(creds) {
+  return httpAuth.connect(httpAuth.basic({}, function(user, pass, cb) {
+    return cb(user === creds.username && pass === creds.password);
+  }));
+};
+
 module.exports = {
   view: view,
   redirect: redirect,
   api: api,
+  auth: auth,
   addRoutes: addRoutes
 };
 
--- a/server.js
+++ b/server.js
@@ -63,6 +63,15 @@ app.use('/data/commodities.json', express.static(__dirname + '/data/commodities.
 // get the data decorator
 var data = require('./data/decorator');
 
+if (process.env.HTTP_BASIC_AUTH) {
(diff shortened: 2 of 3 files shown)


lib/server-helpers.js
```diff
@@ -119,10 +119,18 @@ var addRoutes = function(app, routes, data) {
   return _routes;
 };
 
+var httpAuth = require('http-auth');
+var auth = function(creds) {
+  return httpAuth.connect(httpAuth.basic({}, function(user, pass, cb) {
+    return cb(user === creds.username && pass === creds.password);
```
konklone: For this particular line, I think using extra parentheses and being explicit about order of operations would be for the best.

server.js
```diff
@@ -63,6 +63,15 @@ app.use('/data/commodities.json', express.static(__dirname + '/data/commodities.
 // get the data decorator
 var data = require('./data/decorator');
 
+if (process.env.HTTP_BASIC_AUTH) {
+  var bits = process.env.HTTP_BASIC_AUTH.split(':', 2);
```
konklone: For this particular line, it'd be good to have some link, or description, of how you know `HTTP_BASIC_AUTH` is set to the `username:password` value (and ideally, why naively splitting on `:` is known to be acceptable).

konklone:
I left a couple comments. I was a little more demanding than I would be if were relying on a library, rather than implementing the username/password parsing and checking ourselves.

konklone:
This is the most active one I could find, and its main parser seems [rigorously commented and attributed](https://github.com/jshttp/basic-auth/blob/master/index.js):

https://github.com/jshttp/basic-auth

This one might also be good, and has an Express integration example, though it's CoffeeScript and I can't isolate the parsing logic as much:

https://github.com/gevorg/http-auth

I like the first one the best, personally. Both do the header parsing and require you to do the equality check yourself, which is fine, I just recommend using explicit parens so that you're not relying on implicit order of operations. I've been bit by that before. :)

Issue #558, opened 2015-09-11
Update footer with new words


meiqimichelle:
- [x] write words based on narrative and current MSG site
- [x] polish styles as per mock
- [x] HTML/CSS
- [x] Add sitemap/outline to footer

meiqimichelle:
Add sitemap/outline to footer

shawnbot:
I added a task for that in the description, @meiqimichelle

meiqimichelle:
Final task done with #600.

PR #575, opened 2015-09-13
Updates footer text

meiqimichelle:
Addresses #558 

- Removes DOI-exclusive references from footer
- Updates contact info
- Inserts placeholder text describing what will be in the footer

Looks like:
screenshot 2015-09-13 15 06 59


meiqimichelle:
I think we should merge this to master and I will get our product owner to review from there. So, if it looks good to you locally, please do the honors @gemfarmer .

Issue #522, opened 2015-07-13
Interactive county narrative subnav


meiqimichelle:
We will need a tab-like interaction for the county narratives. I would normally use jQuery for something like this. @shawnbot do you have a code snippet I can borrow, or like the select issue #520, can you send some tips my way?

![screenshot 2015-07-13 09 44 51](https://cloud.githubusercontent.com/assets/4827522/8655144/22db8c58-2944-11e5-9857-8bd8703c3701.png)

shawnbot:
Does it need to be tabs, or can these just be anchor links?

meiqimichelle:
They are anchor links. I mean -- they need the 'click link/tab, show a div over on the left and at the same time hide a div over on the left.' A nice class swicheroo. Does that make it clearer?

meiqimichelle:
I will merge my latest so you have the restructured community pages as per our last discussion.

shawnbot:
Sorry, I mean specifically, e.g.: `<a href="#geology">Geology</a>` links to `<section id="geology">`. This seems like a good start to me, then we can make them tabby as a second pass?

meiqimichelle:
Oh oh oh, sure, esp. if my ask is not a matter of "here's some instructions, follow them." I thought it might be an easy answer, like the select one turned out to be. I can do anchors for now.

shawnbot:
Yeah, I think we'll want to employ an approach along the lines of [practical ARIA tabs](http://heydonworks.com/practical_aria_examples/#tab-interface). Definitely another opportunity for a simple web component like the [aria-accordion](https://github.com/shawnbot/aria-accordion).

meiqimichelle:
On second and third thoughts (and after talking this through with some lovely people), tabs are not hte best approach. A sticky nav that updates as you scroll down (ie, you always see the nav, and the 'active' class moves depending on where you are on the page) would be better.

meiqimichelle:
The nav was made sticky with #585. (WOO) Leaving this open and moving back to backlog in case we have time to add active state class changes depending on where you are on the scroll down the page.

meiqimichelle:
Done.

PR #585, opened 2015-09-14
Sticky subnav on communites subpage

gemfarmer:
For #522.
Makes right-side navigation div sticky.
No changes made to accommodate mobile devices.


meiqimichelle:
:+1:

Issue #562, opened 2015-09-11
County narrative selector should persist


meiqimichelle:
If I choose a county I'm interested in on the selector, once it loads the name disappears. It should persist.

meiqimichelle:
Awww, nothing there:
![screenshot 2015-09-10 20 20 28](https://cloud.githubusercontent.com/assets/4827522/9806462/6d057102-57f9-11e5-8816-1a7b6f395d9c.png)

meiqimichelle:
Closed with #587

PR #587, opened 2015-09-14
Persist communities selector

gemfarmer:
For #562 County narrative selector should persist.
* made the `request.url` the source of truth to persist data.


meiqimichelle:
YAY

Issue #586, opened 2015-09-14
Confirm all datasets are using the latest


meiqimichelle:
@meiqimichelle to get latest datasets
- [x] company filter on current beta site
- [x] BLM fee download on current beta site (on documentation page)
- [x] revenues dataset on Version Next

@meiqimichelle to double-checkout live version of BLM dataset
- [x] BLM

@shawnbot to switch datasets - latest datasets [on share drive](https://drive.google.com/drive/folders/0B2CjDILjK8_jMHhWZVZNN0dMZ1E)
- [x] revenues dataset #590
- [x] company filter on current beta site

shawnbot:
Confirmed that the company revenues data in Google Drive is the same as [what's in gh-pages](https://github.com/18F/doi-extractives-data/blob/gh-pages/static/data/CY13_Revenues_by_company_03_18_2015.csv).

meiqimichelle:
Awesome, thanks @shawnbot !

meiqimichelle:
Confirmed that the BLM dataset is same as what's live on the beta site.

PR #590, opened 2015-09-15
[WIP] Update revenue data

shawnbot:
This PR updates the onshore and offshore revenue data per #586, using the latest Excel files from ONRR. I've documented the update process [on the wiki](https://github.com/18F/doi-extractives-data/wiki/Updating-the-Data).

One thing I did notice is that there are some new rows in the onshore revenue data that don't list a county. I'd like to validate that these rows made it into the state rollups before this merges.


meiqimichelle:
@shawnbot Let's talk about this today at standup. There should be data now that doesn't have a county, in particular for hardrock. Are you saying there is in addition some new data, and you just want to make sure its getting into the totals?

meiqimichelle:
I see this is checked off in #586. Should I review and merge?

PR #595, opened 2015-09-16
Update exports data

shawnbot:
This PR updates the exports data from Census to more closely match what we're using in the revenue views. This is ready to merge; the next step will be wiring up the data on the map at URLs like:

```
/resources/oilgas/exports/US
/resources/oilgas/exports/onshore/CA
```


data/Makefile
```diff
@@ -98,8 +98,10 @@ output/state/gdp.tsv:
 # state export figures
 output/state/exports-by-industry.tsv: input/census/top-state-exports.tsv
 	mkdir -p $(dir $@)
-	bin/parse-census-commodities.js \
-		$< > $@
+	bin/parse-census-commodities.js $< | \
```
shawnbot: Note: there's a much more liberal association of HS6 codes with commodities that we can enable with the flick of a command line switch (`bin/parse-census-commodities.js --liberal`) on this line. See `data/bin/parse-census-commodities.js` at about line 34 for the mappings, and check the URLs in the comments.

Issue #597, opened 2015-09-16
Data files should match URL structure


shawnbot:
In order to support the map on the resource pages, the data URLs should be made to mimic the URL structure so that we don't have to keep maintaining the rapidly growing [mess of conditionals](https://github.com/18F/doi-extractives-data/blob/75716fa1af5beb7a30a7fcb004efea61af96001a/views/resource.html#L202-L231) in `view/resource.html`. For instance, this URL:

```
/resource/all/revenue/US
```

displays a state and offshore ("regional") choropleth of total revenues across all resources, which is fed by `/data/regional/revenues.tsv` and summed up at runtime. Ignoring for now whether we should have a truly "national" view that summed up all of the states and just showed a line chart instead of a map, we should at least generate distinct data files for each of these views ahead of time, which will simplify the page (and JavaScript) logic and make any issues easier to debug (rather than hunting down whether we're rolling up by the right column on one page, we can just look at the generated data file).

This is something that we would probably do as part of #591, because that would allow us to ensure that one data file gets generated per URL in whatever process generates the page structure. The view-specific data files could even sit alongside the pages, e.g.

```
resource/
  all/
    revenue/
      US/
        index.html
        data.csv

      offshore/
        CHU/
          index.html
          data.csv
        ...

      onshore/
        AL/
          index.html
          data.csv
        ...
```

Maybe we'd need a more descriptive name for our view-specific data than `data.csv`, but I think you can see where I'm going here. Then the page templates for each different level of geographical hierarchy could change which map gets displayed (using our `<eiti-map>` custom element), but they would always be colored with `data.csv` (or whatever it's called) relative to the current URL.

Of course, doing this in Jekyll is going to produce a large number of files and might make build times totally unreasonable. What I don't know (and what we'll soon discover if we go down the road of #591) is whether Jekyll's incremental builds are smart enough to avoid building the entire site when, say, a single data file changes. We'll see!

shawnbot:
OK, I'm working on this now to unblock #606. I'm going to match the structure of the resource page's hash URLs and place each of the relevant files in a `/data/resources/` directory:

```
/data/resources/:resource/:datatype/:regiontype.csv (US)
/data/resources/:resource/:datatype/onshore/:region.csv (state)
/data/resources/:resource/:datatype/onshore/:region/:subregion.csv (county)
/data/resources/:resource/:datatype/offshore/:region.csv (offshore area)
```

e.g.

```
/data/resources/all/revenue/US.csv
[...]

Issue #553, opened 2015-08-07
Explore pulling Exports data into map


meiqimichelle:
These are the commodities groups:
- oil and gas
- non-energy minerals
- coal
(no renewables)

Two statistics
- %age of total
- Absolute value

Levels
- National
- State
(no county)

Dataset: https://www.census.gov/foreign-trade/statistics/state/data/index.html

shawnbot:
OK, so in the [exports-data branch](https://github.com/18F/doi-extractives-data/compare/exports-data) I've made some tweaks to get the state export data more in shape. The columns of [state/exports-by-industry.tsv](https://github.com/18F/doi-extractives-data/blob/exports-data/data/output/state/exports-by-industry.tsv) look like this as of e9e187e6:

* `State`: the two-letter state abbreviation
* `Commodity`: the individual commodity name ("Oil", "Coal", "Mining", etc.)
* `HS6`: the HS6 code used to derive the commodity name
* `Year`: the year
* `Value`: the absolute value (in dollars, I assume)
* `Share`: the percentage of that state's exports in that year
* `Resource`: the more general "resource group" identifier, e.g. `oilgas`, `coal`, etc.

I haven't looked at Isabelle's notes to see if there are more useful HS6 mappings, but the current ones are [hard-coded here](https://github.com/18F/doi-extractives-data/blob/dbdaebfeb291996d94a515a5a03712b44e5d849e/data/bin/parse-census-commodities.js#L21-L36), and require a data re-build in the data directory:

```sh
make -B output/state/exports-by-industry.tsv
```

Right now, for instance, the "Mining" commodity rows are categorized in the "other" resource group because we don't have a "Mining" commodity. The next thing to do is to look at the HS6 dictionaries ([here](http://www.foreign-trade.com/reference/hscode.cfm?code=25), [here](http://www.foreign-trade.com/reference/hscode.cfm?code=26), [here](http://www.foreign-trade.com/reference/hscode.cfm?code=27) and [here](http://www.foreign-trade.com/reference/hscode.cfm?code=28)) and map the 2- or 4-number codes to more specific *commodities*, then ensure that each of those has an entry in `data/commodities.json`. The goal here would be to re-build the data file and have it produce zero rows with `Resource=other`:

```sh
make -B output/state/exports-by-industry.tsv
egrep other !$
```

meiqimichelle:
MVP implementation is done for the Sept17 milestone so I'm removing from that list, but preserving the issue generally.

[...]

PR #598, opened 2015-09-16
Show exports data on the US map

shawnbot:
This gets us part way there with #553 by filling in colors on the map for states when you visit `/resources/all/exports/US`:

image

as well as the resource-specific national views, like `/resources/minerals/exports/US`:

image

It also activates the "data type" drop-down so that you can select Revenues and Exports, and disables the other two options for now. As mentioned in #597, the logic determining which data gets displayed on each page is a mess, but this'll work for now.

### DIFF
--- a/views/resource.html
+++ b/views/resource.html
@@ -60,11 +60,11 @@
 
           <div class="group_inline-right">
             <form>
-              <select>
-                <option value="revenue">Revenues</option>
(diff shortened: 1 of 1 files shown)


meiqimichelle:
Awesome! LGTM and :boom: :mouse:

Issue #580, opened 2015-09-13
Revenue process: non-energy minerals


meiqimichelle:
- [x] Text [on share drive](https://docs.google.com/document/d/1cwNnTW91G_jNsufVqvC1pMHgE7QQvkbWsLImVTiL_ZI/edit)
- [x] Background image [here](https://github.com/18F/doi-extractives-data/tree/rev-process/img/revenues)
- [x] Page styles

meiqimichelle:
Illustration. Text is not accurate.
![1 1 res to rev hardrock minerals](https://cloud.githubusercontent.com/assets/4827522/9908713/bac0c1c2-5c49-11e5-90cf-7dfae3a101b3.png)

meiqimichelle:
Closed with #599

PR #599, opened 2015-09-17
Revenues process: hardrock

meiqimichelle:
Addresses #580


meiqimichelle:
@gemfarmer Ready for review! FYI I adjusted the padding under the coal and hardrock illustrations so that they display nicely on my screen, which is what I'll be presenting from tomorrow. It may display strangely on your end though :( I left `TODO` in the css where those custom paddings are.

Issue #591, opened 2015-09-16
Switch back to Jekyll


shawnbot:
The whole Node server thing we've got going on is incurring an amount of technical debt that I'm not entirely comfortable with. I'd like to explore the idea of moving back to Jekyll so that we can leverage [Federalist](http://federalist.18f.gov/) for deployment, and make the site easier to work with, especially for those of us without much Node experience. This involves:

1. Migrating any data that's being iterated over in templates to [Jekyll data](https://jekyllrb.com/docs/datafiles/).
2. Evaluating whether [Jekyll collections](http://jekyllrb.com/docs/collections/) will solve the problem of deeply nested URL structures, like the `/resources/:resource/:datatype/onshore/:state/:county` hierarchy.
3. Restructuring the data build process to generate whatever full paths aren't covered by collections. If collections can't be made to generate those more complex URL structures, we would need to script the generation of them using the data as input. For instance, we would have a script that would generate a page for each combination of resource, data type, state and county in our data, and write it to the desired path.

One thing we'd get out of this would be a Sass stack that just works, rather than the troublingly broken `node-sass-middleware` package. Another would be static output, so we could host it on S3. And I think the advantage of creating a file structure that's more familiar would benefit everyone, since Jekyll is a well-documented thing that lots of people use and understand.

shawnbot:
Ben Balter wrote up a [nice guide](http://ben.balter.com/2015/02/20/jekyll-collections/) to using Jekyll collections that's worth reading.

shawnbot:
Here's another [guide to using Jekyll without plugins](https://captnemo.in/blog/2014/01/20/pluginless-jekyll/).

shawnbot:
As a step in this direction, I threw together a Node [module](https://github.com/shawnbot/data-directory) last night for reading a Jekyll `_data` directory into a single JS object. My thinking here is that we can move the "structural" data elements like `commodities.json`, which define the structure of certain parts of the site, into our `_data` directory, where they can be treated equally by both Jekyll and our Node scripts that set up the rest of the data-driven scaffolding.

shawnbot:
So, last week I created [a repo](https://github.com/shawnbot/jekyll-stress-test) to test Jekyll with various configurations of pages and data.

[...]

PR #610, opened 2015-10-06
Wire static pages

gemfarmer:
Fixes #608

I migrated the static pages to jekyll pages and kept all the former functionality intact. Pages live as `.md` files, but besides the front matter are still completely html.

@shawnbot @meiqimichelle up for review!

PR #613, opened 2015-10-07
Switch to Jekyll

shawnbot:
This will address #591.


gemfarmer:
@shawnbot sorry for the slow response. I was afk at an in-person background check interview. Because the `communities` pages function as is on #610, I'm thinking we can get my work there merged before diving into turning certain static pages into collections. 
After the branches are synced I will start another branch to handle those specific changes. Sound good?

meiqimichelle:
WOOOOOOOOO :dancers: :dancers: :confetti_ball: :ship:

PR #615, opened 2015-10-07
[WIP] Bh wire static pages

### LINKED ISSUE #614: Create communities collection
Create a communities `_collection`. Alike pages should be put in a custom collection instead of being put in pages subdirectories. 

Also, begin templating pages so that specific information can be passed through the front matter and edited more easily. This isn't particularly useful right now, but if more `communities` are added, it will save time.

gemfarmer:
Fixes #614


gemfarmer:
@shawnbot I didn't end up breaking the pages apart too much. Do you think I should do this now, or wait until we know that we will have more pages?

gemfarmer:
Closed with #623

PR #616, opened 2015-10-08
use datatype metadata

shawnbot:
This PR is a proposal for how we can handle the problem of different data types having drastically different URLs and structures (column names, etc.). It puts all of the data types into a Jekyll site data file (`_data/datatypes.yml`) and uses that both to output the `<select>` element for choosing a data type and introduces a `DataType` class on the JavaScript side that determines the data URL from each data type's "spec".

Let me know what you think, @gemfarmer!

### DIFF
--- a/_data/datatypes.yml
+++ b/_data/datatypes.yml
@@ -54,7 +54,7 @@ revenue:
         regiontype: onshore
         region: true
       value: Revenue
-      geo: FIPS # group by county
+      geo: +FIPS # group by county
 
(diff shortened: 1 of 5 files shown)


gemfarmer:
@shawnbot thank you thank you this is way cleaner!

gemfarmer:
@shawnbot actually looking into this longer, I noticed the logic for filtering the data in the `updateData` method is off. Looking into it now.

gemfarmer:
Let's get this merged! :ship:

PR #623, opened 2015-10-10
Bh wire static pages

### LINKED ISSUE #614: Create communities collection

gemfarmer:
Fixes #614 

Makes a `communities` collection for the existing communities, making it much easier to add content, especially footnotes.


gemfarmer:
@shawnbot or @meiqimichelle Could you take a look and see if you have any objections here! If none, I will merge :)

PR #601, opened 2015-09-17
[WIP] Active state on sticky nav

gemfarmer:
For ticket #522 
Functionality:
* [ ] on scroll toggles `active` class
* [ ] on clilck toggles `active` class

PR #630, opened 2015-10-15
Active State on Sticky Nav

gemfarmer:
For ticket #522. Instead of #601 
Functionality:
* [x] on scroll toggles `active` class
* [x] on clilck toggles `active` class


gemfarmer:
@shawnbot applied old changes to new branch. Still not the prettiest code, but works.

PR #631, opened 2015-10-15
Adds Gemfile

meiqimichelle:
Adds Gemfile so we can `bundle install` dependencies at version numbers we choose.

For now, the only gems included are:
`'jekyll', '3.0.0.pre.beta9'`
`bourbon`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,6 @@
+# A sample Gemfile
+source "https://rubygems.org"
+
+# gem "rails"
+gem 'bourbon'
+gem 'jekyll', '3.0.0.pre.beta9'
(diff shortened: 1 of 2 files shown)


gemfarmer:
@meiqimichelle add Gemfile.lock to the `.gitignore`.

gemfarmer:
Other than that, works for me locally :shipit:

meiqimichelle:
I'm confused about whether `Gemfile.lock` should be checked in or not. I'm reading that [it depends on the situation](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/), but I'm not sure what situation we fall under.

gemfarmer:
hmm. I guess I'm wrong. I've been approaching it wrong this whole time!

gemfarmer:
ship it :ship: it oh oh :ship: it like a :camera:

PR #644, opened 2015-10-21
Revenue regions

shawnbot:
This PR adds a new `/resources/revenue/` page that shows a map that links to each of the top-level onshore and offshore regions (states and offshore planning areas, which should probably be offshore _regions_) below, with a map for each. I stubbed out some CSS that would turn this listing into a modal interface with one region visible at a time, but wanted to get a feel for having them all on the page first.

The maps are fixed-height, so linking to sections on the page (e.g. `/resources/revenue/#CA`) works. I've also implemented some new logic in `eiti.load()` that spreads callbacks for multiple requests to cached URLs over multiple frames to progressively render the maps. Before this, rendering 60-ish maps on the page was blocking interaction (including scrolling) for over 3 seconds.

Also included:
* Uses single quotes throughout in `_sass/_main.scss`
* Adds a new `commodities` Jekyll data set that will be used to group commodities and list specific commodities for each group.
* Places some offshore planning areas into the Alaska region that were mistakenly assigned to the Gulf.
* Undoes @gemfarmer's addition of the `$baseurl` Sass variable, which was causing build errors in Jekyll 2.5.3.
* Fixes some bugs in the `<eiti-map>` custom element.
* Removes some legacy code in `eiti.js`.

Here's what they look like:

image


gemfarmer:
@shawnbot, what is the status of this branch?

gemfarmer:
@shawnbot Ok great!

gemfarmer:
Checked it out and looks good. I really like where the resources page is going. 

I'm going to merge.

Issue #618, opened 2015-10-08
[EPIC] Reorganize the interactive resources view


shawnbot:
I've been thinking a lot about how we could better organize the resources view since I've now recreated it three times. The hierarchy of inputs we've got feels arbitrary to me, and the number of things that differ between data types is making it hard to manage all of this stuff on one page. Here's how the *data* is organized, to the best of my knowledge:

Data Type | Resources | Sub-types | Geography
:---|:---|:---|:---
**Revenue** | all | royalties, rents, bonus, other | states, counties, offshore
**Production** | all | each resource has different products | states, counties, offshore
**Exports** | most | dollars, share of total | states
**GDP** | most | dollars, share of total (?) | states
**Jobs** | unknown | wage/salary, self-employed | states

Each of these data types requires a different set of inputs. Revenue is the only one that's relatively straightforward in the sense that it applies to all geographies and all resources. Production, though, can't be displayed in any meaningful way without selecting a product first. Exports data is super sparse, and the Jobs data is still a big unknown at this point. Either way, both Exports and Jobs apply only to states.

One thing that's making it hard for us to move forward on this part of the site is the dramatic difference between data types. I think that's as good a justification as any for making the **data type the first thing to select**, because it suggests that more things on the page will change when you select a new one. Doing so would allow us to do most of the UI-related stuff in Jekyll templates rather than JavaScript, which is always a good thing. Bottom line is that I think each of these data sets requires different framing (in terms of content and layout), and thus its own static page.

shawnbot:
Another thing that we haven't talked about his the resource subcategories (AKA the specific commodities), for instance Iron and Copper under Hard Minerals.

shawnbot:
(Note: I added GDP to the list too.)

shawnbot:
Hey @meiqimichelle I started stubbing out a straw man IA [on the wiki](https://github.com/18F/doi-extractives-data/wiki/Information-Architecture) interpreted from the [story board](https://michelle.storiesonboard.com/m/eiti-version-next). Let's talk about this next week!

shawnbot:
I just updated the table to list the revenue sub-types and the two job figures: wage/salary is money from paychecks, whereas "self-employed" is investment and other residual income.

shawnbot:
Here is my sketch from our session today:

[...]

PR #645, opened 2015-10-23
Update GDP figures

### LINKED ISSUE #643: Update BEA data industry codes
@Isabelle1512 says that we should _only_ be using industry code 21 ("Mining") for now, not `21,211,212,327,331` because:

1. 211 and 212 are _subsets_ of 21
1. 327 and 331 were not approved by the MSG

This also applies to [self-employment data](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#self-employment), which we'll be getting from the same BEA API.

Isabelle, this is easy to change, so if it takes some time to get confirmation back from the MSG that's okay, and we can open a separate issue for that.

This would need to change [here](https://github.com/18F/doi-extractives-data/blob/dev/data/bin/get-bea-data.js#L19-L20). Then, to update the GDP data files, run this from the data directory:

```
make -B national/gdp.tsv state/gdp.tsv
```

shawnbot:
This resolves #643 and starts to reorganize the data directory by data type, rather than geography per [this comment](https://github.com/18F/doi-extractives-data/issues/618#issuecomment-150057195). See #643 for notes on what's changed. This is not a breaking change, as we're not currently using the GDP data anywhere on the site.


meiqimichelle:
:mouse: :boom:

Issue #637, opened 2015-10-20
Fine-tune glossary appearance


gemfarmer:
Apply to site
* [x] Book icon/image 
* [x] x icon/image
* [x] Add book icon link to header
* [x] inline term reference (dotted underline)

ericronne:
@gemfarmer 
Here are three versions of the book icon. I assume you will only need one, though, which you can resize and recolor as needed. Reference ...
![glossary icon specs](https://cloud.githubusercontent.com/assets/9259185/10614350/5ea057cc-771f-11e5-9135-e9ead8601b95.png)

Main nav book:
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaMzBHMkpYZndHNmM/view?usp=sharing

Link book:
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaXzhmZjRZMTlmWTQ/view?usp=sharing

Expanded glossary book:
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaY2xjVXJDcklrcU0/view?usp=sharing

Close X icon:
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaekkxeFNISVM3UXM/view?usp=sharing

ericronne:
Some points for possible usability testing...
* [ ] When clicking to expand one word in a list, is it more usable to keep the other words visible, or to hide/filter them out?
* [ ] What is a clear call to action for the filter field?
 - (@gemfarmer's current wording)
 - "Enter a term to narrow the list"
 - _____________ (Greg?)

gboone:
* "Find other terms." 
* "Search for another term"
* "Look up other terms"

I think "find" "search" and "look up" might be more useful for interactions that don't show other terms than the one expanded.

ericronne:
Thanks, Greg. The tool only filters the list of 30 or so words, which will all be listed by default.

gemfarmer:
Closed with  #649

Issue #602, opened 2015-10-02
Set up new css architecture


meiqimichelle:
We should leverage the new 18F web standards and [this structure for overrides](https://github.com/18F/offsite/tree/18f-pages/_sass).

meiqimichelle:
Set up living styleguide. See FEC example.

gemfarmer:
@meiqimichelle I have started to play around with adding the Design Standards into the site. Not as straightforward as I was hoping. I'm waiting for Eric's redesigns to come  though before making any major changes

gemfarmer:
@meiqimichelle, are you imagining that the styleguide would be a separate app, or exist at `/styleguide` like it currently does? It seems like a `/styleguide` would be much more straightforward to implement,

gemfarmer:
Broke living styleguide off into #626

gemfarmer:
Goals/Stategies:
* add us web design standards as a base set of styles
* minimize overrides
* remove dead stylistic code
* narrow content width/change grid
* create new styleguide
* refactor eiti-specific typography
* refactor eiti-specific scss variables
What I have done:
* added us web design standards
* started creating styleguide
* removed unnecessary directories
What still needs to be done: 
* [ ] remov

gemfarmer:
Closed with #649

PR #649, opened 2015-10-23
New CSS Architecture

gemfarmer:
@shawnbot @meiqimichelle @ericronne I'd like to get your eyes on this.

This address #637 (fine tuning the glossary appearance) and #602 (new css architecture)

What does it do?
It creates a new _sass directory structure with the following directories:
* components – single elements or small clusters of elements that contain self-contained styling
* elements – these are eiti specific styles on commonly used elements such as links, lists, forms, typography
* extends – 
* layouts – styles for site structure
* lib – third party styles
* mixins – re-usable functions that are used to define styles site wide
* modules – Complex pieces made of components and elements

** Note: There are several 'modules' that I haven't broken apart or cleaned up yet


.gitignore
```diff
@@ -11,3 +11,4 @@ bower.json
 bower_components
 node_modules
 npm-debug.log
+styleguide
```
shawnbot: Do we really want to ignore this? I thought the idea was that we would keep it in git so it was visible on the live site?

meiqimichelle:
Besides @shawnbot's note, this looks good to me! I ran it locally, and spot checked the file system -- :+1:

.gitignore
gemfarmer: Ah, you are right! It will be included in the next push.

gemfarmer:
Since friday I...
* added all the changes made on #644
* removed the `_scss/modules` directory and moved all the scss files in it to `_scss/components`
* added a `_scss/blocks` directory to house all the non-component page specific code
* changed import structure components, elements, and blocks to import from `all`.

PR #667, opened 2015-10-26
Fix styleguide

gemfarmer:
Fixes #667 

These changes should work, but I can't view the [preview link](https://federalist.18f.gov/preview/18F/doi-extractives-data/fix-styleguide/) @shawnbot @meiqimichelle any suggestions for why I wouldn't be able to view a federalist branch?

screen shot 2015-10-26 at 4 32 26 pm


meiqimichelle:
Another `disk quota error` :sob:

gemfarmer:
:sob: no! I tried deleting unnecessary remote branches. Maybe that will help?

meiqimichelle:
The Federalist team increased our quota, we should be good. Or rather -- errors now are not because of the disk quota :trollface:

gemfarmer:
So I'm not sure why I didn't think of this before, but I'm pretty sure that the issue is that the npm task is never triggered by the jekyll build. Looking into ways to configure jekyll to accept alternative default build commands. @shawnbot do you have any experience doing this?

gemfarmer:
Actually, scratch that. I doubt that is the issue because stylguide is already in the `_site` directory

gemfarmer:
Actually, I think I just needed a less compiler for the site to generate kss.css

gemfarmer:
Ah! it was a .gitignore issue. This PR should fix the issue we were having with the missing styleguide kss.css!

meiqimichelle:
Looks good now! :mouse: :boom:

PR #671, opened 2015-10-27
[Sprint-MightyMoose] Dev --> staging

meiqimichelle:
This PR pushes [Sprint-MightyMoose](https://github.com/18F/doi-extractives-data/issues?utf8=%E2%9C%93&q=is%3Aissue+milestone%3ASprint-MightyMoose+) work to our staging environment.

- Sets up automated styleguide. #626
- Adds glossary feature. #549
- Gets Federalist working #604, #621, #629
- Sets up new css architecture #602

PR #676, opened 2015-10-28
hound syntax edits

gemfarmer:
For #671 

There were several hound syntax errors that were addressed here. 

I didn't address all of the errors. Many of them were from public/kss.js and vendor/list.min.js, both third party files.

@meiqimichelle

### DIFF
--- a/_sass/mixins/_type-mixins.scss
+++ b/_sass/mixins/_type-mixins.scss
@@ -13,61 +13,58 @@
   @if $level == h1 {
     // bold 30/40
     font-size: 1.875rem; // 30px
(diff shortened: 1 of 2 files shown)


gemfarmer:
I would prefer single quotes, but I'm fine with either. For now, I was just listening to hound. We could leave them in and change the config instead?

gemfarmer:
Sure, so that @meiqimichelle can merge the staging PR. For now I will keep them as single quotes.

gemfarmer:
@shawnbot ok, single quotes restored!

PR #706, opened 2015-11-05
Content bugs

nicoleslaw:
Overall, updating Get involved copy and making minor copyedits based on @Isabelle1512's feedback.

It sounds like we'll change the back links to proper breadcrumbs, but I tightened up the copy yesterday in the meantime.


meiqimichelle:
@gemfarmer I assigned this to you because there are conflicts with your latest -- probably just like the back links n stuff, but you're probably going to be the fastest at resolving :-/

gemfarmer:
Conflicts resolved. Assigning it back to you @meiqimichelle

gemfarmer:
@nicoleslaw did you want to take a look at the 'Did you know' content? I wrapped text in an `<span class="accordion-more"></span>` that would be initially hidden. I broke things off at the end of sentences, so maybe take a look and see if that break needs to be different.

meiqimichelle:
@gemfarmer Nicole added the longer 'did you know' content back in but it doesn't seem to be working on Federalist yet. I'm going to merge this for now (maybe its just a Federalist build problem), and issue the 'more' as a bug -- #715.

Issue #709, opened 2015-11-05
Stub Footer


gemfarmer:
* Home
* Search

* About USEITI (/about/)
* Executive Summary About USEITI (/about/report/)
* What's New (/about/new/)

* How it works (/resource_revenues/)
* Governance (/resource_revenues/governance/)
* The Process (/resource_revenues/:resource)

* Explore Data (/explore/)

* Case Studies (/communities/)
* Downloads (/downloads/)
* Glossary

Pages to be stubbed:

* Executive Summary About USEITI (/about/report/)
* What's New (/about/new/)
How it works
* Overview (/resource_revenues/overview/)
* Governance (/resource_revenues/governance/)
* Downloads (/downloads/)

Issue #701, opened 2015-11-03
Stub out Explore Data Landing Page


gemfarmer:
Refer to:
#655 Wireframes
#688 Design

gemfarmer:
Initial stub: https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-data-landing/explore/

PR #711, opened 2015-11-05
Stub footer links

gemfarmer:
For #709 

All of the links should work except explore because of explore data landing page issue which will do this and more ( #701 ).

:dragon: :fire: 

Preview here: https://federalist.18f.gov/preview/18F/doi-extractives-data/stub-footer/

@nicoleslaw @meiqimichelle


meiqimichelle:
@gemfarmer 'What's New' doesn't work. Missing ending slash?

gemfarmer:
@meiqimichelle Geez louise those forward slashes

PR #719, opened 2015-11-06
Pre demo Tweaks

gemfarmer:
Just a few style tweaks before the demo

* adjust padding on nav :last-item
* makes stubbed explore pages look nicer
* sub in new glossary icon (not in drawer itself. Need separate icon for that
* landing intro page changes @ericronne 
* offshore renewables tab fix
* fixed how it works right section. They were already changed, but somehow were overwritten
* made nav "Downloads" active styled

* updated name format linter to include .class-something_else: 
Play around with the convention regex here: https://regex101.com/r/vY2oS9/2

Though, doesn't look like hound is picking it up on github. Hmm
`Selector header-nav_item should be written in lowercase with hyphens` 
Thoughts @shawnbot @meiqimichelle ??


gemfarmer:
Hmm. Interesting. Well in that case this is ready for a merge! I'm going ot assign it to you @meiqimichelle , feel free to pass it off!

PR #726, opened 2015-11-09
Glossary updates

nicoleslaw:
- Updated terms and definitions from the final online content doc
- Added definitions from the diving graphics data section
- Copyedited for clarity

I would like to do one more pass on this branch after it's merged so I can check the formatting and link to some of the definitions inline in the Process sections.


meiqimichelle:
lgtm

Issue #752, opened 2015-11-09
Content: Add state legal and fiscal information


nicoleslaw:
- [x] Find a place for the content on pg 34-43 of the online only report
- [x] Ask @gemfarmer to help stub out the page
- [x] Put the content into HTML and copyedit as needed

nicoleslaw:
This one can be in the same subsection: [How it works > Governance](https://github.com/18F/doi-extractives-data/wiki/Information-Architecture) > State legal and fiscal information

gemfarmer:
@nicoleslaw I made the initial stub. Let me know if you want me to add anything else!

meiqimichelle:
@gemfarmer @nicoleslaw See my note in Slack -- I don't think there is a page for 'governance'

nicoleslaw:
@gemfarmer Passing this to you to check the styling. ✌️

Issue #696, opened 2015-11-02
Redo sticky nav


gemfarmer:
The sticky navigation needs to be turned into some kind of component.

The 'sticky' portion should be independent of the active state click/scroll functionality. So with that being said this issue should do the following:

* [x] Create re-useable component for a sticky area
* [x] Add page specific active state click scroll functionality for the following pages: 
* * [ ] `/case-studies/`
* * [ ]  `/governance/legal/` a la #782 

Blocked by #768 and #783

Some food for thought:
https://github.com/filamentgroup/fixed-sticky
https://github.com/1wheel/graph-scroll
https://github.com/sakabako/scrollMonitor

PR #782, opened 2015-11-13
Adding state and fiscal links doc

nicoleslaw:
@gemfarmer This is ready for your review (for #752). Heads up on the nested `ul`s. I tried to keep them to one level.

Also, please check my nav structure. I feel like this page is okay to be one long page, but we could also divide it up into subsections by state.

### DIFF
--- a/_how_it_works/governance-legal.md
+++ b/_how_it_works/governance-legal.md
@@ -2,355 +2,318 @@
 title: How it Works – State Legal and Fiscal Information
 layout: default
 permalink: /how-it-works/governance/legal/
-nav_items:
-  - name: ak
-    title: Alaska
-  - name: az
-    title: Arizona
-  - name: ca
-    title: California
-  - name: co
-    title: Colorado
-  - name: il
-    title: Illinois
-  - name: ky
-    title: Kentucky
-  - name: la
-    title: Louisiana
-  - name: mn
-    title: Minnesota
-  - name: mt
-    title: Montana
-  - name: nv
-    title: Nevada
-  - name: nm
-    title: New Mexico
-  - name: nd
-    title: North Dakota
-  - name: ok
-    title: Oklahome
-  - name: pa
-    title: Pennsylvania
-  - name: tx
-    title: Texas
-  - name: ut
-    title: Utah
-  - name: wv
-    title: West Virginia
-  - name: wy
-    title: Wyoming
 ---
 
 <div class="container-outer container-padded">
 
   <h1>{{page.title}}</h1>
-  
+
   <h3><a name="ak">Alaska</a></h3>
-  
+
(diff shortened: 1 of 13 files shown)


gemfarmer:
@nicoleslaw Ok added the list styles. I made a lot of changes relating to the list style changes that I made. 

I deleted the `nav_items` stuff. Right now that only works on the case studies pages, though given how long the page is, we probably should have a selector or a sticky navigation area to allow users to jump to a section.

gemfarmer:
@ericronne any ideas for creating a navigation to the different subsection on this page?
https://federalist.18f.gov/preview/18F/doi-extractives-data/state-local-fiscal-content/how-it-works/governance/legal/

ericronne:
We could try a right-side nav, a la case studies. Looks like we have 13 states, tho, so that might be pushing it. Still may be worth a try. That will help with line lengths, too.

Also we should use the bread crumbs for the header...

How It Works /
State Legal and Fiscal Information

gemfarmer:
@ericronne @nicoleslaw 
1. I will add the breadcrumbs
2. I think we should have some sort of navigation tool with that many states. Otherwise it is just a wall of text. Possible options that I see and thoughts on them:
* Sticky nav a la case studies. _potential downside: 13 states is a lot to have on a side navigation_
* Accordion nav with top open. 
* selector a la new explore/revenues/ page. Maybe just shows one state at a time?
* combination of the accordion nav, selector. Selector opens accordion?

ericronne:
**Sticky option 1**
Probably too long for a non-scrolling menu.
content-level page copy 5

**Sticky option 2**
Might be okay
content-level page copy 6

**Sticky option 3**
Also might be okay
explore - landing

nicoleslaw:
Thanks, guys! Yeah, it's way long without some sort of anchor/jump links. I don't mind v3; would that also work for mobile?

ericronne:
maybe accordion for mobile

gemfarmer:
Alternatively, in mobile the selector could stack on top fixed?

ericronne:
hm maybe.
guess i have some designing to do....

usEITI:
I liked option three as well. I think there are 18 states so too much
scrolling in option 1; I don't hate option 2 because it at least allows the
user to see right away that there are only 18 states and which ones they
are.

On Fri, Nov 13, 2015 at 4:19 PM, Eric Ronne <notifications@github.com>
wrote:

> maybe accordion for mobile
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/doi-extractives-data/pull/782#issuecomment-156559406>
> .
>



-- 
*Judy Wilson*
*Program Manager USEITI Secretariat *
*Office of Natural Resources Revenue*

*judith.wilson@onrr.gov <judith.wilson@onrr.gov>*
*202-208-4410*

gemfarmer:
So option 3?

meiqimichelle:
+1 for option three

gemfarmer:
Just to be clear before I start coding this, the content area is still going to be one long scroll area, right? I could see the alternative (only one state is visible) working well.

gemfarmer:
Because the sticky navigation area is tied up in upgrading the sticky nav code (#696). I'll add what is left over from this to that ticket.

Issue #772, opened 2015-11-10
Improve wording of dynamic Explore Data headings


shawnbot:
I'd like to improve the wording of the dynamic headings in our Explore Data sections, such as [revenue](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/revenue/). We've only done revenue so far, but as we build out production (see #679) and other sections it would be nice to have recommendations for how to string together descriptions of user-selected filters.

For instance, in the case production on federal lands (#679) we have three, possibly four filters that determine what you can see at any given time:

1. Commodity (e.g "Coal" or "All commodities")
2. Region ("Wyoming", "Central Gulf of Mexico", entire U.S.)
3. Year or year range (2008, 2004-2013)

And we have a heading that sums up the selections, but we haven't agreed on how to format it with those values into a coherent phrase. For instance:

* "Coal production on federal lands in Wyoming from 2004 to 2013"
* "Coal production on all federal lands in 2008"
* "All [commodity] production on federal lands in [the] Western Gulf of Mexico from 2004 to 2013"

@nicoleslaw or @gboone, do you have any ideas?

nicoleslaw:
I'm good with this general construction:

`{All production} on {federal lands} {in 2006}`
`{Coal production} on {federal lands} {from 2004 to 2013}`
`{Coal production} on {federal lands in Wyoming} {from 2004 to 2013}`

Let me know if you want to chat through. The dates are nice at the end how you had them, but we could also simplify to `(2006)` or `(2006–2008)` if that gives you a better amount of space to work with.

meiqimichelle:
Done.

Issue #679, opened 2015-10-29
Visual design: Explore > Federal Production


meiqimichelle:
This tracks the visual design of the 'revenue' data type in our explore data interaction. It follows on the wireframing work of #672.

shawnbot:
Okay, here's my first stab at this. There was one problem that I got a little bit hung up on, which is the requirement that the user select a product before being able to see anything, since we can't sum them up because each product is measured in different units.

On the left is what the product-specific view, which is very similar to @ericronne's design for revenue in #678, but with some different wording and labeling of the units. On the right is what we _could_ display before a product is selected, which is essentially binary: there's production or there isn't.

![image](https://cloud.githubusercontent.com/assets/113896/11069569/8fca3012-878b-11e5-9bb5-1280fb6fbe37.png)

I've cropped out the header and footer to focus on the important bits. Thoughts, @meiqimichelle @ericronne?

meiqimichelle:
Hmmmm. HMMMMMM. At first, I was thinking we could arbitrarily select a product to display. Then I remembered our old conversations about showing a binary (there is production or not) and your solution here showing show many products per region, and I like it. I also went to the need for an addition indicator for 'select a specific product'. I'm not sure of best placement for that -- maybe @ericronne can weigh on on that one. My initial thought is that you've made a good first guess and we can try it, and see how it goes @shawnbot .

ericronne:
@shawnbot note that i made a few minor tweaks yesterday (see #678) which impact this design...
"- filter hide
- year slider (two options, see far right board for alt)
- removed checkmark as indicator, because A, it implies you can choose more than one and B, it looks too much like the down caret"

https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaMmN1Y09sa3ppclU/view?usp=sharing

ericronne:
At this point, i wouldn't say i know the data well enough to offer a good opinion on what to show before they select any filters. Maybe after playing with it i'll have something to offer!

Meanwhile, a few potential stylistic tweaks ...
![image](https://cloud.githubusercontent.com/assets/9259185/11075212/b5bb9ddc-87b9-11e5-8469-043663addc4f.png)

- The show/hide filter change (noted above)
- Underline the three filter words in the header (dotted)
**Should we break each onto its own line, as i've done w/coal? I'm on the fence.**
- Remove extra space from under the header
- I shifted the US number down to line 1 of the table, as with the desktop design. Good there?

[...]

PR #807, opened 2015-11-18
Explore > Federal Production

shawnbot:
This PR gets the Explore Data > Federal Production section up and running, recycling much of the code from the Revenue section. You can preview it on Federalist here:

https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-production/explore/federal-production/

There are a couple of issues to work out before review and merging:

- [x] aggregating the product counts at the overall and subregion levels is busted
- [x] the bar charts shouldn't show up when we're listing product counts, per #679 
- [x] the timeline and map should display "binary" product counts, rather than with a linear scale (solution: modify the scale if a product isn't selected)
- [x] wording in dynamic headings should follow Nicole's recommendation in #772

I've also included a bunch of fixes for the Explore > Federal Revenue section as I made similar fixes and adjustments in this section.


meiqimichelle:
@shawnbot There might be something wrong with hardrock minerals -- I'm not getting any production numbers:

screenshot 2015-11-19 23 28 09

meiqimichelle:
Aaaahhh, I'm slowly figuring this out -- there's something happening with the `commodity` versus `product` options. When `commodity` shows up, things get wierd. See also -- this list is missing gold:

screenshot 2015-11-19 23 32 49

meiqimichelle:
Final note of the night -- how many products are there? I think the most useful bit of those drop-downs is the big list so you can see all the options and what they're measured in.  Can do just that, or just one top-level grouping plus all the products associated with it?

gemfarmer:
Just added some of my own changes to the branch. Conflict free!

usEITI:
I checked the CY production for federal and gold data is always withheld
because potential impact of the Trade Secrets Act.

On Fri, Nov 20, 2015 at 11:42 AM, Shawn Allen <notifications@github.com>
wrote:

> Yeah that might be a good call, @mieqimichelle. We could just turn it into
> one big drop-down for all the products. Maybe fancy auto-complete later?
>
> If gold isn't showing up, it could be that it's not present in the data.
> Because that product list is generated from the data.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/doi-extractives-data/pull/807#issuecomment-158454556>
> .
>



-- 
*Judy Wilson*
*Program Manager USEITI Secretariat *
*Office of Natural Resources Revenue*

*judith.wilson@onrr.gov <judith.wilson@onrr.gov>*
*202-208-4410*

PR #815, opened 2015-11-19
Update urls and file refs

gemfarmer:
For #794 . This updates the site URLs, file references, and file structure for Case Studies and How It Works. 

A lot moved around, so let me know if you see anything buggy. Ready for review.

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/update-urls/

@meiqimichelle @nicoleslaw

Issue #805, opened 2015-11-17
Visual Bugs: How it works (and global nav)


ericronne:
Thanks @gemfarmer ! :D

![image](https://cloud.githubusercontent.com/assets/9259185/11226383/1a4a81a6-8d45-11e5-89c3-5601b0534e82.png)

meiqimichelle:
Orange checkmark = fixes implemented with #816:
![visual-bug-hunt](https://cloud.githubusercontent.com/assets/4827522/11277550/3d8a4ae2-8ea4-11e5-821e-c07a0c925654.png)

Issue #788, opened 2015-11-16
Visual Bugs: home intro block


ericronne:
- [x]  Add 40 or 50 more pixels space above the text block ("The United States...") to push it down
 - [x] Make that same intro sentence H2
 - [x] Add "USEITI" before "was created" (Ok, @nicoleslaw? I changed this to a complete sentence since we can't control the break of the intro line.)
 - [x] Increase the padding(?) on the right side of that text block (by 10 or 20)

*Note: We may still need to adjust the size of the "USEITI was created ..." text, but let's see how it all looks with the above changes applied.*

Thx!

nicoleslaw:
@gemfarmer Holler at me when you want me to look at the copy. Happy to clean it up. :pencil2:

meiqimichelle:
Live on [dev(https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/) for review.

Issue #749, opened 2015-11-09
Visual bugs: case studies


ericronne:

ericronne:
Here are some tweak suggestions for the case study layout. These should also inform other pp to a certain degree. I marked the changes directly on the image, as you can see. Less trackable, i suppose, but easier to see. LMK if this technique doesn't float your boat, @gemfarmer, thx!

![visual bugs - case study](https://cloud.githubusercontent.com/assets/9259185/11074052/2f75cbe0-87b3-11e5-86f6-73629ca4a604.png)

ericronne:
@gemfarmer
And a couple for the home-page intro block :dancer: 
![image](https://cloud.githubusercontent.com/assets/9259185/11076338/ce8111fc-87bf-11e5-8ad8-1975a8987b80.png)

meiqimichelle:
@ericronne The latest version of the homepage is on [staging](http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/doi-extractives-data/) -- it is much closer to your mock, but could still use your notes. Can you make a new issue for each one of these images so that it is easier to track which ones are complete?

usEITI:
Thank you for the visual mock, it was help for me to understand the tweaks.  They look good.

nicoleslaw:
Related note for this [Overview page](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/communities/), some of the headings are a bit funky. I wonder if we could simplify the page a bit by removing the second set of case study nav, etc.

meiqimichelle:
@nicoleslaw Did your note on the overview page get resolved? If not, can you explain more (or screenshot) what you mean? I'm not following :-/

nicoleslaw:
Not yet; I can show you during our chat, but there are two sections of nav for the case studies (the drop-down menu and the `Choose a community to get started` section) and I find the "six commodities" part to be a bit confusing.

meiqimichelle:
@nicoleslaw Got it. Yes, @ericronne mentioned something about the funky nav there as well. I'm going to open a new issue on case study nav refinement so we can prioritize -- #796

meiqimichelle:
The case study visual bug fixes are ready for review on [dev](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/case-studies/).

usEITI:
Yeah again.  What is the box and icon under Employment for Boone, Logan,
and Mingo Counties?

[...]

PR #816, opened 2015-11-19
Visual bug hunt

meiqimichelle:
This PR is based on #815 . That needs to merge before this. @gemfarmer , this is ready for your review. I'd like to hit the 'diving graphic' visual bugs next :smiling_imp: . I don't think we need @ericronne to review/approve this before merge -- I think of this as an initial pass, and we can open another pull request to get more bugs. That's why I'm not closing the visual bug issues via this PR.

**This addresses some of the visual bugs in #805. Explore data and the global nav now look like:**

screenshot 2015-11-18 21 45 05

**It also addresses the bugs in #788 and some of the sizing issues in the intro in #749. The homepage now looks like:**

screenshot 2015-11-18 22 41 20

**Finally, it also comments out, ipsem-izes or changes some remaining old placeholder text that is now OBE, and we were getting concerned comments on.**


gemfarmer:
Ok. All looks good. I made a few quick changes, but nothing major. I'm going to merge.

Issue #836, opened 2015-11-20
Stub out second page for state legal info


nicoleslaw:
Related to #804: We have two long sections of content about state governance. I think we should divide up `state-legal-fiscal-info` and `state-laws-and-regulations`. We have a page for the second part, but I should move the content in that doc to the other (new) file as it's secondary information.

nicoleslaw:
Thank you!!

PR #842, opened 2015-11-20
new stubs added

gemfarmer:
For #835 and #836 

This creates two new page stubs.

@nicoleslaw, can you take a look and tell me if this is in a good place?

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/new-page-stubs/how-it-works/


nicoleslaw:
Looks great, thank you!!

Issue #686, opened 2015-10-29
Visual design: Explore > GDP


meiqimichelle:
This task is for visual designs based on the wireframes in #682. It is blocked until that task is complete.

meiqimichelle:
I think this one is no longer relevant -- we used similar designs to implement this, right? I'm going to close -- @ericronne feel free to re-open if I've got this wrong. We broke out a remaining design challenge as its own issue in #895.

Issue #682, opened 2015-10-29
Wireframe: Explore > GDP


meiqimichelle:
This task is to create a wireframe for the explore data view of [GDP data](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#gdp).

shawnbot:
Here's the wireframe! This assumes that we're not breaking down figures by commodity group (read the rationale in the [data catalog](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#gdp)).

![image](https://cloud.githubusercontent.com/assets/113896/11347248/65bb3baa-91d5-11e5-9dbc-a3e646a6458b.png)

shawnbot:
@meiqimichelle, would you mind reviewing this?

meiqimichelle:
:+1: :tada: :balloon: :confetti_ball: :shipit:

PR #873, opened 2015-11-24
Adds Explore > GDP

### LINKED ISSUE #854: Code: Explore > GDP
Pending design in #686 (and wireframe in #682).

* [x] load and display $ data
* [x] get % data #650 
* [x] dollar vs percent UI
* [x] load and display % data
* [ ] make the table sortable (% or $)
* [ ] figure out how to figure US level as distinct from region level data
* [ ] get the unit selector working as designed (the `<select>` is a hack)

gemfarmer:
Fixes #854, #686, #682


gemfarmer:
@shawnbot I'm thinking that unless we get county level data, the single bar histogram doesn't' do that much for the single state view. Instead, maybe a $ number, % number, and a pie chart?

Issue #734, opened 2015-11-09
Content: Explore data landing page


meiqimichelle:
The landing page current has filler text for an intro to each dataset interaction. These need words.

nicoleslaw:
All set on this for now.

nicoleslaw:
Wrong issue! :)

nicoleslaw:
@meiqimichelle @shawnbot Do we have any documentation or info about these data sets? 
- Federal revenue by location
- Federal revenue by company
- Company-reported federal revenue (Is this related to reconciliation? We should make the heading clearer as it sounds too much like the previous.)

coreycaitlin:
This page has content now, so I'm closing this out. This wound up being easiest to refine in parallel with the explore data detail pages. (See #959 and #997.)

PR #879, opened 2015-11-24
[WIP] Explore landing content

nicoleslaw:
Fixes most of the sections on #734


meiqimichelle:
Talked to Nicole, there's a comment on what's left on the issue but we can merge this part in the meantime if we want.

Issue #878, opened 2015-11-24
Content: Copyedit footer


nicoleslaw:
- [x] Style updates
- [x] Better placeholder text

PR #874, opened 2015-11-24
Fixes bug in footer hrefs

gemfarmer:
Fixes #868 

This PR addresses the bug that was a follow up to adding the footer. 
* It changes the hrefs so that they no longer have a trailing slash e.g. `/#production`.
* removes old footer files (that I thought had already been removed)
* makes sure that ids on pages corresponded with the hrefs in the footer

Ready for your review, @meiqimichelle

PR #880, opened 2015-11-24
Footer edits

nicoleslaw:
Text refinements for #878


meiqimichelle:
I moved these edits over in #874. Hopefully all transferred well! I changed the text about interviewing because we'll allow free-form comment there as well. Closing this in favor of #874.

Issue #796, opened 2015-11-17
Nav refinement on case studies page


meiqimichelle:
There are two sections of nav for the case studies (the drop-down menu and the `Choose a community to get started` section). The the "six commodities" part is also a bit confusing. The design should be refined to address these issues.

ericronne:
Here's a rough-ish tweak and a county layout notion ...

**Main Case Studies Pg**
![content-level page copy 7](https://cloud.githubusercontent.com/assets/9259185/11446545/152a7f30-94fe-11e5-9b88-16845960dd67.png)

**County Selected**
![content-level page copy 8](https://cloud.githubusercontent.com/assets/9259185/11446552/30074838-94fe-11e5-9177-3a5c24f5efc8.png)

meiqimichelle:
@gemfarmer I started work on this in #910 but stopped before touching the sticky nav. See note in that PR.

gemfarmer:
@meiqimichelle @ericronne What is the status on the individual state images for this issue?

meiqimichelle:
@gemfarmer We don't have them yet. I wasn't sure if we'd have time to implement the state-switching so didn't prioritize getting the images. We can see about getting that moved up if you think this isn't too bad tho!

meiqimichelle:
Done.

PR #910, opened 2015-11-30
New case studies layout

meiqimichelle:
**Work in progress, do not merge**

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/new-case-studies-layout/case-studies/
This PR starts to implement the UI tweaks in #796. Right now, it looks like:
screenshot 2015-11-29 20 21 53

@gemfarmer, I'm sending this your way so you can see what I changed, and also see about the best way to update the sticky nav. I didn't try to fix it since I feel like you said you wanted to improve some other things on it anyways (might be remembering wrong).

### DIFF
--- a/_case-studies/boone-logan-and-mingo.md
+++ b/_case-studies/boone-logan-and-mingo.md
@@ -5,7 +5,7 @@ permalink: /case-studies/boone-logan-and-mingo/
 resource: coal
 nav_items:
   - name: intro
-    title: Introduction
+    title: Top
   - name: geolo
     title: Geology and history
   - name: produ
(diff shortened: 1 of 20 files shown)


gemfarmer:
_taking a look now_ @meiqimichelle I'm going to try to separate the sticky aspect of the nav from the actual content. With time in mind. I'm not sure I will have time to make it an actual component. :(

gemfarmer:
Ok @meiqimichelle I took a look and upgraded the stick nav. It wasn't a full upgrade (still not a true web component), but at least now the code for being sticky and the code for controlling the 'active' state of the nav is separate. It should be much easier to add this to other pages.

meiqimichelle:
Great! I pulled, looks good. I just made a minor tweak to the text of each subnav. I also see that we need to open another issue for the mobile view of this new layout :tired_face: but this is good to go! :mouse: :boom:

Issue #915, opened 2015-11-30
Bug: Opening glossary and closing it opens up global nav


gemfarmer:
If you open the glossary, closing it with the "X" opens the global nav. 

I suspect the problem is the code that causes the nav to toggle when you select "Glossary" within the nav. The nav should only toggle if the glossary is already open.

PR #918, opened 2015-11-30
Mobile Nav Bugs – Round 1

### LINKED ISSUE #913: Visual bug: global nav isn't centered on desktop
It was centered, and then I merged the mobile view PR too quickly, and didn't notice that we had a regression. Booooo. Looks like the following, with USEITI all the way to the left:

screenshot 2015-11-29 20 21 53

It should be centered still, with right and left edges lining up with content below.

gemfarmer:
Fixes #913 and #915 

Other smaller stylistic nav bugs will be addressed in a separate PR.

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/mobile-nav-bugs/

This PR...
* makes it so that the mobile nav opens and closes only when it is supposed to
* centers the header on desktop
* eliminates padding on the right side of the header.

@meiqimichelle ready for review

### DIFF
--- a/_sass/blocks/_header.scss
+++ b/_sass/blocks/_header.scss
@@ -26,10 +26,15 @@
 header,
 .header {
   @include outer-container($max-width);
+
   margin-bottom: $base-padding-large;
   margin-top: $base-padding;
   margin-left: $base-padding;
   margin-right: $base-padding;
+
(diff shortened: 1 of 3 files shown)


meiqimichelle:
:mouse: :boom:

Issue #861, opened 2015-11-23
Code: emulate case studies on 'State Legal and Fiscal Info' section page


gemfarmer:
Currently blocked by #910 . Waiting for new sticky-nav.js script

For 'State Legal and Fiscal Info', extend the template for case studies to work for this page. Nav items should be styled.

There is an issue with design details somewhere, but I can't seem to find it. @ericronne.

https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/how-it-works/state-legal-fiscal-info/

gemfarmer:
@ericronne I can't seem to find the issue where we were talking about the different options, but if I remember correctly, we decided to go with a right-aligned selector. Do you have the mocks for that?

ericronne:
#796 ?

PR #927, opened 2015-12-01
Add selector to state and fiscal info layout

gemfarmer:
For #861 

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/state-fiscal-layout/how-it-works/state-legal-fiscal-info/

This PR:
* Works within the narrative layout to create a state selector that assists in navigation though the wall of text in `how-it-works/state-legal-fiscal-info/`
* Adds selector in mobile view

@meiqimichelle. Ready for review

Still to do: Selector 'active' state should reflect url and place on the page --> #928


meiqimichelle:
:mouse: :boom:

Issue #858, opened 2015-11-23
[WIP] Code: Explore > Exports


shawnbot:
See design in #685 and wireframe in #681.

Issue #685, opened 2015-10-29
Visual design: Explore > Exports


meiqimichelle:
This task is for the visual designs based on the wireframes in #681. It is blocked until that task is complete.

meiqimichelle:
I think this one is no longer relevant -- we used similar designs to implement this, right? I'm going to close -- @ericronne feel free to re-open if I've got this wrong.

Issue #681, opened 2015-10-29
Wireframe: Explore > Exports


meiqimichelle:
This task is to create a wireframe for the explore data view of [exports data](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#exports).

shawnbot:
This wireframe is _very_ similar to #682 because the data has basically the same exact shape: exports and GDP are both measured in $ and % of total at the national and state levels.

![image](https://cloud.githubusercontent.com/assets/113896/11347405/2c6f9980-91d6-11e5-983e-a48ce4a2e748.png)

@meiqimichelle, assigning to you for review.

meiqimichelle:
:+1: :tada: :balloon: :confetti_ball: :shipit:

gemfarmer:
@shawnbot did we decide to not filter by `Resource`? I don't see anything in the wireframes, but the column is there

shawnbot:
@gemfarmer we do have commodities in this data set, but for some reason I think we decided not to make them selectable. Maybe @meiqimichelle remembers?

gemfarmer:
@shawnbot `HS6`, `Commodity`, and `Resource` are all columns.

meiqimichelle:
@gemfarmer It may have been an oversight as we were copying wireframes, or maybe we decided it didn't makes sense because of the weird 'top 25' thing in the data (something about only the top 25 commodities for each state show up, so like maybe California *has* oil, but it comes in at number 26, and so if you filtered by oil California wouldn't show up). Does that jog any memories @shawnbot ?

PR #936, opened 2015-12-01
Explore > Exports

gemfarmer:
For #858, #685, #681 (code, design, wireframe)

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/exports/explore/exports

It is good to go! I used gdp as a template and referred to wireframes in #681.

Questions/todos remaining:
* I did not address the $/% selector. I will do that in a separate PR
* The wireframes did not include the Resource column, so I did not bother adding it. Is this on purpose, or unintentional, @shawnbot 
* Nationwide totals not accurate. Currently they are actually the same values as Alabama (first alphabetically), but this could be remedied if we found data for US totals ($) and shares (%) by year.

@Isabelle1512, @shawnbot suggested that I talk to you about holes in the exports data. 
What's going on: we don't have nationwide totals, so we can't accurately asses the Nationwide $ total or % of total exports.
Do you know where we can get Nationwide total $s and % ? I looked in the executive summary and only found values for specific years and specific industries. 

@meiqimichelle Should the problem above be treated as a bug or blocker? We can merge this and file the Nationwide values as separate bugs to be addressed separately. Or we could wait for the data


data/bin/parse-census-commodities.js
```diff
@@ -95,7 +95,8 @@ async.waterfall([
     rows.forEach(function(d, i) {
       if (i === 0) console.warn(d, years);
       years.forEach(function(year) {
-        var value = +d['val' + year];
+        var value = +d['val' + year] * 1e6;
+        var share = +d['share' + year.substr(-2)] / 100
```
shawnbot: +1

meiqimichelle:
@gemfarmer I say merge this for now and file the nationwide issue as a separate bug to make sure we are pushing as much as we can out for review.

gemfarmer:
Ok. I'll merge and make a separate issue detailing the issue. !

PR #941, opened 2015-12-01
Mobile search text

### LINKED ISSUE #940: Content: Mobile search text
The search field text on mobile (`_includes/layout/nav-drawer.html`) could use some love. It currently says `Search site` under a `Search` heading.

nicoleslaw:
Fixes #940. Ran this by @ericronne as it is slightly inconsistent with how the Glossary works (searching v filtering a subset of results), but it was repetitive with the placeholder text. I looked at other mobile patterns, and it seems to be the norm to move away from placeholder text in favor of a clear search/spotlight button next to an empty field.


gemfarmer:
@nicoleslaw the distinction between the two inputs sounds smart. Should we focus the input when the nav opens? I didn't add this originally because I wouldn't want a keyboard opening up when I open a navigation area, but if we are getting rid of the placeholder, maybe we want something to draw attention to the search. Or does that matter? What are the patterns that you saw?

gemfarmer:
Just made some additions to get the branch up to date :)

gemfarmer:
Also, I'm going to merge this. If you think it is worth having the cursor focus on the search bar, ping me and I can make an issue.

Issue #926, opened 2015-11-30
Data: Explore > Jobs > Wage & Salary


shawnbot:
[This data](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#salary-employment) comes from BLS ([here](http://www.bls.gov/cew/datatoc.htm)) and needs to be downloaded and transformed to match the format we've been using.

shawnbot:
@Isabelle1512 can you help me figure out which field in the data we should be using? I think it's the one labeled "Annual average of monthly employment levels for a given year" [in the field layout](http://www.bls.gov/cew/doc/layouts/sic_csv_annual_layout.htm):

![image](https://cloud.githubusercontent.com/assets/113896/11511405/c1598cde-981d-11e5-9c93-7a9179be0a2c.png)

Just want to make sure. Thanks!

shawnbot:
Whoops, I meant to reference [the1975-2013 field layout](http://www.bls.gov/cew/doc/layouts/csv_annual_layout.htm), which has many more columns:

![image](https://cloud.githubusercontent.com/assets/113896/11511795/dc38a8b2-981f-11e5-89f7-17d99fc9505d.png)

Isabelle1512:
@shawnbot that is the one! (annual_avg_emplvl)

shawnbot:
Thanks, @Isabelle1512! It looks

Quick status update: my first impulse was to try streaming the zip downloads in Node so that we could programmatically grab the one file we need from each 140mb (!!) zip, but both [unzip](https://www.npmjs.com/package/unzip) and [unzip2](https://www.npmjs.com/package/unzip2) both have issues with BLS's zip files. So now I'm writing a Makefile to download each zip and extract the CSVs we care about from each, then I'm putting those into `data/_input/bls`. After that's done, I'll add those files to git and write a script to stream them all in and convert them into the tabular format that we want with just the county-level data.

For now I'm going to focus on getting the NAICS code 21 data only, which is all mining/extractives. This means that we won't have natural resource type breakdowns initially, but we can bring them in during a second pass.

shawnbot:
This is done. We can handle any problems with the data in new issues.

Issue #923, opened 2015-11-30
Data: Explore > Jobs > Self-Employment


shawnbot:
[This data](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#-self-employment) comes from BEA, so we can script accessing the API and getting it in a form that's useful for us.

shawnbot:
@Isabelle1512, can you explain this note in the [data catalog](https://github.com/18F/doi-extractives-data/wiki/Data-Catalog#-self-employment)?

> It [...] must be calculated, by subtracting SA27N Full- and Part-Time Wage and Salary Employment by NAICS Industry from SA25N Full- and Part-Time Employment by NAICS Industry

I'm just not sure what the difference is between the SA25N and SA27N numbers, and why we would subtract one from the other.

shawnbot:
@Isabelle1512 So, when I access the SA27N numbers from the BEA API, they're described as "Private nonfarm wage and salary employment: Mining". That actually sounds like the number we'd want for wage & salary in #926, no? If so, subtracting makes sense: the difference between the total and that number is the self-employment number. Can we use this wage and salary number instead of the BLS one?

shawnbot:
I might be answering my own question here, but I compared the BLS and BEA numbers for wage and salary, and there is a difference. For California in 2013 I get **27,851** jobs from BEA and **24,726** from BLS, for an 11% difference.

shawnbot:
@Isabelle1512 I'm going to assign this to you for review for now. We'll have numbers up on the site for you to check shortly.

shawnbot:
@Isabelle1512 you can review the numbers [on the site](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/jobs/) or in the [raw data files](https://github.com/18F/doi-extractives-data/tree/dev/data/jobs) (TSV, should open in Excel).

Isabelle1512:
[...]

Issue #855, opened 2015-11-23
Code: Explore > Jobs


shawnbot:
See visual design in #860 and wireframe in #683.

shawnbot:
[Ready to preview!](https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-jobs/explore/jobs/)

shawnbot:
Merged, done!

PR #942, opened 2015-12-01
Explore > Jobs (data only)

shawnbot:
This is just the data side of the Explore > Jobs section, which should take care of #926 and #923 for now. These numbers could use a review, but that'll probably be easier to do in the interface than in their current form. I'm going to self-merge this because it doesn't affect anyone, then re-use the branch for building out #855.


meiqimichelle:
:confetti_ball:

PR #945, opened 2015-12-01
Explore > Jobs

shawnbot:
This implements #855 with data from #942.

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-jobs/explore/jobs/


meiqimichelle:
Let's merge this and pull out the remaining issue (total for country showing up on state filter) as a separate issue. :mouse: :boom:

PR #946, opened 2015-12-01
Adds ids to each section of the download page

### LINKED ISSUE #921: Add ids to data downloads page
So that we can link directly to each download from the explore data sections.

meiqimichelle:
Now we can link directly to a dataset on the downloads page. Closes #921.


gemfarmer:
Looks good to me!

PR #963, opened 2015-12-03
All Lands Production

gemfarmer:
Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-all-production/explore/all-lands-production/

This PR is simultaneously getting the data in order and adding the client side html and js.

Where I am at with the data:
I have currently collated the data for coal (just 2013), natural gas, and oil.
The data for natural gas and oil need to be updated because I was drawing numbers from onshore withdrawals, not total withdrawals :disappointed: .

Todo (Data)
- [x] update natural gas and oil data
- [x] find missing coal data (if there is any other than 2013)
- [x] FOUND. adding to pool.
- [x] add renewables data (if it is correct data. appears to be energy production, not extraction), write related script
- [x] update data catalog
- [x] remove all data before 2004 

Where I'm at with the code:
I have copy pasted html and js from what we have for federal data, so there is plenty to do
- [x] Remove ability to browse by county (we only have county level data for coal
- [x] Change display to volume (not $)
- [ ] totals don't update on product change
- [ ] states not filling in without counties present
- [ ] state view: totals are for Product total nationwide, not for a state
- [ ] selecting state does not limit ability to select products


gemfarmer:
@meiqimichelle @shawnbot. The data is finished. after the demo and additional meetings, I'm tempted to merge this, and fix the list of bugs as a separate ticket.

gemfarmer:
Left over todo items moved to #977 to allow for merge

gemfarmer:
@meiqimichelle @shawnbot ready for review!

meiqimichelle:
@gemfarmer I'd like to try rebasing to pull in the latest layout changes. What say you? Good idea/bad idea?

meiqimichelle:
Yeah, nevermind. I think it will be easier to merge and then go back and fix the layout problems, if there are any :P

gemfarmer:
Thanks for merging. It was potentially a nasty rebase. I think you were right @meiqimichelle that it will be better to fix after.

gemfarmer:
@shawnbot yeah it looks your fix did the trick! Thanks

PR #965, opened 2015-12-03
Breadcrumbs to many more pages

meiqimichelle:
Adds latest html and styles for breadcrumbs to many more pages. There are still a few missing because they're in-process data visualizations and so we should wait till those are done to adjust this detail in the code.


gemfarmer:
OMG thank you @meiqimichelle! The breadcrumb inconsistency was actually starting to bother me.
_reviewing_

gemfarmer:
Looks great! :mouse: :boom:

PR #966, opened 2015-12-03
Cleans data viz styles

meiqimichelle:
This PR brings the data visualization code close to Eric's mocks. What this does:
- makes the maps bigger
- gets all typography closer to mocks
- uses an ugly hack to make the summary sentence appear where it's supposed it, but I should work with @gemfarmer or @shawnbot to make this better later. For example, it shouldn't have a blue background when the filters are closed.
- adds links to download/docs
- adds text where applicable
- brings all visualizations that are live on dev to a common point (these edits are on all pages)

## Desktop
screenshot 2015-12-03 16 28 06

## Desktop, filters open
screenshot 2015-12-03 16 28 20

### DIFF
--- a/_explore/default.md
+++ b/_explore/default.md
@@ -72,7 +72,7 @@ permalink: /explore/
 			<h5 class="landing-heading"><a href="{{site.baseurl}}/explore/reconciliation/">Reconciliation</a></h5>
 			<button class="accordion-button" accordion-button title="Toggle for reconciliation"></button>
 			<div class="accordion-content">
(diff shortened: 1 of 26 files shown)


gemfarmer:
The big maps are amazing. Such a big difference!
Also, are filter toggles supposed to be different on mobile? I really liked the big blue buttons!

screen shot 2015-12-03 at 10 21 07 pm

gemfarmer:
I don't think the hack is a blocker. I made an issue here to cover it --> #976

ericronne:
@gemfarmer mobile gets the full bar, desktop gets the button, both eventually will be sticky

Issue #1007, opened 2015-12-07
Bug: case studies 'active' state should reflect url and place on the page


gemfarmer:
* [x] nav should update on scroll
* [x] nav should update on select

PR #1008, opened 2015-12-07
Update Case Studies Nav

gemfarmer:
For #1007 

This monitors updates the open sticky nav on case studies so the correct item is active! Significantly overhauls and componentizes code.

Preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/active-nav/case-studies/

Ready for review


meiqimichelle:
Woooooo!

Issue #995, opened 2015-12-07
Add sticky subnav to long navigation pages


gemfarmer:
Nav should be applied to long law and governance narrative pages.
`how-it-works/state-legal-fiscal-info.md` is a good template. 

the following jekyll frontmatter:
```
nav_description: Jump to a state
nav_items:
  - name: ak
    title: Alaska
```

with corresponding markup: 
```
<h3><a name="ak" data-nav-header="ak">Alaska</a></h3>
```

Issue is complete when:
* [x] nav added to Laws and Governance Pages
* [x] nav added to Natural Resource pages
* [ ] nav added to any additional pages
* [ ] mobile version of nav added to all of the above

meiqimichelle:
This was (all? in part?) done with #1015. I'm not sure if there are other pages that still need this treatment. What do you think @gemfarmer ?

gemfarmer:
I'm not entirely sure what the full list of narrative pages is. I did all of the Laws and Governance pages, but perhaps there are more??

gemfarmer:
It looks like the [Natural Resources](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/how-it-works/#natural-resources) pages also are narratives.

gemfarmer:
I'm going to close this. If we end up adding it to additional issues, we can revisit it.

PR #1029, opened 2015-12-09
Add narrative layout to natural resource pages

gemfarmer:
Addresses part of #995. This doesn't fix this whole issue, only part of it

* Adds narrative layout/sticky nav to the how it works > natural resources pages.
* Adds a `nav_group` option to the sticky nav so that we can have select `opt-group`s

ready for review!


meiqimichelle:
:candy: :mouse: :boom:

Issue #974, opened 2015-12-04
Visual design: charts/images for case studies


meiqimichelle:
At most, each of the 12 case studies has the following needs. 6 of the 12 only have one chart (production). Original charts and location images all live [here on the share drive](https://drive.google.com/drive/folders/0B2CjDILjK8_jN0pUUWk3MG9LbzQ).

## Example: Boone, Logan, and Mingo Counties, West Virginia
https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/case-studies/boone-logan-and-mingo/
### Production chart
![wv-production](https://cloud.githubusercontent.com/assets/4827522/11580852/d937a160-99f8-11e5-96c1-01a0b3672f7f.png)
### Wage chart
![wv-wage](https://cloud.githubusercontent.com/assets/4827522/11580855/ddf21d66-99f8-11e5-9902-4dc4a7ff015d.png)
### Location
![screenshot 2015-12-03 20 05 04](https://cloud.githubusercontent.com/assets/4827522/11580873/2a7c3a36-99f9-11e5-8319-9a8795b62a7c.png)

ericronne:
@shawnbot do you have a favorite source for state images? thx!

shawnbot:
@ericronne we can actually use the web components I made to render these maps. Just show me where they should go and I can drop them in.

ericronne:
ah, natch

shawnbot:
Or @meiqimichelle if you're interested in trying this out, the markup will look something like this:

```html
<svg is="eiti-map" data-path="{{ site.baseurl }}/data/geo/" simplify="1e-2"
  projection="albersCustom" preserveAspectRatio="xMidYMid meet">
  <g data-url="us-topology.json"
    data-object="counties"
    data-filter="properties.state === 'WV' &amp;&amp; properties.name.match(/Boone|Logan|Mingo/)"></g>
  <g data-url="us-topology.json"
    data-object="counties"
    data-filter="properties.state === 'WV'"
    data-mesh="true"></g>
</svg>
```

You'll also want to be sure that the component is included:

```html
<script src="{{ site.baseurl }}/js/components/eiti-map.js"></script>
```

shawnbot:
Then you should be able to style them by targeting these CSS selectors:

* `path.mesh`: county outlines
* `path.feature`: the highlighted county features

@ericronne filling in the other counties black will be harder with this approach. Is it okay to leave them white?

ericronne:
[...]

PR #1033, opened 2015-12-09
Adds selected county images to case studies

meiqimichelle:
For #974.

Adds state outlines with relevant counties colored for case studies.

screenshot 2015-12-10 19 54 47

### DIFF
--- a/_layouts/case-studies.html
+++ b/_layouts/case-studies.html
@@ -188,7 +188,7 @@ layout: default
 						data-mesh="true"></g>
 					<g data-url="us-topology.json"
 						data-object="counties"
-						data-filter="properties.state === 'MN' &amp;&amp; properties.name.match(/St Louis/)"
+						data-filter="properties.state === 'MN' &amp;&amp; properties.name.match(/Saint Louis/)"
 						class="case_studies-selected_counties"></g>
 				</svg>
 			{% elsif page.permalink == '/case-studies/desoto/' %}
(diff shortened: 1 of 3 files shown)


_layouts/case-studies.html
```diff
@@ -34,9 +34,38 @@ <h3 class="case_studies_content-icon"><i class="icon-iron"></i>Iron</h3>
...
+
+			{% if page.permalink == '/case-studies/' %}
+      	<img src="{{ site.baseurl }}/img/counties/12-map.png" alt="Map with locations of the twelve communities given in-depth coverage in this section of the site." class="case_studies_intro-image">
+			{% elsif page.permalink == '/case-studies/boone-logan-and-mingo/' %}
+				<svg is="eiti-map" data-path="{{ site.baseurl }}/data/geo/"
+				  simplify="1e-2"
+					projection="albersCustom"
+					preserveAspectRatio="xMidYMid meet"
+					zoom-to="WV">
+					<g data-url="us-topology.json"
```
shawnbot: @meiqimichelle this is all it took! I left the other states in there to provide context, but you can filter them out by adding `data-filter="id === 'WV'"` to the states layer.

shawnbot: Oh, and one more thing: the highlighted counties aren't filled right now, but when they are they still won't show up because the other counties are drawn on top of them. You can fix this by reversing the order of the two counties layers to put the highlighted ones on top.

gemfarmer:
:tada: :balloon:

Issue #977, opened 2015-12-04
Code: Fix bugs on Explore > All Lands Production


gemfarmer:
Spillover from #963 

Still to do on Explore > All Lands Production
- [x] totals don't update on product change
- [x] states not filling in without counties present
- [x] state view: totals are for Product total nationwide, not for a state
- [x] selecting state does not limit ability to select products

meiqimichelle:
Copying over status from PR #1036 for the remaining part of this task:

The only thing left to do is add topojson for joint region that I have added Pacific alaska which contains both the offshore regions Pacific and Alaska. @shawnbot how did you generate the topojson? I was logging some of the make commands that seemed relevant, but unable to figure out where I could merge the two. :(

From talking to @mentastc the most important part is that the Pacific alaska is included in the total and the bar charts (correct me if that is wrong, Chris).

The current status for that feature is:

- I have added the svg path, and included it in the filter selector
- It is not clickable as a region :( (not sure how to address this)
- The total displays, but not with a bar
- The region is not visually represented on the map ( I believe this is because it is not its own topoJSON feature, so is not being matched with the data source)

PR #1036, opened 2015-12-09
[WIP] Explore production bugs

gemfarmer:
preview: https://federalist.18f.gov/preview/18F/doi-extractives-data/explore-production-total/explore/all-lands-production/

* [x] coal
* [x] natural gas
* [x] oil
* [ ] offshore oil

@meiqimichelle Ok. I'm taking a break on this one. I'm stuck and I could focus my energy on something else for a little bit.

The only thing left to do is add topojson for joint region that I have added Pacific alaska which contains both the offshore regions Pacific and Alaska. @shawnbot how did you generate the topojson? I was logging some of the make commands that seemed relevant, but unable to figure out where I could merge the two. :(

From talking to @mentastc the most important part is that the `Pacific alaska` is included in the total and the bar charts (correct me if that is wrong, Chris).

The current status for that feature is:
* I have added the svg path, and included it in the filter selector
* It is not clickable as a region :( (not sure how to address this)
* The total displays, but not with a bar
* The region is not visually represented on the map ( I believe this is because it is not its own topoJSON feature, so is not being matched with the data source)


gemfarmer:
@meiqimichelle Ok. I'm taking a break on this one. I'm stuck and I could focus my energy on something else for a little bit.

The only thing left to do is add topojson for joint region that I have added `Pacific alaska` which contains both the offshore regions `Pacific` and `Alaska`. @shawnbot how did you generate the topojson?

meiqimichelle:
I copied the remaining part of this task over to #977. The rest looks good! WOOOOO!

Issue #877, opened 2015-11-24
Content: Documentation > Reconciliation


nicoleslaw:
Add reconciliation info (p. 1-7 from Executive Summary Reporting and Reconciliation .docx)
- [x] What is the scope of the revenue payment data reconciliation?
- [x] How did the Independent Administrator perform the reconciliation?

coreycaitlin:
Okay to close this out, or do we want to pretty up the tables this week?

meiqimichelle:
The content it there, so close! I think there's another issue for making tables across the site better.

Issue #1041, opened 2015-12-10
Explore > chart titles and total lines : match mocks


meiqimichelle:
Right now, our chart titles and total lines aren't clear (which is a heading, and which is a total?). 
<img width="714" alt="screenshot 2015-12-09 20 09 27" src="https://cloud.githubusercontent.com/assets/4827522/11705880/f0ba841c-9eb0-11e5-927c-1c7cd7f37532.png">

They need to match our [mock](https://cloud.githubusercontent.com/assets/9259185/11066186/c0242096-878c-11e5-8335-3c2a34f2297a.png).

meiqimichelle:
We're a little closer now when #1038 merges:

<img width="736" alt="screenshot 2015-12-09 22 40 45" src="https://cloud.githubusercontent.com/assets/4827522/11707836/3582c5e0-9ec6-11e5-83f4-73c805ce72f6.png">

meiqimichelle:
Overtaken by events. We restructured the way the explore data sections look and are coded, so this is no longer relevant.

PR #1051, opened 2015-12-10
Reconciliation documentation: sidebars and tables

coreycaitlin:
This adds sidebars and tables to the reconciliation documentation page (#877). The tables are not beautiful, but they're there.


coreycaitlin:
Fixed one mention of the number of companies reporting, which should help close out #1041.

Issue #1061, opened 2015-12-10
Link on homepage card exec summary needs to be updated


meiqimichelle:
The homepage card for the exec summary is an old link. It needs to be updated to go to the exec summary page.

meiqimichelle:
Done with #1062

PR #1062, opened 2015-12-10
Fix executive summary homepage card

coreycaitlin:
it links to the right place now, per #1061


meiqimichelle:
:boom: :confetti_ball:

PR #1063, opened 2015-12-10
Show regions with withheld federal production

### LINKED ISSUE #935: Don't exclude withheld production values
The production totals are off right now because I'm [filtering out withheld values](https://github.com/18F/doi-extractives-data/blob/f99e11d04171f5a0e39635b566b78bf41c5cb0c0/data/bin/normalize-county-production.js#L44-L47) from our output data. We shouldn't be doing this.

The solution is probably just to comment out [these lines](https://github.com/18F/doi-extractives-data/blob/f99e11d04171f5a0e39635b566b78bf41c5cb0c0/data/bin/normalize-county-production.js#L44-L47) from the script that "normalizes" production values.

shawnbot:
This fixes #935 by including withheld production volumes in the data that the front end displays, and updates the JavaScript to list regions with withheld volumes as "(withheld)".


meiqimichelle:
Woooooo!

screenshot 2015-12-10 15 35 05

Issue #1040, opened 2015-12-10
Explore > all maps: region hover states 


meiqimichelle:
The explore data maps need hover states so people know that regions are clickable.

shawnbot:
See [here](https://github.com/18F/doi-extractives-data/blob/dev/_sass/components/_eiti-map.scss#L17) and add a `path.feature` rule.

PR #1067, opened 2015-12-10
Lighens up svg lines. Adds hover state.

meiqimichelle:
For #1040.

- Lightens svg lines a tad as per design direction
- Adds simple hover state, although this could be improved later. Right now, any hover comes in behind the `style` being set in the JS so we can't use fill, and stroke is also a little wonky. Ideally, we'd replace this solution with something that truly loads on top of the svg.

screenshot 2015-12-10 16 32 34


gemfarmer:
Woo! :boom: :mouse:

PR #1068, opened 2015-12-11
Adds case study images

meiqimichelle:
Resizes and switches out case study images for latest versions.


gemfarmer:
very nice! :boom: :mouse: :boom:

Issue #1053, opened 2015-12-10
Explore > Exports: hide nationwide totals


shawnbot:
Because #948

meiqimichelle:
Done with #1070.

Issue #1045, opened 2015-12-10
Explore > jobs: intro sentence should include job type


meiqimichelle:
The intro sentence on the jobs page doesn't indicate which type of job you're looking at. It should.

`Extractive industries jobs in the entire U.S. (2013)` -->

`Extractive industries jobs (wage & salary) in the entire U.S. (2013)`

meiqimichelle:
Done with #1070.

Issue #1037, opened 2015-12-10
Explore > Jobs: Values should be on the left of the bar


meiqimichelle:
All our charts now have values on the left side of the bars except jobs

<img width="747" alt="screenshot 2015-12-09 19 38 06" src="https://cloud.githubusercontent.com/assets/4827522/11705360/6f9818e4-9eac-11e5-8af4-b985490051a7.png">

This might be a part of #955 and will be resolved when that is done.

Issue #889, opened 2015-11-25
Fix eiti-map component in Firefox


shawnbot:
Something in or related to the [eiti-map component](https://github.com/18F/doi-extractives-data/blob/dev/js/components/eiti-map.js) isn't working in Firefox, and I actually think that it's the [`eiti.load` function](https://github.com/18F/doi-extractives-data/blob/dev/js/eiti.js#L19). We could suss this out with some in-browser unit tests.

shawnbot:
Elaborating on this a bit: I think the issue is that in the bowels of `eiti.load()` there's a deferral of cached requests using [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) that just isn't :fire: in Firefox. The maps don't load, and FF doesn't log any errors in the console, so I'm fairly certain that nothing is actually throwing an error.

shawnbot:
Turns out it's because the `document.registerElement()` polyfill that we're using doesn't support type extension (that's the feature that allows us to "extend" the `<svg>` element). So now I'm modifying the custom element to _wrap_ SVG elements like this:

```html
<eiti-map data-path="etc" zoom-to="id" and-so-on>
  <svg>
  </svg>
</eiti-map>
```

usEITI:
You guys are amazing

On Thu, Dec 10, 2015 at 7:48 PM, Shawn Allen <notifications@github.com>
wrote:

> Turns out it's because the document.registerElement() polyfill that we're
> using doesn't support type extension (that's the feature that allows us to
> "extend" the <svg> element). So now I'm modifying the custom element to
> *wrap* SVG elements like this:
>
> <eiti-map data-path="etc" zoom-to="id" and-so-on>
>   <svg>
>   </svg>
> </eiti-map>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/doi-extractives-data/issues/889#issuecomment-163798023>
> .
>



-- 
*Judy Wilson*
*Program Manager USEITI Secretariat *
*Office of Natural Resources Revenue*

*judith.wilson@onrr.gov <judith.wilson@onrr.gov>*
*202-208-4410*

shawnbot:
Fixed in #1070!

PR #1070, opened 2015-12-11
<eiti-map> component fix, and some explore tweaks

shawnbot:
This PR primarily deals with #889 and changes the `eiti-map` custom element from one that extends SVG elements (`<svg is="eiti-map"></svg>`) to a standalone element that expects to have an SVG element as a child (`<eiti-map><svg></svg></eiti-map>`). This requires all of the pages that use the component to be modified, and while I was in there I fixed a couple of other things:

* #1053 Explore > Exports: hide nation-wide totals
* #1037 Explore > Jobs: move number to the left of the bar
* #1045 Explore > Jobs: add job type to filter description sentence


meiqimichelle:
Looks good!

Issue #978, opened 2015-12-04
Code: Home > Explore data intro


meiqimichelle:
Based on visual design in #743 and content in #951.

meiqimichelle:
@shawnbot Unless you say otherwise, I'm taking this out of this sprint. We'll use flat images for now, @ericronne is on the job in #1030 .

shawnbot:
Sounds good, @meiqimichelle.

meiqimichelle:
Graphics are ready to go in #1030 . @gemfarmer, I assigned this to you so you can track the getting the tabs working but we can split it.

meiqimichelle:
I lie. I'm going to assign this to me because I'm going to start it tonight. :tada:

meiqimichelle:
Reference for tabs:
http://heydonworks.com/practical_aria_examples/#tab-interface

gemfarmer:
@meiqimichelle Ok. Let me know if you need me to take a look at the tabs. I was looking at that reference earlier, and it looks pretty slick!

meiqimichelle:
Styles are v1 done when #1075 merges. Remaining bugs:

- [ ] I changed the order of the `<li>` and `<a>` in the tab HTML so I don't think all the JS is targeting properly. Arrow and tabbing to the, er, tabs doesn't work.
- [ ] I used another `<li>` and `<a>` combo in the content section itself for the links in the bottom-right of each section, which I think is causing issues. When you select a non-default tab, you can't click on those links. They kinda reload the homepage in an odd way. I think they're being co-opted by the tab JS.

ericronne:
Good stuff @meiqimichelle!

The pngs look a little fuzzy. Should i re-render them at exact size, as opposed to 2x?

ericronne:
Here's the [correct revenue map png] @meiqimichelle (https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaM05WOGdRd1ZYYXc/view?usp=sharing)

meiqimichelle:
@ericronne do all the pngs look fuzzy, or just the map? Replacing the map now....

meiqimichelle:
Oh and @ericronne , the tabs do a little dance when you click on them because of the bold on the selected tab. If we want no dance, then we should not change the font weight between the selected/not-selected tabs.

ericronne:
Yep anticipated that

ericronne:
[...]

PR #1075, opened 2015-12-11
Explore data homepage teaser

meiqimichelle:
For #978.

Styles are done! Looks like:
screenshot 2015-12-10 23 16 41

JS is mostly done! I added bugs to the original issue.


gemfarmer:
Ok. Merging for now. We can address those link bugs tomorrow!

Issue #1034, opened 2015-12-09
Explore > sticky filter area


gemfarmer:
The filter area on the explore pages should be sticky.

Should be able to use sticky component to handle this

meiqimichelle:
Done with #1080.

PR #1080, opened 2015-12-11
[WIP] Add sticky filter

gemfarmer:
For #1034 

@meiqimichelle ready for review. Probably missing some details, but much better than before.


meiqimichelle:
WOOOOOOOOOOO THIS IS AWESOME! Checked in Chrome and Ffx, seems good! We can add details later, tho honestly I don't see too many off the bat! :mouse: :boom:

Issue #1082, opened 2015-12-11
Bug: Explore > mobile filters gone


meiqimichelle:
I think something regressed when we added the sticky filters on desktop

/cc @gemfarmer

meiqimichelle:
Done with #1084

PR #1084, opened 2015-12-11
mobile sticky filter

gemfarmer:
@meiqimichelle add on to #1080 

* makes the sticky filter work in mobile


meiqimichelle:
Lovely! We'll fix the styles later, merging! WOOO! Closes #1082.

PR #1107, opened 2015-12-12
Add SVG map of the 18 in-scope states

meiqimichelle:
This PR uses the same technique as we used in the case study narratives to draw states/counties to draw the US with 18 states highlighted.

### DIFF
--- a/_how-it-works/laws-and-regulations/state-laws-and-regulations.md
+++ b/_how-it-works/laws-and-regulations/state-laws-and-regulations.md
@@ -30,30 +30,26 @@ nav_items:
   {% include hash_selector.html %}
 </nav>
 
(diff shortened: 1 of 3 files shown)


meiqimichelle:
Ah. Actually, I think this is something broader. Now our county SVGs aren't showing up either:

screenshot 2015-12-14 11 07 19

meiqimichelle:
Fixed thanks to direction from Shawn! This PR now updates the old case study code to the latest implementation of the `<eiti-map>` component as well as adds an 18-focus-state map as shown below. Ready for review.

screenshot 2015-12-14 14 24 35

meiqimichelle:
Eh, everyone's super heads-down today. Merging.

Issue #1057, opened 2015-12-10
Explore > offshore area names need adjustement


meiqimichelle:
We now call offshore areas by their short names, which leads to this in the bar chart:
![screenshot 2015-12-10 13 33 01](https://cloud.githubusercontent.com/assets/4827522/11727789/3fcefad2-9f43-11e5-9667-3bc97fc32ecd.png)

Alaska is doubled because on is offshore. We should rename the offshore areas so that this is clear.

meiqimichelle:
@gemfarmer Is this something I can change in a text file, or is this something we're aggregating/creating somewhere? I will fix if you can point me in the right direction.

meiqimichelle:
@gemfarmer I don't remember where this one landed. If it is easy, maaaaybe we can fit it in this sprint. Not sure if there's time.

meiqimichelle:
This should be live on `dev`. :tada: It will merge live as soon as the cross-browser map issues are resolved. YAYS!

PR #1108, opened 2015-12-12
Change offshore area names to include the word offshore

meiqimichelle:
This is for #1057. Right now, our offshore areas are called, for example, `Alaska`. This causes confusion when they're listed in the bar chart, for example next to the state of Alaska.

I searched for the only instance of `Pacific alaska` and found what I think is the csv that controls those names. I think the data needs to be re-loaded or something tho before the changes can take effect. Either way, someone else should look at my suggested change.

/cc @gemfarmer @shawnbot

### DIFF
--- a/_data/offshore_regions.csv
+++ b/_data/offshore_regions.csv
@@ -1,6 +1,6 @@
 name,id,viewbox
-Alaska (offshore),alaska,75 415 202 173
-Atlantic (offshore),atlantic,730 125 230 440
-Gulf (offshore),gulf,470 440 295 135
-Pacific (offshore),pacific,20 60 161 341
-Pacific-Alaska (offshore),pacific alaska,20 60 180 530
+Offshore Alaska,alaska,75 415 202 173
+Atlantic Ocean,atlantic,730 125 230 440
+Gulf of Mexico,gulf,470 440 295 135
+Pacific Ocean,pacific,20 60 161 341
+Offshore Pacific-Alaska,pacific alaska,20 60 180 530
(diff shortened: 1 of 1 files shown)


_data/offshore_regions.csv
```diff
@@ -1,6 +1,6 @@
 name,id,viewbox
-Alaska,alaska,75 415 202 173
-Atlantic,atlantic,730 125 230 440
-Gulf,gulf,470 440 295 135
-Pacific,pacific,20 60 161 341
-Pacific alaska,pacific alaska,20 60 180 530
+Alaska (offshore),alaska,75 415 202 173
+Atlantic (offshore),atlantic,730 125 230 440
+Gulf (offshore),gulf,470 440 295 135
```
shawnbot: Could we just say "Gulf of Mexico" here? And likewise, "Atlantic Ocean" and "Pacific Ocean"?

meiqimichelle: Yep! Anything that distinguishes it from other things. So -- still Offshore Alaska? And then what you suggest for the rest?

shawnbot: Yeah, I think "Offshore Alaska" or "Alaskan Offshore" make sense. Maybe @mentastc has opinions?

_data/offshore_regions.csv
```diff
@@ -1,6 +1,6 @@
...
-Alaska,alaska,75 415 202 173
-Atlantic,atlantic,730 125 230 440
-Gulf,gulf,470 440 295 135
-Pacific,pacific,20 60 161 341
-Pacific alaska,pacific alaska,20 60 180 530
+Offshore Alaska,alaska,75 415 202 173
+Atlantic Ocean,atlantic,730 125 230 440
+Gulf of Mexico,gulf,470 440 295 135
+Pacific Ocean,pacific,20 60 161 341
+Offshore Pacific-Alaska,pacific alaska,20 60 180 530
```
meiqimichelle: How about these options @shawnbot ?

PR #1114, opened 2015-12-14
Change search results requested to 50

dhcole:
Currently it's at 200, which takes too long to return results. 50 at a time seems more reasonable.


meiqimichelle:
Thank you @dhcole !

PR #1138, opened 2015-12-15
Sticky case studies

gemfarmer:
Fixes a bug in the case studies subpages, where the sticknav would render on top of the map. I created a bit of a hack with the data-attributes in the sticky nav to get around this.

Can be reviewed!


meiqimichelle:
Looks good!

PR #1145, opened 2015-12-15
Browser fixes

### LINKED ISSUE #1143: Explore > maps/charts aren't working in IE

shawnbot:
This PR gets the maps and charts working in Firefox and (mostly<sup>1</sup>) in IE11 and fixes #1143. Here's what it took:

1. For IE11, we needed a [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) polyfill, since that browser's native implementation doesn't allow you to construct one directly.
1. IE9+ (in theory, but only tested in IE11) has a non-confirming implementation of the [DOMTokenList](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList)'s `toggle()` method because it ignores the second parameter, as in `element.classList.toggle('foo', false)`. I've added a shim for that specific method.
1. Upgraded to [document-register-element](https://github.com/WebReflection/document-register-element/) v0.5.3 and added it as a dependency in `package.json`.
1. Got the bars working in all explorations by using the custom element constructor instead of `document.createElement()`. See [this note](https://github.com/18F/doi-extractives-data/issues/1143#issuecomment-164899010) for more info.
1. Cleaned up some code and added some jshint comments for hound.

<sup>1</sup> Re: maps in IE11, there appears to be an issue with the viewBox scaling when maps are zoomed in.


gemfarmer:
I didn't do a complete cross browser test, but all of the changes that you mentioned appear to be working!

Issue #1157, opened 2015-12-18
Remove "This site is in development" from banner


gemfarmer:
<img width="261" alt="screen shot 2015-12-18 at 11 59 02 am" src="https://cloud.githubusercontent.com/assets/4803473/11903430/c396ce30-a57e-11e5-8941-13ec92352998.png">

meiqimichelle:
These should not be removed because this is meant to show that we're constantly improving the site.

coreycaitlin:
I'm going to take a look at what we've done with other sites post-launch, and potentially discuss with the content group. Assigning to myself as a reminder!

meiqimichelle:
From the USDS family:
![screenshot 2015-12-28 10 41 10](https://cloud.githubusercontent.com/assets/4827522/12022698/97bad922-ad4f-11e5-9345-ac95e857325c.png)

coreycaitlin:
:heart_eyes: I like that!

coreycaitlin:
My conversation with the content critique group raised a bunch of useful questions. Putting them here for visibility:

- For user research: how current do users perceive the site to be? What expectations do they have about currency/frequency of updates?
- How frequently do we intend to update the "What's New" page? Will this page age and then go away, or will we update it every 3-6 months? If the latter, we need to document that plan and include a date stamp on the page somewhere.
- How else could we use the homepage to set user expectations about currency / continuous improvement?
- The "in progress" / "in development" phrasing may imply that users should not expect the site to be "final" — and might imply that they should expect bugs, incomplete sections, or other issues.
- "In progress" / "in development" language may also imply a pace of work being done (or a pending launch) that's no longer accurate.
- Where would users look to find out when the data was last updated, or when to expect another update?
- Option: include a [changelog](http://keepachangelog.com/); [example](http://data.influenceexplorer.com/docs/changelog/)

coreycaitlin:
Possible interim language (to be revised based on user testing):

> [See what's new](https://useiti.doi.gov/about/whats-new/) or [help us improve this site](https://ethn.io/38933).

This makes no statements about the site's finality or status, but clearly communicates that _some_ stuff is new and we're planning to keep changing things for the better.

coreycaitlin:
@meiqimichelle ^^ how does this sound? Okay to surface that feedback link a little higher and change the "in development" language?

meiqimichelle:
That sounds great! Really great feedback from the group, too!

coreycaitlin:
[...]

PR #1158, opened 2015-12-18
Site sprucing

gemfarmer:
This PR does a few things:

* removes "This site is under development" and link to `whats-new/` from the top banner #1157
* addresses a scrollbar issue that Michelle was seeing in chrome and firefox on the explore filter
* rearranged the selector on the case-studies subpages. Previously, the map for the page and selector were on the bottom of all the text in mobile. I moved these to below the introduction. @meiqimichelle could you take a look and see if you think this is a good place for these?


meiqimichelle:
@gemfarmer the banner info shouldn't be removed, unless you heard otherwise. They are there to show that the site is under constant improvement, and we'll show the latest on that what's new page.

gemfarmer:
@meiqimichelle Ok. reverted those changes. Maybe eventually the language should change slightly, to make it sound like it is being maintained and not developed?

meiqimichelle:
@gemfarmer Yes, good point on the language. We can run that by @coreycaitlin and see what she thinks. And I checked the case studies mobile solution. I think that's perfect for now! I'm not sure if there's a better thing for the long term, but I'd rather do some site testing first anyways, and it solves the no-way-to-navigate-on-mobile problem.

gemfarmer:
:boom: :mouse:

PR #1162, opened 2015-12-18
HTML fixes and tests

shawnbot:
This PR brings in [html-proofer](https://github.com/gjtorikian/html-proofer) to check our HTML and fixes most of the reported errors. You can run it with:

```sh
# update your ruby deps
bundle
# run the test
npm run test-html
```

The only real code change here is in the two production-related Explore pages, where I had to modify the filter description sentence to list the product instead of the commodity. In both cases this was always saying "All", but now it works as expected when you select a product.

The remaining reported errors are definitely non-critical:

* The link to the units selector on Explore > Jobs is invalid because I haven't added it yet.
* The JS docs (which are out of date) have some bad anchor hrefs.
* The styleguide has some missing `alt` attributes and bad anchor hrefs.


gemfarmer:
I'll review this by the end of the day :)

gemfarmer:
This looks good. Thanks for adding the proofer, @shawnbot !

Merging...

Issue #1127, opened 2015-12-15
Add more terms to the glossary


coreycaitlin:
Splitting this off from #908. Terms to add:

More words to add:

- [x] Independent Administrator (should be able to pull from Exec Summ)
- [ ] ~~ad valorem~~
- [x] GDP (see p. 61 of the Executive Summary)
- [x] locatable?
- [x] Fiscal year

From @gemfarmer:

Federal Production and All Lands production have some units that should be defined for a variety of products.

* [x] tons (also short tons)
* [x] metric tons
* [x] barrels (any certain number of barrels e.g. bbl, thousand barrels could fall under this)
* [ ] ~~cubic feet (will be either Mcf or MMcf. thousand and million, respectively)~~
* [x] megawatt hours (Mwh)
* [ ] ~~gallons (any variation on)? too basic?~~
* [ ] ~~pounds? too basic?~~
* [ ] ~~tickets (to measure quartz crystal)~~

meiqimichelle:
re: the units, are we defining other units in the glossary? We could define those sorts of things in data documentation instead.

gemfarmer:
@meiqimichelle Units seem like perfect a candidate for the glossary. Especially for the abbreviated ones. I could see someone wanting to quickly reference what a `Mwh` or `bbl` without having to check out and sift through all the documentation. 

That being said, I think it should be in the data documentation as well.

coreycaitlin:
For "locatable minerals," most definitions are pretty circular, but even a slightly circular version might be better than nothing. @meiqimichelle, how's this?

> Locatable minerals are minerals that may be “located” and obtained by filing a mining claim. Locatable minerals include gold, silver, copper, lead, and many other metallic and nonmetallic minerals.

meiqimichelle:
@coreycaitlin That makes sense to me, and I feel like it explains the point. I don't mind that its is semi-circular ;)

@gemfarmer yeah, good point

coreycaitlin:
Closing this out. Let's create a new issue if we come up with more terms that need defining.

Issue #531, opened 2015-07-16
Content: county narratives are missing sidebars


meiqimichelle:
There is extra information in the county narrative write-ups in the form of pull-quotes. These need to be transitioned to the online version in an elegant way.

meiqimichelle:
@coreycaitlin I'm not sure this is still relevant, or if you ended up addressing this elsewhere, or if the final version of the county narrative content didn't have important sidebars. Please close if this issue isn't useful.

coreycaitlin:
- [ ] Campbell County: _mill levy_ definition
- [ ] DeSoto Parish: _millage tax_ definition
- [ ] Elko/Eureka: ton vs. metric ton clarification
- [ ] Greenlee County: ton vs. metric ton clarification
- [ ] Humboldt/Lander: ton vs. metric ton; _mineral resource potential_ definition; _proved reserves_ definition
- [ ] Kern County: U.S. oil barrel definition
- [ ] Marquette County: ton vs. metric ton clarification
- [ ] North Slope: _mill rate_ definition; _reclamation_ definition
- [ ] Pima County: ton vs. metric ton clarification
- [ ] St. Louis County: ton vs. metric ton clarification
- [ ] West Virginia: ton vs. metric ton clarification; "A commodity's sale price does not always equate to its market price."

Hey! We found some good, concise unit definitions. I'm going to pull some of these into the glossary, and will use this checklist to note when I've set up the glossary links in the case studies.

coreycaitlin:
@Isabelle1512 I was just going to add the `millage tax` definition to the footnotes for DeSoto Parish, but I don't entirely understand it. The sidebar reads:

> A millage tax is the amount of tax payable per dollar on the assessed value of a property. Each mil is worth one-tenth of a cent, or $0.001.

And the text reads:

>The DeSoto Parish Emergency Medical Services (EMS) Department formed in 2001, funded by a four percent millage tax.

Could you help me clarify what a "mil" is, and how these two things fit together?

coreycaitlin:
A second checklist for adding glossary links to the case studies themselves:

- [] Campbell County: _mill levy_ definition
- [ ] DeSoto Parish: _millage tax_ definition
- [] Elko/Eureka: ton vs. metric ton clarification
- [] Greenlee County: ton vs. metric ton clarification
- [] Humboldt/Lander: ton vs. metric ton; _mineral resource potential_ definition; _proved reserves_ definition
- [] Kern County: U.S. oil barrel definition
[...]

PR #1169, opened 2015-12-24
4 more glossary terms

coreycaitlin:
This PR adds four terms to the glossary: tons, metric tons, locatable minerals, and mill levy.


coreycaitlin:
Ah! This helps with #1127 and #531.

Issue #996, opened 2015-12-07
JavaScript optimization


shawnbot:
Here's our pre-launch checklist:

- [ ] measure JS file size and load time on some pages to determine whether it's worth optimizing
- [x] bring in [WebPack](http://webpack.github.io/) and add `build` and `watch` scripts to `package.json`
- [x] bundle up sitewide (shared) JavaScript (`js/eiti.js`, et al)
- [ ] bundle page-specific scripts (`js/pages/*.js`)

gemfarmer:
@shawnbot I'm going to start working on this.

PR #1174, opened 2015-12-29
JS optimization

gemfarmer:
For #996 
What I have done:
* Created files that consolidate scripts on a given page.
* `main`, `homepage`, `explore`, and `narrative`
* these files will be compressed if they are pushed to `staging` or `master` branches 
--> Issue here. The `webpack` command in `_plugins/build.rb` is taking forever to build on the production server, so I'm wondering if it might be quicker/simpler to remember to add the webpack minification to the staging and master deployment workflow? This is less than ideal, but so far Federalist has taken an hour to build the site with the minification script running :(. @shawnbot Is there a particular reason why you wanted to use webpack and not gulp or browserify? 

Notes: 
* For the explore pages I am leaving the page-specific scripts and the `eiti.explore.js` scripts unconsolidated and unminified. These should eventually be encorporated into the `explore.min.js` script on that page, but for now shouldn't grossly affect page performance.
* A similar issue exists with the numerous references to `accordion.js`
* Many of the hound "errors" are intentional, namely the `'x' is not defined` errors
* As it stands, before merging to staging/ production, someone needs to run `export NODE_ENV=prod && webpack` to ensure that the files are minified. There is a build file: `_plugins/build.rb` that was intended to change the value of `NODE_ENV` based on the branch, but it does not appear to be either 1.) Building the site OR 2.) running the `webpack` command that would conditionally minify files. I have temporarily commented out this file.


gemfarmer:
Ok. Making list of remaining optimizations here: https://github.com/18F/doi-extractives-data/issues/1199

I think that removing lodash will probably be a small fix. There weren't that many times that it was used on the client.

jQuery might be harder/ take more time to remove because it is used more.

gemfarmer:
@shawnbot I got that. I just thought it might also be nice to remove jQuery. Then again, it isn't that big and is touching quite a few things, so maybe not now. _checking your new builds now_

gemfarmer:
The automatic build doesn't appear to be working for me. 
To test if the build is minifying based on the branch, I wanted to push unminified scripts with the hope that a new jekyll build would overwrite them.
 I set the NODE_ENV: `export NODE_ENV=dev` and built the scripts: `webpack`. The script was pushed to github unminified.
It doesn't appear that Jekyll is rebuilding/overwriting this build, because it is still unminified: https://federalist.18f.gov/preview/18F/doi-extractives-data/js-optimization/js/lib/main.min.js

Am I going about this the wrong way, @shawnbot ?

gemfarmer:
@shawnbot I saw you had pinged Dave Cole about build logs. Did you determine the issue?

gemfarmer:
@shawnbot good point. We should merge this then!

PR #1196, opened 2016-01-05
Added several glossary terms and glossary links in case studies

coreycaitlin:
Adds a few glossary terms from the case study sidebars (#531), plus links in the case studies. Also adds a couple of footnotes to case studies. See also: #1127 

Also, this is a test to see if everything works...


coreycaitlin:
@gemfarmer or @meiqimichelle — I think this wraps #531, and should be ready to go when someone has a chance to review.

PR #1217, opened 2016-01-20
Fixes maps and charts in Firefox and Safari

gemfarmer:
This PR addresses the bug that was preventing maps and charts to display on Firefox. 

This appears to be caused by `use-strict` being added to `eiti.js`, and leaving this error:
```
Uncaught Strict mode does not allow function declarations in a lexically nested statement.
```

Because having the document `use-strict` is a good thing, I edited the function declaration of `CustomEvent` [~here~](https://github.com/18F/doi-extractives-data/compare/dev...ffx-safari#diff-c390f5f37e805b47f623b379698d75a8R635)

@shawnbot can you review to make sure that this isn't going to result in issues that I'm not foreseeing?

I also added a [new command](https://github.com/18F/doi-extractives-data/pull/1217/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R45) to address an issue that I was having – being able to run `weback --watch` locally.


js/eiti.js
```diff
@@ -632,20 +632,20 @@
    * <https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent>
    */
   (function () {
+    function CustomEvent(event, params) {
```
shawnbot: This shouldn't be necessary. Also, we should consider using a packaged CustomEvent polyfill now like [this one](https://www.npmjs.com/package/custom-event) now that we're bundling.

package.json
```diff
@@ -42,6 +42,7 @@
     "test": "mocha",
     "test-html": "bundle exec htmlproof --disable-external _site",
     "webpack": "webpack",
+    "webpack-watch": "webpack --watch",
     "watch": "watchy -w '_sass' -- npm run build-styleguide",
```
shawnbot: Maybe call this script `styleguide-watch` now for consistency?

shawnbot: Or rather, `watch-styleguide` and `watch-webpack` for consistency with the `build-` ones?

js/eiti.js
shawnbot: And if we're not going to use a module, then I would prefer to keep the try/catch method and simply set `window.CustomEvent = function(...)` to make it work in strict mode.

js/eiti.js
```diff
@@ -635,17 +635,6 @@
     try {
       var e = new CustomEvent('foo'); // jshint ignore:line
     } catch (error) {
-      function CustomEvent(event, params) {
```
gemfarmer: @shawnbot is this what you had in mind? I removed this function all together and required the `custom-event` module [here](https://github.com/18F/doi-extractives-data/pull/1217/files#diff-9d87fecdcf0f321286d9d33562860b0bR21)

gemfarmer:
@shawnbot I didn't see [your comment above ^^](https://github.com/18F/doi-extractives-data/pull/1217#issuecomment-173708638). You are welcome to take a stab at it. I made some edits based on your suggestion. Mostly by adding the `custom-event` module.

js/eiti.js
shawnbot: This also fails in strict mode because the function is hoisted out of the `catch` statement. Let's just use [this polyfill](https://www.npmjs.com/package/custom-event).

js/src/main.js
```diff
@@ -18,8 +18,10 @@
   // EITI
   // TODO: require explicit dependencies (such as d3 and queue) in eiti.js
   // rather than here
+  require('custom-event');
```
shawnbot: I think this needs to be written:

```js
var CustomEvent = require('custom-event');
```

because the polyfill doesn't explicitly set `globals.CustomEvent` (or `window.CustomEvent`).

gemfarmer:
Is this ready to merge, @shawnbot ?

meiqimichelle:
_is excited_

gemfarmer:
@shawnbot some of the case studies maps maps aren't loading. Actually, only the first one, boone, logan, mingo is loading.

Other than that things look good. I'm investigating now, but you have an idea as to why one would load, but not the others, let me know.

gemfarmer:
@shawnbot Ok this should be good to go except it looks like I reverted d3 to a previous version. I probably need to run `npm install --dev` and rebuild.

gemfarmer:
@shawnbot @meiqimichelle ready for final review!

_sass/blocks/_all.scss
```diff
@@ -11,6 +11,7 @@
 // Styleguide blocks.case-studies
 @import 'case-studies/_intro';
 @import 'case-studies/_content';
+@import 'case-studies/_map';
```
shawnbot: We should fix these before merging.

_sass/blocks/case-studies/_map.scss
```diff
@@ -0,0 +1,19 @@
...
+  stroke: $mid-gray;
+  stroke-width: 0.5px;
+  fill: transparent;
+
+  path {
+    vector-effect: non-scaling-stroke;
+  }
+
+  g {
+	  &:first-child {
```
shawnbot: whitespace :trollface:

gemfarmer:
Yes, but I didn't think it got caught in a loop. But it does look like your commits are duplicated

gemfarmer:
@shawnbot Ok. It should be up to date with `dev`. Looking for your review :)

PR #1221, opened 2016-01-22
Rename offshore areas

### LINKED ISSUE #1139: Rename offshore regions
This started as #1108, but when I changed the data files things stopped working. The reason is lame: the `name` column values in `_data/offshore_regions.csv` (which gets turned into `site.data.offshore_regions` in Jekyll templates) are meaningful in that _some_ of the data sets contain uppercase region names ("Gulf") as opposed to the more id-like ones like "gulf". That's why we need the `REGION_ID_NAME` JS dictionaries in some of the explore visualizations: to map between names in the data and the ids of their corresponding geographic features.

The real fix here is to ensure that the all of the data that the visualizations are loading has ids that match the geographic ids. The `bin/abbr-state.js` script in our data directory does something similar for states (replacing "Alabama" with "AL", for instance), and it should be easy to whip up an offshore region equivalent. Then we'd strip out references to `REGION_ID_NAME` in all of the `applyFilter()` functions in our explore sections, ensure that everything works, and rock on.

gemfarmer:
Fixes #1139 

This fixes the offshore naming problem that we were seeing, namely the confusion between `Alaska` and `Offshore Alaska`. It fixes the issue in the page-specific scripts, so it will need to be updated if we refactor page-specific scripts.

Previews for the three pages where offshore regions are present:
* [federal revenue by location](https://federalist.18f.gov/preview/18F/doi-extractives-data/naming-offshore/explore/federal-revenue-by-location/)
* [federal production](https://federalist.18f.gov/preview/18F/doi-extractives-data/naming-offshore/explore/federal-production/)
* [all lands production](https://federalist.18f.gov/preview/18F/doi-extractives-data/naming-offshore/explore/all-lands-production/)

Ready for review!


meiqimichelle:
We should have some consistency in how we name offshore areas. Right now we have:
Offshore Alaska
Pacific
Atlantic
Gulf

@coreycaitlin , what do you think these should be? Maybe -->
Offshore Alaska
Pacific Ocean
Atlantic Ocean
Gulf of Mexico

coreycaitlin:
I like those! "Offshore Alaska" and "Gulf of Mexico" help clarify things (since Alaska also has a Gulf), and adding the word "Ocean" makes it more human language.

meiqimichelle:
Back to you, then @gemfarmer -- can we also adjust the other offshore names here?

gemfarmer:
@meiqimichelle @coreycaitlin Indeed we can. I'll have this ready in a bit for re-review

gemfarmer:
@meiqimichelle @corey. New regions added. Ready for review again!

meiqimichelle:
Looks great! :mouse: :boom:

PR #1236, opened 2016-01-29
added term and links for IA and Mwh

coreycaitlin:
Finally, a glossary item for Independent Administrator! Also Mwh. Let me know if either needs work. Related to #1127.


meiqimichelle:
:tada:

PR #1269, opened 2016-02-11
Glossary term opens on click

### LINKED ISSUE #1251: Glossary: clicking on item doesn't open definition
When you click on a glossary icon, the glossary window opens, and filters the glossary definitions, but does not open that definition.

Previously the glossary did this, was this an active or an incidental change, @meiqimichelle ?

gemfarmer:
Fixes #1251 

Clicking on a glossary term reference now opens the definition of that item when it opens the glossary.

[>>PREVIEW<<](https://federalist.18f.gov/preview/18F/doi-extractives-data/glossary-click)

The primary goal of this PR was quite simple, and was fixed with the first commit (6177ac1). There was an easier way to do this, so most of this PR is actually a refinement of the glossary markup, which includes adding the existing accordion component instead of using glossary-specific code.

The changes on the How it Works and Explore landing pages, and diving graphics pages are markup alterations needed after refactoring the accordion js. It was previously a bit hacky, this is better.

Ready for review! Please take a look at the pages mentioned above as well. I want to make sure I didn't miss anything! I was pretty thorough, so this shouldn't be an issue. :)

### DIFF
--- a/_explore/default.md
+++ b/_explore/default.md
@@ -24,7 +24,7 @@ permalink: /explore/
 	</div>
 </section>
 
-<section accordion class="container-outer landing-wrapper">
+<section accordion accordion-desktop="false" class="container-outer landing-wrapper">
 
   <section class="container">
     <a id="revenue" class="link-no_under"><h3 class="landing-section_category">Revenue data</h3></a>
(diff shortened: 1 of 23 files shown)


meiqimichelle:
:confetti_ball: :tada: :mouse: :boom:

Issue #1268, opened 2016-02-11
Add 2014 'by company' data to site


meiqimichelle:
We've gotten another year of [federal revenues by company](https://useiti.doi.gov/explore/federal-revenue-by-company/#) data. We need to look at the data file and see if it is in the same format. If no, we either need to have it re-created in the same format, or have Judy move our coding priorities to adjusting the current visualization to fit the new format.

meiqimichelle:
From @mentastc : `so previously we had lumped companies below the threshold or one mine as one category, ONRR still did this but the also pulled those companies up into the data set and put withheld where there otherwise would have been a number.`

meiqimichelle:
Shawn took a look at this and is easily able to *switch out* the datasets (we can stop showing 2013 data and show 2014 data instead). To update that page to show both datasets, he says that this will be decent chunk of work because the datasets are relatively large and so we'd want to look into a better way work with the data so that it loads in a useful amount of time for the user.

@usEITI Do you want us to switch out the dataset that's shown on the site now to the new one, and/or do you want us to wait until we can get the multi-year functionality working?

gemfarmer:
@shawnbot: I was thinking this problem could be approached differently by creating an entirely separate page for the 2014 dataset and displayed in tabs. The benefit in my mind would be that we wouldn't have to merge the datasets and we wouldn't have to make a interface for a year filter. It also might be more maintainable as new datasets are added.

shawnbot:
That sounds like a great solution to me, @gemfarmer! Are you thinking _entirely_ different URLs, as in `/explore/federal-revenue-by-company/2014/`?

gemfarmer:
@shawnbot That is what I was envisioning, with a UI similar to what we did with [onshore/offshore diving graphics pages](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/how-it-works/onshore-oil-gas/)

shawnbot:
I've updated the current status of this in #1270, along with a plea to get @ericronne's :eyes: on the tabs.

meiqimichelle:
Waiting implementation of @ericronne 's design direction, which is noted in #1270

meiqimichelle:
[...]

PR #1270, opened 2016-02-11
[WIP] Update Federal Revenue by Company data

shawnbot:
[:eyes: Preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/company-data/explore/federal-revenue-by-company/)

This adds the 2014 data for Federal Revenue by Company in #1268. TODO:

- [x] Resolve styling for the year tabs. Help us, @ericronne!

- [ ] Turn company revenue years into a collection?

  Right now the company revenue years (2013 and 2014 for now, but presumably a longer list in the future) live in `site.company_revenue_years` as defined in the `_config.yml`, but this is not the right place for them. They should live in a separate Jekyll collection, which takes them out of the `_explore` directory but makes it easier to add years in the future.

- [ ] Address withheld figures:

  From what I can tell, there weren't any withheld figures in the previous data set, so we never bothered to accommodate them. There _are_ withheld figures in the new data, so as a stopgap measure I've excluded them from the data that the site uses. Let me know if we need to show the withheld figures on the site and we'll figure out how to do that correctly.


ericronne:
@shawnbot et al, something like this?

image

ericronne:
(Still messing with other design options fyi...)

gemfarmer:
@ericronne definitely less obtrusive than a set of tabs.

ericronne:
I really like the look of that, but it looks kind of like a slider (mostly because we have numbers as the labels). Also it can't extend beyond two, so we would have to design something different if we wanted to offer more years down the road. Here are some options. I lean toward **C** …

image

ericronne:
Let's go with B, if all agree

meiqimichelle:
:+1:

Issue #1278, opened 2016-02-17
Code: Explore > Reconciliation, summary chart, v2


meiqimichelle:
From #1252, we've refined the summary chart on the reconciliation page. The latest design is:

![image](https://cloud.githubusercontent.com/assets/9259185/13014475/9ade8eb6-d179-11e5-85be-68c19addb193.png)

For now, we don't need to click-to-sort, just the updated info in the chart.

PR #1284, opened 2016-02-18
New Reconciliation summary chart

gemfarmer:
For #1278 .

The PR removes the old chart html and js and adds the new one based on [this mock](https://cloud.githubusercontent.com/assets/9259185/13014475/9ade8eb6-d179-11e5-85be-68c19addb193.png)

- started adding flowchart with @meiqimichelle 

@ericronne [PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/reconciliation-new-summary/explore/reconciliation/)


ericronne:
Awesome @gemfarmer! Just need new copy for this line @meiqimichelle ...
image

meiqimichelle:
@ericronne @gemfarmer Do you think it still needs a sub-head now, or should we make the headings for 'threshold' and 'variance floor' glossary-clickable instead?

ericronne:
@meiqimichelle @gemfarmer if we kept the subhead, in theory we could preserve the table column headers clickable for sorting (eventually at least)

gemfarmer:
@ericronne @meiqimichelle I think the subhead is nice, especially if it allows the headers to be clickable.

How about something like:

> To be considered "material", a variance must exceed both the threshold and variance floor

ericronne:
Should the subhead serve to explain that what is considered a material variance differs for various types of revenue? EG "The criteria for what is considered a material variance differs depending on the type of revenue."

Kind of wordy

gemfarmer:
@meiqimichelle the main pain points seem to be 
- determining how long the "long" stems need to be as the dialog heights change with screen size. Haven't thought to much about how to do this with css (wouldn't be too bad with js)
- determining the horizontal location of the stems with different width bubbles. I'm thinking that the way to approach this is to probably do some math to determine where the middle of each of the 12 columns is and create a class that will create the stem and down arrow at the middle of a declared column. Might be easier to talk over our pairing session tomorrow.

gemfarmer:
@ericronne ready for review!

@meiqimichelle I ended up adding js for that one extension line, but leaving a long string of media queries as a no-js fallback.

gemfarmer:
I fixed the issue that was preventing the searchable by-company data from showing up.

Issue #1233, opened 2016-01-28
Explore > Reconciliation: Explain material variances


coreycaitlin:
On the reconciliation page, it would be great to be able to click on each red number and see the explanation for that variance.

meiqimichelle:
Also, we should change the red to bold (or another indication of special-ness).

@coreycaitlin , @splatts -- do we have a list of the explanations for the variances somewhere?

splatts:
@meiqimichelle @coreycaitlin I sent Corey the appendix, which our Reconciliation team mentioned had the explanation of variances. @coreycaitlin did that give you what you needed? I know we discussed last week, so lmk if you need anything else!

coreycaitlin:
Yep, I've got that document — just need to work through it and pull out (and possibly edit?) the explanations.

coreycaitlin:
Now working through the explanations to come up with more concise versions of each.

Quite a few of the explanations are variations on the following type:

> $54,318,974 was reported by Chevron Corporation as paid in December 2012 but was recorded by ONRR in 2013.

Does that phrasing seem accurate, @splatts?

usEITI:
Correct.  We chose as an MSG to report on a calendar year basis.  When a
company submits a payment, there is a brief lag time when the ONRR system
records the payment as received. Hence the December / January timing
discrepancy.

On Mon, Feb 1, 2016 at 8:20 PM, Corey Mahoney <notifications@github.com>
wrote:

> Now working through the explanations to come up with more concise versions
> of each.
>
> Quite a few of the explanations are variations on the following type:
>
> $54,318,974 was reported by Chevron Corporation as paid in December 2012
> but was recorded by ONRR in 2013.
>
> Does that phrasing seem accurate, @splatts <https://github.com/splatts>?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/doi-extractives-data/issues/1233#issuecomment-178293473>
> .
>



-- 
*Judy Wilson*
*Program Manager USEITI Secretariat *
*Office of Natural Resources Revenue*

*judith.wilson@onrr.gov <judith.wilson@onrr.gov>*
*202-208-4410*

gemfarmer:
@coreycaitlin I just merged a branch with dummy footnotes added. Tge references are generated, so you only need to worry about the footnote text :)

meiqimichelle:
[...]

PR #1287, opened 2016-02-22
revised material variance footnotes to be clearer

coreycaitlin:
Rewrote the footnotes explaining material variances to clarify in response to user feedback, per #1233.


meiqimichelle:
Reviewed, looks great! :boom: :mouse: :confetti_ball:

Issue #1249, opened 2016-02-04
Visual Bug: de-xmas the home-page disbursements graphic


ericronne:
I removed the line rules, should help!
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaY2ZDMExlQUpRRG8/view?usp=sharing

@gemfarmer

PR #1303, opened 2016-03-03
Update Disbursements image

gemfarmer:
For #1249 

@ericronne Not sure why this didn't make it into the last vizbug update, but here it is:

:sunglasses: [PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/xmas-image/)


ericronne:
swoon! ty

Issue #1312, opened 2016-03-08
Company Revenue subtotals are incorrect


shawnbot:
The per-company totals on the [live site](https://useiti.doi.gov/explore/federal-revenue-by-company/2013/) are wrong. This looks like a summing issue that I introduced with [this change](https://github.com/18F/doi-extractives-data/commit/b1a9b9d329b6769f538b1fe82fc7716b6d98d892#diff-271a8b374ac426c5019416957329cab1L171).

![image](https://cloud.githubusercontent.com/assets/113896/13612856/6f69c274-e51e-11e5-947f-35a9bc83e553.png)

meiqimichelle:
Did this get fixed with those pushes to production earlier this week? If yes, did we make sure to merge them back into `dev` ?

shawnbot:
Merged staging to dev, and we're good!

PR #1313, opened 2016-03-08
Fix company subtotals

shawnbot:
[:sunglasses: preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/fix-company-subtotals/explore/federal-revenue-by-company/2015/)

This hotfix addresses #1312 by restoring the summing logic for per-company totals to its original form. I assumed (incorrectly) that there was only one row in the data for each row in these tables.

I also fixed a text wrapping issue with long negative values, which you can see [here, before](https://useiti.doi.gov/explore/federal-revenue-by-company/2015/#type=Other+Revenues):

image

and [after](https://federalist.18f.gov/preview/18F/doi-extractives-data/fix-company-subtotals/explore/federal-revenue-by-company/2015/#type=Other+Revenues):

image

### DIFF
--- a/_sass/blocks/explore/_regions-list.scss
+++ b/_sass/blocks/explore/_regions-list.scss
@@ -63,6 +63,7 @@
   td.value,
   th.value {
     text-align: right;
+    white-space: nowrap;
   }
 
   thead {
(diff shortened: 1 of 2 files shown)


gemfarmer:
@shawnbot Looks like they add up!

Issue #1325, opened 2016-03-15
Company Revenue doesn't display withheld values correctly


shawnbot:
Back when I did #1302 I introduced a summing bug that later led to #1312. The fix in #1313 broke the withheld values display, though. As you can see in [this example](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/federal-revenue-by-company/2015/#search=black+hills&commodity=Coal), the Black Hills Corp has withheld coal royalties, but we display `$0` instead of `Withheld`.

The fix here is to check for whether _all_ of the values (or just the one) to be summed up in a row are withheld, and display `Withheld` if that's the case; otherwise, sum up the numeric values and ignore the withheld ones. [BHP Billiton LTD](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/federal-revenue-by-company/2015/#search=bhp) is a good test case because it has [multiple royalties with only coal withheld](https://github.com/18F/doi-extractives-data/blob/dev/data/company/revenue/2015.tsv#L185-L187). This means that you should see a numeric value [with no revenue type selected](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/federal-revenue-by-company/2015/#search=bhp) (which is currently the case), but Coal should be listed as `Withheld` [with royalties selected](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/federal-revenue-by-company/2015/#search=bhp&type=Royalties) instead of `$0`.

shawnbot:
cc @mentastc for visibility

PR #1327, opened 2016-03-17
Fix withheld values displays in company subtotals

shawnbot:
This is a fix for #1325.


gemfarmer:
@shawnbot checked the logic and made sure the numbers add up! Looks good to me :)

PR #1328, opened 2016-03-18
508 compliance

gemfarmer:
I used the [HTML Code Sniffer](http://squizlabs.github.io/HTML_CodeSniffer/) to find accessibility issues with the site.

[:sunglasses: Preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/508-compliance/)

For this PR I addressed only what the sniffer found as 'errors'. There are still many 'warnings' and 'notices' that could be addressed with a more thorough sniff.

The errors that I corrected:
* replaced `<i class="fa ...."></i>` with `<icon class="fa ...."></icon>` for semantic reasons. This is the bulk of the PR diff
* Made header order make sense. We shouldn't skip from `<h1>`s to `<h3>`s, etc. To keep styling consistent, I used the class tag for the former heading: `<h2 class="h3">`. Fortunately, I think we were _pretty_ aware of this as we built the site, so this wasn't too bad.
* Made sure that the `<iframe>`s on the How it works pages had a `title` attr.
* Made sure all links had content that was hidden, but visible to screen readers.
* **This is the most important change to review**. [I changed the blue used site-wide](https://github.com/18F/doi-extractives-data/pull/1328/files#diff-c39df5e25beee216b7499d5ace3f7fa5L43) so that the contrast in the link below (on the about page), had a high enough contrast. If you don't like this blue, @ericronne, I could isolate the new blue to this link.
screen shot 2016-03-21 at 3 39 03 pm
* For the [eiti map](https://github.com/18F/doi-extractives-data/pull/1328/files#diff-a79ba3a65e14563d27aafe220d141f00L172) svgs, I moved the data being stored in the `xlink:href` attribute to the `xlink:data-href` attr and made the `href="#somestate"` as well as adding and `title` and `name` attributes for screen readers.
* For company revenue, the table tag structure for the data was confusing for accessibility. We had [`<th>`s nested in `<tbody>` tags](https://github.com/18F/doi-extractives-data/pull/1328/files#diff-271a8b374ac426c5019416957329cab1L214). Instead of nesting the `<th>`s in `<thead>`s, I decided to change the `<th>`s to `<td>`s

The only 'errors that I neglected to address was one issue with explore data map svgs. Within every eiti svg there is a list of anchor tags that contain a path defining the svg boudaries. At the end of it all is an attribute-less  `<a>` that contains the boundaries to all of the states. @shawnbot, I checked out eiti-map.js and am unclear how this is being generated. I don't think it is causing major accessibility issues, but in the interest of being thorough, I'm curious how it is being created.

```
<eiti-map>
  <svg>
    <g>
      ...
      <a data-href="#?region=MN"><path yadayada><title>Minnesota</title></path></a>
      ...
      <a data-href="#?region=WY"><path yadayada><title>Wyoming</title></path></a>
      <a><path yadayada></path></a>
    </g>
  </svg>
</eiti-map>
```

screen shot 2016-03-18 at 4 20 35 pm


gemfarmer:
Ok @meiqimichelle I extrapolated on the changes that were made!

ericronne:
@gemfarmer i had the type set to #065D83, which is a little darker than what you're using. I probably neglected to spec it for you. Fne as is, but feel free to switch it for the darker blue if so inspired. thx!

image

gemfarmer:
@ericronne ok. I can change it to `#065D83` if that is what you prefer. What about the rest of the site though? Do you think the blue being used, `#157bac`, should be changed to `#065D83` everywhere, or just in this one instance?

ericronne:
@gemfarmer No, just change it when over the blue. Over white it doesn't look blue enough. thx!

gemfarmer:
@ericronne ok!

meiqimichelle:
@gemfarmer This looks good to me! After you make the change @ericronne requested, do you want to merge, or wait for @shawnbot to respond to the question about the empty `<a>` ?

gemfarmer:
@shawnbot you are right!! 
@meiqimichelle I made the blue change only affect that one area on the about page (per @ericronne suggestion)

meiqimichelle:
@gemfarmer is that the only place that there's white-on-blue? I believe @ericronne 's feedback was that the change should impact all instances of that combo.

gemfarmer:
@meiqimichelle The combo was mid blue background with blue text. I didn't come across that anywhere else (or at least the sniffer didn't)

Issue #1206, opened 2016-01-14
404 Error page


gemfarmer:
Currently:

<img width="785" alt="screen shot 2016-01-14 at 2 25 03 pm" src="https://cloud.githubusercontent.com/assets/4803473/12338018/af8738ce-baca-11e5-8703-418bddc3a4f3.png">

gemfarmer:
Update: #1360 pushed the error pages to the live site, but do not show up when triggered, as I had previously hypothesized. Because the error pages work locally within a jekyll framework and only struggle when deployed via Federalist, I am assuming that this is an issue with Federalist. Will talk to Jeremia and see if there are successful instances of Federalist sites with error pages.

gemfarmer:
PR on Federalist to resolve this issue: https://github.com/18F/federalist-docker-build/pull/7
When it is merged, it **should** enable a 404 on Federalist.

PR #1340, opened 2016-03-24
Error pages

gemfarmer:
I finally made some error pages! For #1206 

I made a 404 and 500 page.

[:sunglasses: Federalist Preview](https://federalist.18f.gov/preview/18F/doi-extractives-data/error-pages/hurr)

@meiqimichelle @shawnbot @ericronne 

screen shot 2016-03-24 at 1 49 17 pm

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,5 +1,6 @@
 ---
 permalink: /404.html
+layout: error
+title: 404
+subtitle: Page not found
 ---
-
-YOOO
--- a/404.md
+++ b/404.md
@@ -1,0 +1,5 @@
+---
+permalink: /404.html
+---
+
+YOOO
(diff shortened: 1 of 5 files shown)


gemfarmer:
Ok. For some reason the 404s aren't working on Federalist. Will have to investigate

gemfarmer:
After doing some more research, my new hypothesis is that the error pages will work on the federalist `site` deployment on its default branch, but not on any of the `preview` urls. Unfortunately, we test this until we make another release. Fortunately, if this is not the case, the site will be exactly the same as it currently is – no risk!

@meiqimichelle do I have your blessing to merge this PR?

meiqimichelle:
@gemfarmer Yes, let's try it!

PR #1351, opened 2016-03-28
Allow custom Google Analytics tracking ID on Federalist

### LINKED ISSUE #1343: Use a different Google Analytics tracking ID on Federalist
Right now we've got a bunch of Federalist URLs mixed in with those of the production site, and while it's not skewing the numbers because the paths are different, it does make it harder to skim lists of paths and get a sense for popular landing pages. The ideal solution is to use a different GA tracking ID on Federalist than we do on the production site. We can do this by putting tracking ID into `_config.yml`, pulling it out of [the JavaScript](https://github.com/18F/doi-extractives-data/blob/294a5457e399dd4dc62c570cf78cfb564958d410/js/src/main.js#L46) (it should probably be inlined at the bottom of the default layout), then setting up our Federalist config like so:

```yaml
google_analytics:
  # XXX here is our Federalist tracking ID
  default: 'XXX'
  # only use this tracking ID on master (production)
  master: 'UA-48605964-8'
```

and in our template we would use:

```
{{ google_analytics[site.branch]|default:google_analytics.default }}
```

This works because `site.branch` gets set by Federalist at build time, so when `site.branch == 'master'` the output value would be the production tracking ID; otherwise `XXX`. And if we _only_ put the `google_analytics` setting in the Federalist config (as opposed to hard-coding it in `_config.yml`) then we can further refine the page template to just skip the GA bit altogether and not do any analytics when it's running locally.

shawnbot:
This fixes #1343 by moving the Google Analytics tracking ID to the config key `google_analytics`. It should be an object with keys for specific branches, and a `default` key to use if no branch-specific ID is provided. I haven't specced this out in `_config.yml` because it isn't necessary for local builds, but Federalist allows us to provide a custom config on top of ours, and it now looks like this:

```yaml
google_analytics:
  default: false
  master: 'UA-48605964-8'
```

Once this is merged, our Federalist builds on `dev` (and every other branch off dev from now on) will no longer send tracking to GA until we specify a new `default` ID. @meiqimichelle, I assume that you know how to get that, right? If not, this would be a good excuse for me to figure it out :)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,6 +1,9 @@
 title: USEITI
 
-description: USEITI is part of an international effort to promote open and accountable management of natural resources. This site provides data about how the U.S. manages extractive industries.
+description: >
+    USEITI is part of an international effort to promote open and accountable
+    management of natural resources. This site provides data about how the U.S.
+    manages extractive industries.
 
 # permanent url, for use in meta references
 url: https://useiti.doi.gov
--- a/_includes/google_analytics.html
+++ b/_includes/google_analytics.html
@@ -1,4 +1,3 @@
-{% if include.id %}
 <!-- Digital Analytics Program roll-up, see the data at https://analytics.usa.gov -->
 <script src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js" id="_fed_an_ua_tag"></script>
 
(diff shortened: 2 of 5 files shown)


gemfarmer:
_reviewing_

PR #1353, opened 2016-03-31
[WIP] Create SQLite database for data

shawnbot:
Okay, so @gemfarmer I'd like your eyes on this before I go much further, but I _think_ I'm on the right track with this. Here's what I've done so far:

1. I locked us in at Node ~~v0.12.7~~ v4.2.6 because it was the only version for which I could get all of our dependencies working. One workaround I found to npm dependency install hell was to do this:

  ```sh
  rm -rf node_modules
  npm cache clean
  npm i --dev
  ```

  So let me know if that works for you. I would've preferred to upgrade to Node v5.x because that version's npm uses a flat dependency tree, but SQLite support is [apparently busted](https://github.com/datanews/tables#install) in Node v5.

1. I brought in a forked version of [Tables](https://github.com/datanews/tables) that uses [tito](http://github.com/shawnbot/tito), which gives it TSV support out of the box. Tables provides a nice interface for importing tabular data into SQLite with [Sequelize](http://docs.sequelizejs.com/en/latest/) under the hood.

1. The `data/createdb.sh` script runs the Tables CLI tool once for each table to be imported, then runs some queries that create views for state and national views on top of the county data. You'll need to run this from the data directory:

  ```sh
  cd data
  ./createdb.sh
  ```

1. The `data/bin/query.js` script takes a SQL query as its only argument and spits out the results as TSV. This is much more useful for checking whether things are working than `sqlite3`'s output, and we can tweak the script to output other formats if need be.

  ```sh
  # also from the data directory
  ./bin/query.js "SELECT * FROM state_revenue WHERE state = 'WY' AND year = 2011"
  ```

Anyway, take a look, run these commands to make sure that they work on your setup, and let me know what you think! If this is good, I think we should be able to at least get all of the data into the DB in a couple of days.


gemfarmer:
Trying now

meiqimichelle:
@shawnbot Is this still relevant or was this overtaken by the state profile work? Thanks!

Issue #1347, opened 2016-03-28
Prototype: State profile pages


meiqimichelle:
This will come from #1334.

shawnbot:
The next step is to generate pages for each state that list out data for the following metrics:

* Federal revenue
* Federal production
* All lands production
* Jobs (economic impact)

This should provide us with some useful input for the wireframing process, since it will show us, for instance, how many products we will need to show under each section for a particular state or offshore area.

meiqimichelle:
[Notes doc](https://docs.google.com/document/d/1z31GwYr9g5ArS8uE_XV0LQEjP0WyIFvGXX_HWDmDXus/edit#)

meiqimichelle:
Newer, better start of wires [here](https://docs.google.com/document/d/1_l5WbEQZAOfO-oUD3P1t06G1OjA7jeYMvf2DLqm9ZYs/edit#heading=h.z2y0fwq4j1y5) (that's a google drive doc, limited access), based on Shawn's prototypes (example with Utah data [here](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/UT/) ).

shawnbot:
@meiqimichelle in the spirit of the chat we had at the IRL about when wireframing is appropriate, I'm considering renaming this to a _prototyping_ exercise. What say you?

coreycaitlin:
@shawnbot: running list of content issues to refine/revisit —

- "Geothermal" probably needs to be "Geothermal *energy*" to make sentences work
- "Hardrock" should be "hardrock *minerals*" to make sense
- NGL should be "natural gas liquid"
- Oil and Gas ~(Non-royalty)~: "produce" is the wrong verb here. If possible, could we specify that this revenue is to "secure rights or explore for" oil and gas?
- "co2" should be "carbon dioxide," and we should figure out what that means

coreycaitlin:
^^ [this fix](https://github.com/18F/doi-extractives-data/pull/1355/commits/d5bb0aa9465bf1b5ee82dcde2d520bd56616c337) starts fixing these! It'll take another pass through the page to make sure they're all making sense, but they're way better now, thanks to @shawnbot.

meiqimichelle:
@coreycaitlin should this issue be closed?

coreycaitlin:
👍  sounds good to me!

PR #1355, opened 2016-04-06
[WIP] State page prototype

shawnbot:
[:sunglasses: preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/CA/)

This is a _very_ work in progress PR that adds state pages with data from our new database workflow in #1353. The goal here is to determine whether we're on the right track in our wireframing exercise, #1347. 

There's no index of all of the state pages yet, but you can see them by changing the URL, for instance with [Utah](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/UT/) and [Texas](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/TX/). Right now we've only got revenue and production on federal land, but all lands production and jobs are on the way.


gemfarmer:
@shawnbot --> eee5880 adds Exports data. Not sure if we want to include more (namely rank), but this should be a start

meiqimichelle:
@shawnbot _it begins!_ I'm excited. As for the two years -- yep, not super interesting. I remember on our beta site back in the day we did the bubble chart -- and switching between the years turned out to be a pretty big yawn because the funds don't change much year to year. The story here is not likely in the year-over-year change.

gemfarmer:
@shawnbot didn't see this. Will run the commands and get back to you.

gemfarmer:
@shawnbot looks great _except_ I'm running into this error: `/bin/sh: ./node_modules/.bin/tito: No such file or directory` which is preventing all the tables from being added. I'm guessing you added `tito` to bin but didn't save it??

gemfarmer:
Did you add `nodemodules/.bin` to your `PATH`?

gemfarmer:
@shawnbot just took another stab at this. I originally installed the dev dependencies with node v4.2.6 instead of v5.0.0, and for whatever reason it didn't add `tito` to the `node_modules/.bin` directory. I tried again using v5.0.0 and all the correct dependencies were added and the commands run perfectly. 

Here is a sample from my diff `git diff origin/state-pages _data`. Basically the only difference I saw was slight adjustment to the yml syntax. Is that what you were hoping?

screen shot 2016-04-19 at 3 22 46 pm

gemfarmer:
@shawnbot 😍 nice! looks awesome

gemfarmer:
Closing in favor of #1651

Issue #1367, opened 2016-04-11
Markdownify pages


gemfarmer:
It would be lovely, both for the site's reusability, maintenance to make as many pages as possible 100% markdown. 

@coreycaitlin and I will pair to assemble a list of content pain points in the site. For now:

- [x] #1198 Glossary items
- [x] How it works pages
- [x] Narrative pages

gemfarmer:
* [ ] `about/report`
* [ ] `about/whats-new/`

meiqimichelle:
Starting review, lookin' good in general! Here's a few edits:
![markdownify-edits-1](https://cloud.githubusercontent.com/assets/4827522/14546629/9c0c5350-025b-11e6-9b49-4cd7fc10d709.png)

meiqimichelle:
Missed a list:
![screenshot 2016-04-14 16 17 07](https://cloud.githubusercontent.com/assets/4827522/14546733/61c0e476-025c-11e6-9025-bc475f66aaf7.png)

meiqimichelle:
Boldify sentence starters (I don't know what to call those, haha) :)
![screenshot 2016-04-14 16 18 42](https://cloud.githubusercontent.com/assets/4827522/14546769/92b31fea-025c-11e6-8ef4-c608eda3c009.png)

gemfarmer:
@meiqimichelle I think I've addressed all of these concerns!

Issue #1198, opened 2016-01-05
Simpler way of adding glossary links


coreycaitlin:
Right now, adding a link to the glossary requires pasting in the html: `<span class="term term-p" data-term="sample term" title="Click to define" tabindex="0">sample term<i class="icon-book"></i></span>`

Is there a way to make these simpler and more Markdown-flavored, ala footnotes in the case studies, where we use: `[^17]`?

One tricky thing about this is that the term doesn't always match the glossary term perfectly (singular v. plural, acronym v. full term, and so on), so we may need a way to connect them semi-manually.

gemfarmer:
@coreycaitlin as far as I know there isn't an easy, pre-baked way to do this.

I could imagine something that would look more or less like this being useful:
`A lot of people are talking about [extractive industries](^extractive industries) right now...`
which would allow for variations like
`Every [extractive industry](^extractive industries) needs to comply with...`

gemfarmer:
For my reference on making markdown extensions: 
* https://www.npmjs.com/package/markdown-flavor-maker

gemfarmer:
@coreycaitlin I think the time has come for me to work on this. What are your thoughts about the proposed alternative syntaxes mentioned above.

In your opinion, is `{% include term.html id="extractive industries" val="extractive industry" %}` much worse than [extractive industry](^extractive industries) ?

coreycaitlin:
Sorry, this didn't make sense until I clicked through to see it on GitHub!

Hm, that syntax is definitely trickier (I'd still need it stashed for copy-pasting), but slightly easier than straight HTML. I'm kind of ambivalent on whether it's easier _enough_ to be worth implementing.

coreycaitlin:
I like that there's an easier default one for when the term does match up with the glossary, though. That seems compelling.

gemfarmer:
With #1368 this syntax has changed. There are 4 different variations:
1. The displayed text is exactly the same as the glossary term text. The looks like this: 

`{{ "text" | term }}`

2. The displayed text is different than the glossary term. This looks like: 

`{{ "text" | term:"different text" }}`

3. The displayed text is exactly the same as the glossary term text, AND is directly followed by punctuation. This looks like:

`{{ "text" | term_end }}`.

4.  The displayed text is different than the glossary term, AND is directly followed by punctuation . This looks like: 

`{{ "text" | term_end:"different text" }}`

meiqimichelle:
We did this! Closing.

PR #1368, opened 2016-04-13
Markdownify

gemfarmer:
[🌻 PREVIEW 🐦 ](https://federalist.18f.gov/preview/18F/doi-extractives-data/markdownify)

I have started converting some of the pages to markdown, starting with default list styling and layout reformatting.

Checklist from #1367:

Pages
------
* [x] ⭐ `about/report`
* [x] ⭐ `about/whats-new/`
* [ ] `how-it works` steps
* [ ] `how-it-works` laws and regulations
* [ ] `how-it-works` natural resource pages
* [ ] How it works – Federal agencies @meiqimichelle is this in about???
* [ ] Case studies
* [x] ⭐ Downloads subpages (not `downloads/`
* [x] ⭐ Simplify error page layout. I'm not sure why I did it so poorly before.

Items:
------
- [x] ⭐ update default list styling. styles needed to accommodate new default in the following places: 
  - [x] ⭐ footer items
  - [x] ⭐ glossary
  - [x] ⭐ how it works steps
  - [x] ⭐ downloads subpages
- [x] make new `content` layout to handle about pages
- [x]  #1198 Glossary items: new term item documentation [here](https://github.com/18F/doi-extractives-data/issues/1198#issuecomment-210089014)

To figure out:
------
- [ ] Best way to add classes/custom styles to markdown elements


gemfarmer:
@meiqimichelle when either of you get a chance, could you preview the ⭐ 'd items in the list above to see if there have been any visual regressions. Also, @coreycaitlin feedback on the first wave of markdownification would be great! Wanna know if I'm headed in the wrong direction!

gemfarmer:
@shawnbot 
An example of what I am talking about would be `downloads/disbursements`. The first few paragraphs can't be fully turned into markdown because they are reliant on classes for styling, so I have been forced to hodge podge, like so: 
`<p class="case_studies_intro-para" markdown="1">The amount of money earned from extraction of natural resources on federal lands that is disbursed to various legislated funds. Our fund overview dataset is from the [Office of Natural Resources Revenue][~1], which is part of the Department of the Interior. In addition, we have more detailed datasets on two of the funds: [Land and Water Conservation Fund][~2] (LWCF) and [National Historic Preservation Act][~3] (NHPA).</p>`

gemfarmer:
This isn't the worst thing ever, but for people who are editing/maintaining the site, I'm sure it would be a little strange to see 80% markdown, with some odd html sprinkled here and there. Some of that might not be avoidable, but a lot of it could be if we were able to add classes somehow. OR build our css around elements??

gemfarmer:
@shawnbot by page data are you imagining that most of the site would be edited via jekyll frontmatter?

gemfarmer:
@shawnbot I'll watch the talk. If it's the one you mentioned in the last rally, then I've been meaning to get to it anyway. I think this shouldn't be a major problem with most pages. Especially if we have a class or id on the top level element that we can define page-specific styles within.

_plugins/glossary.rb
```diff
@@ -0,0 +1,52 @@
...
+      # <span class="term" data-term="word" title="Click to define" tabindex="0" style="margin-right:5px">
+      #   word
+      #   <icon class="icon-book"></icon>
+      # </span>
+      # {{ "word" | term:"alternative def" }} >>>
+      # <span class="term" data-term="alternative def" title="Click to define" tabindex="0" style="margin-right:5px">
+      #   word
+      #   <icon class="icon-book"></icon>
+      # </span>
+      unless term
```
shawnbot: I feel like it's better to use if/else here rather than `unless`. Also, wouldn't it be more DRY to just do:

```rb
term = term or str
return "..."
```

_plugins/glossary.rb
```diff
@@ -0,0 +1,74 @@
...
+      # </span>
+      # {{ "word" | term:"alternative def" }} >>>
+      # <span class="term"
+      #       data-term="alternative def"
+      #       title="Click to define"
+      #       tabindex="0" >
+      #   word
+      #   <icon class="icon-book"></icon>
+      # </span>
+      unless term
```
gemfarmer: Personally, I prefer a negative to a double positive, but if people have strong objections, speak now!

shawnbot: +1 to `if/else` instead of `unless/else`

pages/whats_new.md
```diff
@@ -0,0 +1,31 @@
...
+breadcrumb: About
+---
+
+The initial USEITI beta site launched in early 2015, offering visualizations that showed how natural resources generate revenue across the country and where the money goes.
+
+This site, which launched in December 2015, makes the data easier to use, improves the site structure, and provides context about the government's role in extractive industries.
+
+We've also added new information and datasets to explore:
+
+* **Production data:** Sales volumes and revenue have long been used to estimate how much of each resource was produced. Here, for the first time, you can find more precise data on what resources were produced on federal lands. [Explore federal production data.][~1]
```
shawnbot: I've never seen the `[~1]` convention used before. Is that a thing? My preference would be to give these more legible names, e.g.

```md
[Explore federal production data.][federal production]

[federal production]: url
```

gemfarmer: It isn't a convention that I'm aware of. I did it this way because having as short a string as possible seemed consistent with the goal of not interfering with the content text. Since we are already using the `[^1]` notation for footnotes, it seemed like using the same method with a different symbol made sense. @coreycaitlin from a content management perspective, does one method seem preferable?

_plugins/glossary.rb
gemfarmer: Changed!

gemfarmer:
Other than [this question raised by @shawnbot about markdown link notation](https://github.com/18F/doi-extractives-data/pull/1368#discussion_r59929508) this PR is ready for a mergin'. @coreycaitlin @meiqimichelle @RyanSibley I would be interested to you weigh in on markdown link notation

pages/whats_new.md
shawnbot: I think my biggest concern is that if we shuffle the content around later and the order of the references changes, then the numbers would cease to make sense. It's not that big of a deal, though, so don't let it hold up this PR!

gemfarmer:
Can this be merged?? @meiqimichelle

coreycaitlin:
Sorry @gemfarmer and @shawnbot — finally took a look at your Markdown-link question, at long last!

I actually like old-fashioned simple Markdown notation for links, because it keeps the URLs closer to the anchor text and avoids the temptation to create redundant link titles (which can be an accessibility annoyance).

For instance:

```
<p class="case_studies_intro-para" markdown="1">As part of fulfilling the international [EITI Standard](https://eiti.org/document/standard), the {{ "Independent Administrator" | term:"independent administrator (IA)" }} worked with the [Multi-Stakeholder Group](https://www.doi.gov/eiti/FACA) to develop an Executive Summary of the USEITI 2015 Report.</p>

This document, which you can download below, includes an overview of the contextual narrative and an account of the reporting and reconciliation process for the EITI Standard in the U.S.

However, many parts of the report are primarily available online. Explore this site to:

* [See maps and charts]({{ site.baseurl }}/explore/) of extractive industries data.
* Read [12 case studies]({{ site.baseurl }}/case-studies/) about how extractive industries have affected specific counties.
* [Download data]({{ site.baseurl }}/downloads/) and view data documentation.

In many parts of this site, you'll also find invitations to discuss and participate in USEITI actitivies both online and in person.
```

Issue #1366, opened 2016-04-08
508 issues


nickbristow:
https://useiti.doi.gov/how-it-works/

There appears to be a lot of buttons on the page with no value. They don't display on the page, but they are showing in this tool. http://wave.webaim.org/extension/

Either give the buttons values, or remove them from the page as they don't appear to be used?

@meiqimichelle

meiqimichelle:
@gemfarmer any thoughts on this one? off the top of my head, perhaps this is related to the accordions on mobile, but I haven't looked at the code, as you know, in a veeerrrryyyy long time :(

nickbristow:
@meiqimichelle @gemfarmer I think it is mobile buttons, so they need `<button value="blah">` or `<button aria-label="blah">`

gemfarmer:
@nickbristow are you referring to these? If so, yes these are mobile only buttons. Those accordions do need aria attributes (`aria-label`), but if they are hidden (`display: none;`) on desktop, why would that be an accessibility issue on desktop? Shouldn't they only _need_ labelling on mobile? 
<img width="822" alt="screen shot 2016-04-08 at 12 22 28 pm" src="https://cloud.githubusercontent.com/assets/4803473/14391963/979fc468-fd84-11e5-9b83-752383c0c6ba.png">

nickbristow:
@gemfarmer technically, yes. I just assume they aren't labeled in mobile atm?

gemfarmer:
@nickbristow They are not.

gemfarmer:
They do have a title though ```<button class="accordion-button" accordion-button title="Toggle for ownership"></button>```

nickbristow:
@gemfarmer title isn't great for compatibility among screen readers, especially on mobile.

gemfarmer:
:+1: ok. Good to know.

PR #1373, opened 2016-04-19
Accessibility wave 2

gemfarmer:
For #1366 

👀 [PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/accessibility-wave-two/) 👀 

I have made several small accessibility changes to all non-explore pages.
- Glossary (labels and buttons)
- Header, several labelling and button errors
- Updated the accordion so that it is using ARIA components. Still not using the [18F accordion](github.com/18F/accordion) because it uses jquery and underscore, and I like that EITI's accordion doesn't have dependencies. Might be worth contributing to that accordion so that it doesn't require dependencies.
- implementation of new accordion

@nickbristow I am still running into some errors/warnings with HTML_CodeSniffer and WAVE telling me that labels aren't properly set up on the search bar and glossary, but I believe that they are properly labelled. Any best guesses as to why this might be happening?

### DIFF
--- a/js/components/accordion.js
+++ b/js/components/accordion.js
@@ -19,7 +19,8 @@
 
     // determine truthiness an attribute of an element
     elemStatus: function (elem, attr) {
-      if (elem.getAttribute(attr) === 'true' || elem.getAttribute(attr) === true) {
+      if (elem.getAttribute(attr) === 'true' ||
+        elem.getAttribute(attr) === true) {
         return true;
       } else if (elem.getAttribute(attr) === 'false' ||
           elem.getAttribute(attr) ||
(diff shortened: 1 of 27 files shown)


js/components/accordion.js
```diff
@@ -34,22 +60,90 @@
...
+          accordionItems = accordion.querySelectorAll('[accordion-item]');
+          // check if node list is different than list of children
+          if (accordionItems.length !== accordion.children.length) {
+            this.forEach(accordion.children, function(index, item) {
+              // add [accordion-item] attribute to list items
+              if (item.tagName.toLowerCase() === 'li' &&
+                !item.hasAttribute('accordion-item')) {
+                item.setAttribute('accordion-item', true);
+              }
+            });
```
gemfarmer: I agree that making loops within loops is generally bad practice, but definitely the most straightforward approach when I am dealing with lists of lists.

gemfarmer:
@meiqimichelle ready for review!

meiqimichelle:
Something is not happy on the homepage :( Firefox and Chrome.
screenshot 2016-04-22 16 14 28

gemfarmer:
@meiqimichelle addressed tab panel issue with 4d5b29e

Issue #1381, opened 2016-04-27
Speed up development build process


gemfarmer:
With all the data in the `_data` directory, Jekyll builds are getting a bit crazy. I'm not sure how we would approach improving the Federalist build speed, but I think we could increase development builds by telling jekyll what to and not build/re-build. 

If we wanted to do development on : ```bundle exec jekyll serve --build-only 'maps'```

where `build-only` is a variable that determines which files get excluded. Or something like that

devhttps://jekyllrb.com/docs/configuration/

PR #1383, opened 2016-04-27
Add maps config to speed up build

gemfarmer:
@shawnbot for #1381 

I went about it differently, but I think it works! I made a `_maps.yml` configuration file by running 

`bundle exec jekyll build --config _maps.yml` with this config:

```
exclude:
  - lib
  - node_modules
  - package.json
  - show
  - test
  - styleguide-template
  - 'Gemfile*'
  - pages
  - img
  - _case-studies
  - _downloads
  - _explore
  - _how-it-works
  - _federal-revenue-by-company
  - _meta
  - _sass/blocks
  - _includes/case-studies
  - _includes/how-it-works
  - _includes/layout
  - _includes/glossary.html
  - _includes/hash_selector.html
```

and the build is waaay faster

screen shot 2016-04-27 at 12 21 31 pm


gemfarmer:
Ah ➕ . I was trying to figure out how to extend the config. If that works then A++

gemfarmer:
Yeah that still isn't great. :/. Definitely more that can be excluded. 

I also tried running `bundle exec jekyll build --config _maps.yml` without the duplicative material AND without extending the config and the page ran fine. If it isn't necessary, probably slightly faster

PR #1385, opened 2016-04-29
Revenue map and data

gemfarmer:
[🎱  P R E V I E W 🎱 ](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages-bh/states/WY/)

* adds data to `_data/county_revenue`
* adds revenue map to `state-page.html` layout
* adds data to `_data/federal_county_production`
* adds federal production maps to `state-page.html` layout

screen shot 2016-04-29 at 4 28 51 pm


gemfarmer:
@shawnbot I think these are good to go when you have a moment

PR #1392, opened 2016-05-09
Create land ownership stats

shawnbot:
This first pass creates site data for top-level federal land ownership stats, which we can refer to like this in templates:

```html
{{ site.data.land_stats[state_id].federal_acres | intcomma }} acres
of land in {{ state_name }}
({{ site.data.land_stats[state_id].federal_percent | percent }}%)
are federally owned.
```

If we need to get into agency-specific ownership stats, we can pull in the `data/land-stats/agency-stats.tsv` file as well. I left a note in `data/land-stats/README.md` about the source of the data: [the PDF](http://www.fas.org/sgp/crs/misc/R42346.pdf) from which I pulled the data was published by the Library of Congress's Congressional Research Service in December, 2014.


gemfarmer:
@shawnbot Maybe consider adding something like this to the README for land-stats?

---
## Data files
- [land-stats.tsv]({{ site.baseurl }}/data/land-stats.tsv): total and federal acreage for each state. 
This is the primary source data for Land Ownership
- [agency-stats.tsv]({{ site.baseurl }}/data/agency-stats.tsv): total acreage by agency for each state.
- [land_stats.yml]({{ site.baseurl }}/_data/land_stats.yml): a yml file to allow for easy templating. 
This is the data that HTML and Markdown pages are directly referencing.

## Transform scripts
- [transform.js]({{ site.baseurl }}/data/transform.js): script to convert `land-stats.tsv` to an SQL table, 
and subsequently to `land_stats.yml`

gemfarmer:
If that looks good, I can add and merge :)

gemfarmer:
@shawnbot I'm not sure I follow. I wasn't proposing that we move anything

gemfarmer:
ok cool  😎

Issue #1375, opened 2016-04-19
All lands production figures: multiply by 1,000?


shawnbot:
We're currently multiplying the all lands numbers by 1000 in our data script ([here](https://github.com/18F/doi-extractives-data/blob/dev/data/bin/group-all-production.js#L354) and [here](https://github.com/18F/doi-extractives-data/blob/dev/data/bin/group-all-production.js#L534)), but we should be doing this in the database (in #1353) so that the numbers we output in the HTML are accurate.

PR #1395, opened 2016-05-09
Production units

shawnbot:
[😎 Preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/production-units/states/CA/)

This PR adds a `units` field to all of the federal and all-lands production site data. The YAML is a bit more structured now (since we have to group by "product" at the level of product + units, as with multiple Geothermal products), and I've tried to make the two production data sets as similar as possible in the process. The ONRR (federal) and EIA (all lands) data each have their own directories now, each with its own README that explains what the files.

There are some things missing or wrong here that we should address in separate issues:

* For all lands renewables, the units are unknown. The templates simply list the word "units" for commodities without known units. (#1396)
* The multiples-of-1000 issue #1375 is unresolved, but I did work toward it. For instance, I changed the units for all lands gas from `MMcf` to `mcf` (as listed by the federal data) and multiplied the volumes by 1,000.


gemfarmer:
`data/bin/group-all-production.js` could probably be removed, along with a few abandoned transforms.

gemfarmer:
maybe we hold off on removing those in this PR though, it is a bit out of scope.

PR #1401, opened 2016-05-12
change automagic sentences to be clearer

coreycaitlin:
Pulled in a few more sentences from the gdoc and edited summary sentences to represent the data more precisely.

### DIFF
--- a/_layouts/state-page.html
+++ b/_layouts/state-page.html
@@ -190,7 +190,7 @@ layout: default
 
     [TABLE OR GRAPHIC SHOWING RATES FOR RELEVANT COMMODITIES?]
 
-    <h3>Revenue from federal lands</h3>
+    <h3>Revenue from federal land</h3>
 
     <p>{{ site.data.land_stats[state_id].federal_percent | percent }} percent of land in {{ state_name }} is owned by the federal government.</p>
     <p>In {{ year }}, companies paid the federal government a total of ${{ revenue_total | intcomma }} to extract natural resources on federal land (or lease federal land for that purpose).</p>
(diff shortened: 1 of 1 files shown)


_layouts/state-page.html
```diff
@@ -264,11 +264,10 @@
         {% assign product_name = product[0] %}
       {% endif %}
       <h3>{{ product_name }}</h3>
-      <p>{{ state_name }} generated
-        {% if revenue %}${{ revenue | intcomma }} in
-        {% else %}no{% endif %}
-        federal revenue from {{ product_name | downcase }} in {{ year }}.
-        Historically:</p>
+      <p>In {{ year }}, 
+        {% if revenue %}companies paid ${{ revenue | intcomma }} to extract{% else %}the federal government did not collect any revenue from extraction of{% endif %}
```
gemfarmer: Can these if/else statements be broken up on to multiple lines like above, or does it mess with the spacing?

_layouts/state-page.html
```diff
@@ -349,11 +350,9 @@
         {% assign units = product[1].units | downcase %}
 
         <h4>{{ product[1].name }} ({{ units | default: 'units' }})</h4>
-        <p>{{ state_name }} produced
-          {% if volume %}{{ volume | intcomma }}
+        <p>{% if volume %}{{ volume | intcomma }}
           <span class="units">{{ units | default: 'units' }}</span>
-          of {% else %}no{% endif %}
-          {{ product[1].name | downcase }} in {{ year }}. Historically:</p>
+          of {{ product[1].name | downcase }} were {% else %}No {{ product[1].name | downcase }} was {% endif %} produced on federal land in {{ state_name }} in {{ year }}.</p>
```
gemfarmer: Similarly, can this be broken up onto multiple lines? If it creates weird spacing we can leave it, but otherwise, something like this might keep things _slightly_ more legible

```
<p>
  {% if volume %}
    {{ volume | intcomma }}
    <span class="units">{{ units | default: 'units' }}</span>
    of {{ product[1].name | downcase }} were 
  {% else %}
    No {{ product[1].name | downcase }} was 
  {% endif %} 
  produced on federal land in {{ state_name }} in {{ year }}.
</p>
```

gemfarmer: looks like I was too late commenting. Small changes at any rate :)

coreycaitlin:
Good point @gemfarmer — working it into the next round!

PR #1412, opened 2016-05-14
Modularize state pages

shawnbot:
[🔌  Preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/modular-state-pages/states/UT/)

I'm working on breaking up the state pages into more manageable pieces, in the form of includes. More on this front next week, but the basic gist is that most of the includes should be usable on pages for either offshore regions or the entire US if need be, in addition to states. Some of these look up data on their own, which I think is not the right way to do it; they should accept data directly rather than having to get it from `site.data`. ~~More on this next week!~~

I think this is good to go. It's not 100% "there" yet, for some of the reasons listed above; but it's good enough for now. I'm going to file another issue for the tidying phase of this work.


gemfarmer:
@shawnbot what is the status here? It says WIP, but if we want to move data alongside maps, I want to make sure I'm working from the latest

gemfarmer:
@shawnbot: I think I’m going to head out for the day. It was good to familiarize myself with the state pages again. If I do do some sort of data-beside-map view, I think I will want to break the state-page.html datasets into another set of includes, like this PR. Let me know what you find and what you think of that idea and I can take a stab at it in the morning.

gemfarmer:
@shawnbot I think location is fine for now. I guess what I meant in the above comment was to break the state page into a set of includes that would include copy and the `display-{{type}}` includes. I don't think it will take that long. I'll put together a PR and you can let me know what you think. A

As for this PR, it looks good!

Issue #1403, opened 2016-05-12
Cross-browser test SVG work


meiqimichelle:
Test SVG stuffs in all the browsers:

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] IE 11

shawnbot:
Once #1424 is merged, IE 11 will be good to go! 🎉

shawnbot:
💥

PR #1424, opened 2016-05-18
IE SVG map fixes

shawnbot:
[😅   Preview on Federalist](https://federalist.18f.gov/preview/18F/doi-extractives-data/ie-svg-fix/maps/test/)

This wraps up #1403 with some IE11-specific SVG fixes, specifically:

1. We've included a [fork](https://github.com/18F/svg4everybody) of [svg4everybody](https://jonathantneal.github.io/svg4everybody/), which shims support for SVG "external resources" (referenced by the `<use>` element) in older/non-compliant browsers. Caveats:

  * For now we're referencing our forked branch in `package.json`, but we've opened an upstream [pull request](https://github.com/jonathantneal/svg4everybody/pull/117) with our first round of improvements, and will be filing another one if that one is merged.

1. We've implemented the [padding hack](http://tympanus.net/codrops/2014/08/19/making-svgs-responsive-with-css/) with some base CSS and a `svg_viewbox_padding` template filter, which you can see in action in [state](https://github.com/18F/doi-extractives-data/blob/ie-svg-fix/_includes/state-map.html#L1-L2), [county](https://github.com/18F/doi-extractives-data/blob/ie-svg-fix/_includes/county-map.html#L1-L4), and [land ownership](https://github.com/18F/doi-extractives-data/blob/ie-svg-fix/_includes/location/section-ownership.html#L8-L11) maps. Caveats:

  * This particular method is by no means a "final" solution, because currently it requires us to mirror the non-100 width percentages in both the CSS and the templates, which is 💩 . More research is needed to find a more sustainable way of handling this, or if we can get away with simply setting the percentage widths on an outer element.
  * For now I've done away with the Chrome-specific [technique](http://codepen.io/shawnbot/pen/KzGdzE) of using CSS padding to add a "buffer" around the bounds of our zoomed in maps, because it doesn't work in any of the other browsers I've tested. I'm looking for another way to do this, but for now we'll either have to 1) be okay with the bounds of states running right up to the edge of state maps, or 2) go back to buffering the viewBox values in our SVG generation scripts.

1. I also fixed a naming collision with an IE-specific table element property, `cells`, in the bar chart component. These now work in IE11, too!


gemfarmer:
@shawnbot I'm not too worried about the whitespace in those plugins. Glad you found the bar issue too!

Merging!

Issue #1399, opened 2016-05-11
Write tests for data


gemfarmer:
This is a bit of an epic and ongoing process. To start, we need to write tests to: 
* confirm the data in its **raw form** is the same as it is **in the db** as it is in **the yml** that we are using for templating
* make sure that all our $ figures are being parsed properly, etc.
* unit test utility functions

#1440 started adding tests for revenue, but we still have more to do:
---
- [x] Reinstate Circle CI
- [ ] Add tests for other data types
  - [x] Disbursements #1455
  - [x] Revenues #1440
  - [x] Production #1500 
  - [ ] Jobs #1501
  - [ ] GDP #1502 
  - [ ] Exports #1503

meiqimichelle:
@gemfarmer to reach out to @shawnbot to schedule pairing on this.

gemfarmer:
@shawnbot could we set aside some time later in the week to discuss this? I think it would be good to get this started sooner than later.

gemfarmer:
@shawnbot just sent you an invite for Thursday afternoon. Let me know if that works for you

gemfarmer:
We need to break this up in to sub issues! Doing this now!

meiqimichelle:
Thanks for making those issues! I'll make an epic label and assign all those to it, and then close this issue.

meiqimichelle:
These are all labeled now, so closing this larger issue in favor of the smaller issues.

PR #1440, opened 2016-05-26
Add state revenue data tests

shawnbot:
This should start us off on #1399 by introducing a patterns for testing the federal revenue data in the `_data` directory. Currently it only tests the latest onshore (state) data from Nathan, for which @mentastc has very helpfully provided us with a pivot table that corresponds to our rollups by revenue type. Here are the gotchas that @gemfarmer and I ran into:

1. The pivot table that Chris generated had a couple of issues with non-unique commodity values that included stray spaces. We collapsed these two instances by hand, but to save us time and headaches in the future we should request that ONRR ensure that all of the fields are trimmed before passing then off to us.
1. There are some rounding errors, but we've worked around them by asserting that our values fall within acceptable ranges. Specifically:
  * The revenue type aggregates pass if they're within $1 of the values in the pivot table.
  * The "all commodity" totals in the our commodity-grouped state revenue data pass if their values are within $_x_, where _x_ is the number of commodities in a grouping. In other words, the rounding error should only ever be as high as the number of dollar amounts we're summing.

The data tests are currently organized under `test/data`, and I've updated `package.json` to include them when you run `npm test`. We should hook this repo back up to Circle CI so that we can start running these tests whenever changes are pushed.


gemfarmer:
Looks good! ‼️

Issue #1439, opened 2016-05-26
Bug: Beckley search doesn't work from the input field


gemfarmer:
The Beckley search is up and running, but only works properly from the input in the /search-results page, not he input in the header :(

Issue #1438, opened 2016-05-26
Don't hard code beckley API key


cmc333333:
https://github.com/18F/doi-extractives-data/blob/dev/js/components/search.js#L26 hard-codes the API key into the search function. While they key isn't sensitive (it's provided to the browser), it should be configurable

PR #1442, opened 2016-05-27
Make Beckley search more accessible, searchable from header, configured in yml

gemfarmer:
[PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/beckley-federalist/search-results/?q=oil)

For #1439 and #1438 

This PR makes it so that the header is searchable from the header input, moves the api_key for beckley into the `_config.yml`, and removes several of the repeating ids that were causing the `seach-results` page to be less accessible and fail the HTML_CodeSniffer.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,6 +11,8 @@ url: https://useiti.doi.gov
 # app version number
 version: v1.3.2
 
+beckley_api_key: LXJh2PKSC6zxY0YNuBRYgIj2JxSPcDwSPCZuHBG1
+
 sass:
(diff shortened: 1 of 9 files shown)


meiqimichelle:
Looks like Beckley might be down again. We're going to get that running before merge here so we can make sure to test this branch properly.

gemfarmer:
@shawnbot I addressed the require issue you mentioned [above](https://github.com/18F/doi-extractives-data/pull/1442#issuecomment-222806848)

Issue #1434, opened 2016-05-25
Add sparklines or barchart component


gemfarmer:
### Refs
Design: #1348 
Code: #1431

Like this!

![screen shot 2016-05-25 at 4 44 14 pm](https://cloud.githubusercontent.com/assets/4803473/15557308/f43caed8-2297-11e6-824e-ed6788ff6a11.png)

shawnbot:
I'd love @ericronne's input on #1448, then we can close this out when it merges.

meiqimichelle:
@ericronne @shawnbot Did y'all chat on this? Can we merge? Thx!

shawnbot:
Should we turn this into the bar chart design issue?

meiqimichelle:
Sure! To be clear, I think you mean: turning the line charts into the latest design on the bar charts, right @shawnbot ?

shawnbot:
Yep, @meiqimichelle!

ericronne:
@shawnbot 
Assuming not all bar charts will cover a tidy 10-year stretch, we need to set the design pattern for when we have various numbers of years. Here's what a seven-column chart would look like if we merely stretched it ...

![image](https://cloud.githubusercontent.com/assets/9259185/16463160/1eedc3b0-3dfa-11e6-879a-a8f8c02e885c.png)

This may not be bad (at least for now), especially if these charts with various numbers of columns don't commingle.

Also this is one way we could treat gaps in data ...
![image](https://cloud.githubusercontent.com/assets/9259185/16463200/606591b0-3dfa-11e6-9d43-e4083e2d2797.png)

shawnbot:
Yep, good call @ericronne. The outlier here might be whichever of the all lands production commodities has data back to the 1970s, but in that case we can either constrain the year range the data side (to fit the narrowest common window), or explicitly limit all of the related bar charts to the same year range so it doesn't look weird (and so that it's less confusing). cc @meiqimichelle

usEITI:
👍

On Wed, Jun 29, 2016 at 2:07 PM, Eric Ronne <notifications@github.com>
wrote:

> @shawnbot <https://github.com/shawnbot>
> Assuming not all bar charts will cover a tidy 10-year stretch, we need to
> set the design pattern for when we have various numbers of years. Here's
> what a seven-column chart would look like if we merely stretched it ...
>
> [image: image]
[...]

Issue #1348, opened 2016-03-28
Visual design: State profile pages


meiqimichelle:
This will come from #1347 and can hopefully begin a week from today.

ericronne:
@gemfarmer here are some primordial oozings for the state template.
Note that — as an experiment — i've narrowed the overall grid from 1024 to 960. Feel free to use the current grid for now, and we can explore the idea of narrowing in a separate issue.

[Sketch file](https://drive.google.com/open?id=0BwWYNZcEDfwabmc4N0VoYnJxelE)

I've left the process section in, but as we discussed, maybe we should fold that into a tab on the fees-by-commodity chart.

Reference image...
![states-part01](https://cloud.githubusercontent.com/assets/9259185/15514925/1a2ce3f2-21b1-11e6-80a0-2ee59e00598e.png)

ericronne:
@meiqimichelle @gemfarmer here's an alt label for the commodity revenues tab...
![image](https://cloud.githubusercontent.com/assets/9259185/15520397/c3d2cbfa-21cb-11e6-80c7-19b415624942.png)

ericronne:
@meiqimichelle should that ^ table have a year label?

meiqimichelle:
@ericronne Yes! It will always show the most recent year (this is our idea anyways). Then, the breakdown tables/sparks will show the trends and various years.

ericronne:
@meiqimichelle @gemfarmer @shawnbot 

Let's call this "v1.0" (tho i haven't yet designed the disbursements chart, and i'm sure there are fine pts in here left to iron out).

Checking out of EITI for a week or so, but please reach out with any questions &c!! I'm only a DM or appear.in meetup away...

**[Sketch file on gdrive](https://drive.google.com/open?id=0BwWYNZcEDfwaa1pFSDEzWGtIVnc)**

### Reference
![state page - utah](https://cloud.githubusercontent.com/assets/9259185/15556353/1da9c6de-2293-11e6-884b-7c71d15082e6.png)

ericronne:
@gemfarmer @shawnbot @meiqimichelle What do you think of this design for the expanded commodity revenues tables?

![component - expanded commodity table](https://cloud.githubusercontent.com/assets/9259185/15910869/928c49b8-2d92-11e6-826b-5d20295be8a2.png)

[ :gem: Sketch file on gdrive](https://drive.google.com/open?id=0BwWYNZcEDfwadFhYaXhGN1pudFE)

[...]

PR #1448, opened 2016-06-01
State profile charts

### LINKED ISSUE #1477: Charts should not show zero for years that are out of range
## Expected Behavior
Line charts should not dip to zero when there is no data, e.g. when the provided years are out of range. I think the fix for this is just to add a call to [line.defined](https://github.com/d3/d3/wiki/SVG-Shapes#line_defined) so that the lines aren't drawn if the the y-value at any point is not a number.

## Actual Behavior
The line charts currently plot any missing value as zero, which is misleading for many of the charts:

image image

## Steps to reproduce the behavior
Visit any [state page](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-charts/states/CA/#employment) and look at the employment or production sections. Both of these data sets have a different year range than the revenue data (which is the one we're using for the entire page), so they all tail off to zero.

shawnbot:
Fixes issue #1434

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-charts/states/UT/#revenue)

- [x] Adds revenue $ line charts
- [x] Adds production volume line charts
- [x] Adds charts for GDP
- [x] Adds charts for exports
- [x] Adds charts for jobs
- [x] Cleans up relative scaling hack
- [x] Matches comps in #1348

I also fixed #1477 in the process, by making all of the charts show only their own relevant year ranges.

/cc @ericronne


_sass/_variables.scss
```diff
@@ -61,6 +61,7 @@ $light-black: #303030;
 
 // State pages
 $green-land: #d5e5cb;
+$chart-green: #a5d78a;
```
gemfarmer: consider consolidating these two variable colors to stay consistent. If we set `$green-land: #a5d78a;` the land ownership map looks like this: 
screen shot 2016-06-07 at 5 04 53 pm

gemfarmer: Actually, nvm. It looks fine for Utah, maybe even more vibrant, but on other states where there is less federal land ownership, it starts looking overwhelmingly bright:

screen shot 2016-06-07 at 5 12 28 pm

_sass/blocks/_chart-list.scss
```diff
@@ -0,0 +1,43 @@
+.chart-list {
+  @include font-size(0.8);
+  @include outer-container();
+
+  .chart-title {
+    border-bottom: 1px solid #ddd;
```
gemfarmer: maybe `$mid-gray: #ebebeb;`?

gemfarmer:
This looks good to me. Definitely in line with the existing sass organization.

_sass/blocks/_chart-list.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  figure {
+    margin-top: 0;
+  }
+
+  figcaption {
+    font-weight: $weight-bold;
+    display: flex;
+    padding: 5px; // FIXME this is an approximation
```
gemfarmer: will `$base-padding-lite` work here? It generally ends up computing to around 5px

gemfarmer:
@shawnbot I removed the conflicts on this branch here: #1480

gemfarmer:
@shawnbot I think it makes more sense to simply rebase on this branch

gemfarmer:
reviewing now, @shawnbot !

gemfarmer:
Looks awesome! A few things.

- I made an issue [here](#1490) for the spark line "Details" toggle.
- Is the GDP chart not getting a line chart because of the %/# split?
- a few layout issues, but that is outside the scope of this PR

I vote merge

PR #1455, opened 2016-06-03
Adds disbursements tests

gemfarmer:
This PR adds a new set of tests for the disbursement data @ `_data/state_disbursements.yml`

The tests put in place caught differences in the data! It turns out that we were generating our output data using both the state disbursement and historic preservation disbursement data. This PR is working on the assumption that the historic preservation data is a subset of the state disbursement data, but we should verify that this is the case!

Also adds an additional pivot table to the [gDrive](https://drive.google.com/drive/u/0/folders/0B4b900J0no9nSHFCY0VoWkNfYTQ)
/cc @shawnbot @mentastc


gemfarmer:
srsly. Do you think it is ready to merge, @shawnbot ?

PR #1479, opened 2016-06-08
Regional collections

### LINKED ISSUE #1476: Convert states and offshore areas to collections
Currently, the [states], [offshore regions], and [offshore areas] are listed only in CSV files. Soon we're going to need to accommodate state-specific content (and probably data), e.g. for opt-in states, which is not easy to incorporate into the CSV files. We should ditch those altogether and turn the lists of different region types into collections, which also allows us to ditch the kinda janky Jekyll plugin that currently generates the state pages and _might_ be contributing to our slow build times.

[states]: https://github.com/18F/doi-extractives-data/blob/dev/_data/states.csv
[offshore regions]: https://github.com/18F/doi-extractives-data/blob/dev/_data/offshore_regions.csv
[offshore areas]: https://github.com/18F/doi-extractives-data/blob/dev/_data/offshore_areas.csv

shawnbot:
Fixes #1476 by turning states (California), offshore regions ("Gulf of Mexico"), and offshore areas ("Central Gulf of Mexico") into Jekyll collections.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/regional-collections/states/CA/)

- Instead of using [this plugin](https://github.com/avillafiorita/jekyll-datapage_gen), there is now `states` collection that is used to generate one page per state.
  * This resolves the issue of the state pages having titles like `UT` instead of `Utah` 🎉 
- There's also an `offshore_regions` collection that contains our for offshore regions: Alaska, Atlantic, Gulf of Mexico, and Pacific. These can eventually have URLs of their own, but I've [disabled](https://github.com/18F/doi-extractives-data/blob/5b826a5c30f96c7987e16af0d6f6b77e6465ac9b/_config.yml#L73) the output option for now.
- Offshore _areas_ (the next level down from regions, so the equivalent of counties) have a collection, and can also be given URLs of their own when we're ready. The `permalink` for each of these pages will put them under their offshore region path, e.g. `/offshore/gulf/CGM/` for Central Gulf of Mexico.

/cc @gemfarmer


gemfarmer:
WOO! This is awesome. `to-jekyll-collections` is also pretty slick. I could see that being useful for other projects with cumbersome collections

gemfarmer:
@shawnbot this looks great, but I'm not seeing great load times. Mine have spiked to the 20-30 second range.

gemfarmer:
This is so awesome! 🎉 🎉 MERGED!

PR #1483, opened 2016-06-08
How it works accordion bug

gemfarmer:
[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/hiw-renewables-bug/how-it-works/offshore-renewables/)

Fixes bug that was causing issues with the accordions in offshore renewables. My bad :(

/cc @meiqimichelle


meiqimichelle:
:mouse

PR #1489, opened 2016-06-09
State pages merge

gemfarmer:
### `dev` and `state-pages` had a BABY! 🍼 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages-merge/)

- brings the `state-pages` page up to date with the new changes from v1.3.3 on `dev`

/cc @meiqimichelle


meiqimichelle:
I think this is good to go. Federalist is still down so not everything is previewing correctly, but...I'm pretty sure we're good. I'll pull locally and double-check after merge, too.

Issue #1430, opened 2016-05-25
Add map legends


gemfarmer:
Design: #1348 
Code: #1431

Per @ericronne's designs, add legends that specify what the map gradations _mean_. These should ideally be a part of the map includes.

<img width="233" alt="screen shot 2016-05-25 at 3 09 52 pm" src="https://cloud.githubusercontent.com/assets/4803473/15554639/cf4c0e5a-228a-11e6-8295-e6c156c94208.png">
<img width="227" alt="screen shot 2016-05-25 at 3 09 41 pm" src="https://cloud.githubusercontent.com/assets/4803473/15554640/cf5cb868-228a-11e6-8f81-980be87dc36b.png">

shawnbot:
This might actually be better to generate in the `<data-map>` component, since that's where the color range and number of steps is defined. It doesn't need to be static, since JS is filling in the colors anyway.

Issue #1396, opened 2016-05-10
Define production units for all lands/waters products


shawnbot:
For some EIA commodities, such as renewables, the units are unknown. These are the products:

- [ ] `All Other Renewables`
- [ ] `Biomass (total)`
- [ ] `Conventional Hydroelectric`
- [ ] `Geothermal`
- [ ] `Other biomass`
- [ ] `Solar`
- [ ] `Wind`
- [ ] `Wood and wood-derived fuels`

We should probably create a mapping of these , e.g. in `_data/production_units.yml`:

```yaml
'Biomass (total)': 'tons'
'Wind': 'kilowatt hours'
```

Then, in our templates we would get the units with:

```html
{% assign units = product.units | default: site.data.production_units[product.name] %}
```

shawnbot:
I've added a task list to this issue description with the product names for which we don't have units in both the all lands and federal production data.

shawnbot:
Another thing to consider: Should we have shorthand notation for some units, like `kWh` for kilowatt-hours? If so, our units definition data might look like this instead:

```yaml
products:
  'Wind': 'kilowatt-hours'

shorthand:
  'kilowatt-hours': 'kWh'

**Comment by meiqimichelle** [comment]
@shawnbot is this something that we should resolve before your time gets split?

**Comment by shawnbot** [comment]
Nope, I think y'all can handle this if I'm not able to.

**Comment by coreycaitlin** [comment]
Some of these products would also make more sense with the word "energy" after them (i.e. "hydroelectric energy").

**Comment by shawnbot** [comment]
So I think @coreycaitlin hit upon a pretty good justification for having something like this instead in, say, `_data/production_units.yml` (or whatever):

```yaml
products:
  Hydroelectric:
    name: Hydroelectric Energy
    units: kilowatt-hours

abbreviations:
  kilowatt-hours: kWh
```

Then, in places where we use `product_name` in our Jekyll templates now, we would do something like this:

```html
{% assign product_name = product[1].name %}
{% assign units = product[1].units %}
{% assign units_abbr = units %}
{% if product_name in site.data.production_units %}
  {% assign _product = site.data.production_units[product_name] %}
  {% assign product_name = _product.name | default: product_name %}
  {% assign units = _product.units | default: units | default: 'units' %}
  {% assign units_abbr = site.data.production_units.abbreviations[units] | default: units_abbr %}
{% endif %}
```

meiqimichelle:
[...]

PR #1495, opened 2016-06-14
Map legends

gemfarmer:
WIP #Fixes issue(s) #1430 
 
[🌐 PREVIEW 🌐](https://federalist.18f.gov/preview/18F/doi-extractives-data/map-legends/states/UT/)

new TODO
- [x] address @shawnbot's proposed changes
- [x] replace code with svg-map-legend
- [x] css changes
- [x] make sure that map padding hack is still working 
- [x] reverse order of legend swatches (highest to lowest)

- [x] legends added to... 
  - [x] land owernship map
  - [x] productions maps
  - [x]  revenue maps
  - [x] employment maps
- [x] ensure scale is always accurate (rounding issues)
- [x] ~~add units ($ is not always correct) [NOTE: perhaps units should go in the map caption?] e.g. `County production for Oil (bbl, 2013)` like in the [employment section](https://federalist.18f.gov/preview/18F/doi-extractives-data/map-legends/states/UT/#employment)~~ this was removed due to not knowing where the units will go
- [x] format layout
- [x] add `data-dimensions` attr to non-production maps
- [x] grid (4-4-4 for now @ericronne )


gemfarmer:
@shawnbot, ready for your review!

gemfarmer:
@ericronne would also love to have your eyes on the design. Note that per our conversation about columns/grids I have kept the main column at a 4-4-4 grid. Shouldn't be terribly hard to change this in the future, just didn't want to do anything until I new for certain that we would be using a 3-3-2 grid instead.

gemfarmer:
@shawnbot fair enough. I will get all of the css off the IDs. 

Do you think the general approach within the `data-map` component makes sense?

gemfarmer:
@shawnbot I was able to simply remove the ID selector. It turned out to be unnecessary.

js/components/data-map.js
```diff
@@ -41,7 +48,159 @@
...
+          var swatches = d3.select(this)
+            .selectAll('dl [data-step]')
+            .datum(function() {
+              return +this.getAttribute('data-step') || 1;
+            });
+
+          swatches.style('background-color', scaleSteps);
+
+          format.start = function(value, units) {
+            if (units === '$' || units.includes('dollar')) {
```
shawnbot: `String.prototype.includes()` has pretty [weak browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility). We'll either need a polyfill here or a function that does something like `units.indexOf('dollar') > -1`.

_sass/components/_all.scss
```diff
@@ -2,6 +2,7 @@
 @import 'bar-chart-table';
 @import 'buttons';
 @import 'data-map';
+@import 'data-map-legend';
```
shawnbot: I think that unless there's actually a `<data-map-legend>` component, these styles should live in `_data-map.scss`.

_includes/county-map.html
```diff
@@ -42,4 +56,28 @@
...
+    <h4 class="details-container">
+      <button is="aria-toggle"
+        aria-controls="{{ include.toggle }}">&plus; Details</button>
+    </h4>
+  {% endif %}
+
+
 </div>
+
+<dl class="legend-container">
```
shawnbot: According to [the spec](https://www.w3.org/TR/html-markup/dl.html), `<dl>` is only allowed to contain `<dt>` and `<dd>` elements. I think that this needs to be reorganized.

_includes/location/section-ownership.html
```diff
@@ -36,9 +36,33 @@
...
         </svg>
+
+
+
       </div>
 
-      <!-- <figcaption><icon class="icon-plus-lg"></icon> Details</figcaption> -->
     </figure>
+
+    <dl class="legend-container">
```
shawnbot: Same issue here with [children](https://www.w3.org/TR/html-markup/dl.html) of `<dl>`.

js/components/data-map.js
```diff
@@ -41,7 +48,159 @@
...
             .range(colors);
 
+          var scaleSteps = d3.scale[type]()
+            .domain(stepDomain)
+            .range(colors);
+
           marks.attr('fill', scale);
+
+
+          function createSteps (steps, interval) {
```
shawnbot: Couldn't you just use [d3.range()](https://github.com/d3/d3/wiki/Arrays#d3_range) here?

js/components/data-map.js
```diff
@@ -41,7 +48,159 @@
...
+
+          format.start = function(value, units) {
+            if (units === '$' || units.includes('dollar')) {
+              return [format.dollars(value), ' –'].join(' ');
+            } else {
+              return [format.si(value), ' –'].join(' ');
+            }
+          };
+
+          format.end = function(value, units) {
```
shawnbot: This whole function makes me kind of uncomfortable. Could the suffix just come an HTML attribute? See #1395 for some context.

js/components/data-map.js
gemfarmer: ooo I think so

js/components/data-map.js
gemfarmer: it isn't a great function.

Are you suggesting that I add an attribute to `data-map` that takes the units and parses them with a jekyll plugin? I think I'm missing something here. The diff in #1395 was very large, so I'm lost there.

_includes/county-map.html
gemfarmer: hmm. Based on the formatting, I think I need nesting, so I'll try a list approach

_includes/location/section-ownership.html
gemfarmer: 👍

_sass/components/_all.scss
gemfarmer: fair point. I'll put it in the same file and separate code with comments instead.

js/components/data-map.js
gemfarmer: 👍

js/components/data-map.js
shawnbot: Sorry, I meant to reference #1396! Specifically, it looks like everything that isn't measured in dollars is just formatted as SI with a suffix. So I feel like the format could be described as either:

* `units="$"` or `units="dollars"`, which uses `eiti.format.dollars()` to format ranges; or
* anything else, which uses SI formatting a unit suffix ("kWh", etc.) at the end of a range

E.g.

```html
<data-map units="dollars">...</data-map>
...
<data-map units="kWh">...</data-map>
```

As I was writing this I started typing `<data-map units="thousands of pounds">` and realized that for overly long unit descriptions, it's probably best to have the legend's title list the units and leave that string out of the ranges. So maybe you'd mark those up as:

```html
<figure class="map county-map">
  <!-- aria-describedby points the data-map at an "external" legend? -->
  <data-map units="thousands of pounds" aria-describedby="geothermal-county-map-legend">
  </data-map>
</figure>
<figcaption class="map-legend">
  <h4>County geothermal production in 2013 (thousands of pounds)</h4>
  <ol id="geothermal-county-map-legend">
  </ol>
</figcaption>
```

This is a pretty dramatic departure from what you've got, though, so maybe simplify the logic for now based on the data-map's `units` attribute, and let's move this discussion to a new issue? 😬

js/components/data-map.js
```diff
@@ -41,7 +48,149 @@
...
+              domainEnd = [];
+            interval.forEach(function (step, i) {
+              if (i === 0) {
+                domainStart.push(domain[0]);
+              } else {
+                domainStart.push(interval[i - 1] * domain[1]);
+              }
+              domainEnd.push(step * domain[1]);
+            });
+            return domains = [domainStart, domainEnd];
```
gemfarmer: yeah yeah yeah

js/components/data-map.js
gemfarmer: I like the idea of moving the units out of the legend. After I did it, I also noticed that long units look bad, and that shorter ones look repetitive, but I thought I would wait for @ericronne's feedback/redesign. 

I'll try to pair things down where I can for now

gemfarmer: Noted in #1516

gemfarmer:
Just noticed this earlier comment 

> @gemfarmer Is having an HTML hook for each legend item a requirement of d3-svg-legend? Because I don't see why it's necessary. From a progressive enhancement standpoint, my preference would be not to have anything in the HTML that doesn't work as static content.

This PR doesn't actually use [`d3-svg-legend`](http://d3-legend.susielu.com/). When I started the branch I started off by using it, then opted to not because it seemed like what we were doing wouldn't necessarily require all the bells and whistles that it provides, but maybe that was a misstep if you think it is bloating the dom.

js/components/data-map.js
```diff
@@ -41,7 +48,153 @@
...
+          }
+
+
+          function uniq(value, index, self) {
+            return self.indexOf(value) === index;
+          }
+
+          String.prototype.has = function (units) {
+            return this.indexOf(units) > 1;
+          }
```
shawnbot: Also, I would definitely not suggest adding proprietary `String` prototype methods. I think it's better to either do this:

```js
function includes(str, substr) {
  return str.indexOf(substr) > -1;
}
```

or to polyfill `String.prototype.includes`:

```js
if (!String.prototype.includes) {
  String.prototype.includes = function(substr) {
    return this.indexOf(substr) > -1;
  };
}
```

_includes/county-map.html
gemfarmer: this was replaced by `svg-map-legend`

js/components/data-map.js
```diff
@@ -12,6 +17,10 @@
           var type = this.getAttribute('scale-type') || 'quantize';
           var scheme = this.getAttribute('color-scheme') || 'Blues';
           var steps = this.getAttribute('steps') || 5;
+          var units = this.getAttribute('units') || '';
```
gemfarmer: It is highly likely that units will be used in the data map, so I'm thinking I'll keep the units attribute for now. Fine by you @shawnbot ?

gemfarmer:
@shawnbot ready for review!

Issue #1465, opened 2016-06-07
State pages: Edit key facts intro


coreycaitlin:
The "key facts" section should be shorter! and clearer! And probably have some more lead-in content to get users oriented to what they're looking at.

coreycaitlin:
Remaining blockers:

1. Need help fixing the stray sentence about production. Setting up time with Brian!
2. Need another set of eyes to tell me if this is enough context for the page.

meiqimichelle:
This only thing left here is #1 in Corey's list above. @shawnbot , reassigning to you.

shawnbot:
Done!

PR #1506, opened 2016-06-18
Intro revision and other content refinements

coreycaitlin:
Progress on issue #1465 🎉 

- Brings back "key facts", but in simplified sentence form. Hopefully this will help users see quickly what's interesting about extractives in each state, or at least get a sense of the overall scale.
- Fixes a stray "Utah" so that the appropriate state displays instead.
- Change "GDP" in the nav to "economic impact", in hopes of reducing acronyms
- Tidies up a few product names to remove duplication of units
- Downgrades the header level of Land Ownership to avoid confusion (this may need further adjustment, but it's a start)

@shawnbot — tagging you so you can remove the key facts intro sentence for states without products in the #1 spot. (See West Virginia for an example.)

/cc @meiqimichelle: thoughts?! Does this make sense? Do we need more intro/boilerplate here, or do you think it's working in this super-concise version?

### DIFF
--- a/_includes/location/key-all-production.html
+++ b/_includes/location/key-all-production.html
@@ -7,7 +7,7 @@
     <p>{{ include.location_name }} leads the nation in production of these resources:</p>
       <ul>
       {% for product in top_all_products %}
-        {% if product.rank < 3 %}
+        {% if product.rank < 2 %}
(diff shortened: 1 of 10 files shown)


coreycaitlin:
Minor change to only show production commodities (and percentages) for commodities the state is actually a leader in.

Issue #1376, opened 2016-04-21
Decide how to use our exports data in the new state profile pages


meiqimichelle:
The exports dataset is not very straightforward to understand, and doesn't add value to most states (ie, there is no data for many states because if an extractive resource wasn't in the top 25 exports, we get nothing). We should consider re-emphasizing it in the next site iteration.

coreycaitlin:
Given the oddness of this dataset, I would vote to include exports facts — for those states for which extractives are in the top 25 — in the summary section, but not have a deeper section below. The limited nature of the dataset makes it...not that useful year-over-year, if I understand correctly?

shawnbot:
Yep, exactly. +:100:

coreycaitlin:
I'm going to

coreycaitlin:
Update: based on the structure we came up with for #1472, it's looking like we kind of do need to include this on state pages, which brings us back to the fact that we need to write content for two situations:

1. There is exports data for a state because _one or more_ extractive industry products is in the top 25 exports for that state. BUT there may be other products not included in the data because they're, say, 26th in the list of exports. And the year-over-year data may be misleading, particularly if a product has hopped in and out of that top-25 group. Finally, the percentages are largely meaningless, because the totals we have are totals of either all top-25 exports or all extractive industry exports in the top 25, not ALL exports.

2. There is no exports data, because a state's extractive industry products are not in the top 25 exports for that state. This _doesn't_ mean there are no extractive industry exports — or even that they aren't significant! — and may say more about a state's other exports than about extractive industries. In this instance, we don't want to be misleading. It's temping to simply not show this section, but that might be strange.

I'm going to draft content that attempts to solve for both of these situations.

shawnbot:
I'm _pretty_ sure that we have the total exports (for all products, not extractives) in the original data. If the totals that we ended up with are wrong, then that's a bug that #1503 should reveal.

coreycaitlin:
@shawnbot thanks — that would certainly help make this more interesting. I'll write it as if the percentage refers to ALL exports (not just top 25), and let's make sure to check those numbers.

coreycaitlin:
Remaining issues with this section, which I can't figure out how to solve:

* In addition to charts for each product, there are charts for both "Total" and "All." I don't think we need either.
* I'd like the intro sentence to be more specific either _if there is only data for one product_ or _if a product is the top export in the state_. For example:

[...]

PR #1510, opened 2016-06-20
removed exports charts for All and Total

coreycaitlin:
Solves one of the issues in #1376.

Removes charts for "All" and "Total", because they didn't make sense or add information.

/cc @gemfarmer


gemfarmer:
@coreycaitlin looks good! merging

Issue #1432, opened 2016-05-25
Sidenav with submenu


gemfarmer:
Like this! 
Design: #1348 
Code: #1431 

![screen shot 2016-05-25 at 4 36 54 pm](https://cloud.githubusercontent.com/assets/4803473/15557127/ee13704c-2296-11e6-8013-4d321d6c3953.png)

gemfarmer:
@ericronne are we still planning on using this approach?

ericronne:
We may want to test it first, but i still think it's a useful solution. We just need to watch the length.

gemfarmer:
A few to-dos here:
- [ ] add sub menu to jekyll templates (including changes to yaml frontmatter)
- [ ] style submenu
- [ ] make it so that all IDs that match `nav_items` in the front matter become navigable links

PR #1511, opened 2016-06-21
Sidenav with submenu

gemfarmer:
Fixes issue(s) #1432 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/sticky-nav-submenu/)

TODO:
- [x] highlight parent menu item when any child is active
- [x] add all submenu items
- [x] ensure that `select` style navs still work properly

- adds the ability to add subnav items on the sticky menu
- refines the function that determines the "active" nav item
- highlights parent menu item when any child is active
- adds submenu items from mock
- simplifies the syntax and applies markdown where possible. @coreycaitlin, I made changes so that any header can become a nav item if its title is referenced in snake case in the `nav_items` front matter.

So  `## Example header` can be referenced like so in the yaml:

```
nav_items:
  - name: example-header
    title: Example header
```

Longer, more unwieldy titles should probably still be referenced using an id. So the above front matter would need a real html header:

```
<h2 id="example-header>
  Example header that is super long and annoying to look at
</h2>
```

/cc relevant people

### DIFF
--- a/_case-studies/boone-logan-and-mingo.md
+++ b/_case-studies/boone-logan-and-mingo.md
@@ -7,7 +7,7 @@ resource: coal
 nav_items:
   - name: intro
     title: Top
-  - name: geology
+  - name: geology-and-history
     title: Geology and history
   - name: production
     title: Production
@@ -17,7 +17,7 @@ nav_items:
     title: Revenue
   - name: costs
     title: Costs
-  - name: data
+  - name: data-availability
     title: Data
 selector: list
 ---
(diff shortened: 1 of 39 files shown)


meiqimichelle:
This is looking really good! I know you've still got stuff to do, but....yays.

gemfarmer:
@meiqimichelle it is looking done now! Plz review :)

meiqimichelle:
🔥 🔥 🔥

Issue #1513, opened 2016-06-21
State pages: make special cases available in template


gemfarmer:
## Expected Behavior
We should be able to reference whether or not a state is an opt-in state, has case-studies, etc. like so:
```
{% if state_is_opt_in %}
```
or 
```
{% if state_has_case_studies %}
```
## Actual Behavior
This doesn't exist yet!
## Steps to reproduce the behavior
- add statuses to the front matter in a page's markdown file.

@coreycaitlin I can start this and hand off to you!

gemfarmer:
@coreycaitlin, which would you prefer:  
- to have one file that has all the special cases information? like so:

```
AK:
  - case_study
    opt_in
    priority
AZ:
  - case_study
CA:
  - case_study
    opt_in
    priority
LA:
  - case_study
MI:
  - case_study
MN:
  - case_study
NV:
  - case_study
TX:
  - case_study
WV:
  - case_study
WY:
  - case_study

```

- or to edit each state's file? Either works, ready to do whatever is more straightforward on your end.

coreycaitlin:
Hmm, my hunch is that one file per state will be more intuitive to maintain in the long run.

gemfarmer:
Are you thinking it would be more intuitive because when you want to know something about that state you go to that file and you can see the relevant statuses for that state?

Because it does seem to be nice having all of the "special" info in one place.

That being said, if you think that one file per state will be easier to maintain, we already have one file per state, so it really isn't too difficult to just tack on a few extra data for that state.

coreycaitlin:
🎉  All set!

PR #1525, opened 2016-06-23
Special cases

gemfarmer:
Fixes issue(s) #1513

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/special_cases/)

- So far I have added two alternative ways to handle conditional logic for states that are "special", where __special__ means that state has case studies, has opted in, or is a priority state
  - 1: create one data file to manage all state statuses
  - 2: Edit a particular states' statuses in its own file.  

/cc @coreycaitlin


gemfarmer:
wrong branch

Issue #1393, opened 2016-05-09
Glossary accordion bug


gemfarmer:
This bug showed up on @mentastc's PC in Chrome.

- accordion chevron icons were misaligned
- filter functionality didn't work properly when the glossary search input was cleared.

ericronne:
![image](https://cloud.githubusercontent.com/assets/9259185/16203406/b6d33948-36df-11e6-88d0-2b6c492aa811.png)

![image](https://cloud.githubusercontent.com/assets/9259185/16203413/c1e8125e-36df-11e6-9966-d8997cd1c76e.png)

coreycaitlin:
Seeing wackiness in both chrome and safari on my machine:

![screen shot 2016-06-20 at 10 08 54 am](https://cloud.githubusercontent.com/assets/1450354/16203419/0ef2b164-36cf-11e6-8a31-3312a3887a1f.png)

<img width="669" alt="screen shot 2016-06-20 at 10 07 31 am" src="https://cloud.githubusercontent.com/assets/1450354/16203424/13551e5e-36cf-11e6-8021-f6eba414da50.png">

gemfarmer:
I'm going to schedule some time to pair on this with one of you @meiqimichelle @coreycaitlin @ericronne 

'cuz my browser is working 👌

gemfarmer:
Scheduled for this friday. Make it if you can! I just need one of you :)

gemfarmer:
Closes with #1526

PR #1526, opened 2016-06-23
Special cases conditional logic

gemfarmer:
Fixes issue(s) #1513

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/special_cases/)

- So far I have added two alternative ways to handle conditional logic for states that are "special", where _special_ means that state has case studies, has opted in, or is a priority state
  - 1: create one data file to manage all state statuses
  - 2: Edit a particular states' statuses in its own file.  
- added the state regulations, available via an include `location/state-regulations.html`. @coreycaitlin I tried to make this a series of state files, but it was proving glitchy. I think it will make a good enhancement.
- caught glossary bug #1393 
In both instances, I have added a few conditionals that will be available on every state page:

- `is_special_state`: evaluates to true if the state has any _special_ cases
- `is_case_study`:    evaluates to **true** if the state has case studies
- `is_opt_in`:            evaluates to **true** if the state has is an opt in state
- `is_priority state`:  evaluates to **true** if the state is a priority

```
{% if state_is_special %}<h1>You are special</h1>{% endif %}
```
/cc @coreycaitlin


gemfarmer:
@coreycaitlin this is ready for review! You might want to move regulations into another section and remove the nav item. Up to you! I put it there for visibility.

Issue #1491, opened 2016-06-10
State pages: details view for line charts


meiqimichelle:
[placeholder issue for @ericronne to add images for feedback]

ericronne:
@meiqimichelle @gemfarmer @shawnbot 
Revised the design to align with the changes i made to the big revenue table...
![image](https://cloud.githubusercontent.com/assets/9259185/15971182/905e3e6e-2efd-11e6-92d6-df564f36557d.png)

coreycaitlin:
@ericronne For this chart, I'm finding myself a little tripped up on trying to understand all the visual scales both vertically and horizontally — am I missing something, or is the sparkline conveying the same thing as the bars below?

ericronne:
Yeah it's the kitchen-sink view of the same info. Btw the line chart is designed to be interactive, so you could hover over any of the years (or corresponding points along the line) and see the figure for that year.

But your point makes me wonder...
Devil's advocate here (and maybe we should continue with testing before playing with this any more right now), but one alt approach would be to forego bars on the details view, and simply show the figures. We can also consider going back to the original bar-chart design, rather than lines. That many bars on a page gets pretty heavy, though.

ericronne:
@meiqimichelle @gemfarmer @coreycaitlin @shawnbot 

Here are three options for the chart design, each showing both the default and expanded views. Each of these would be interactive in both view states. When collapsed, the number at the bottom changes to match the year (or chart area) which the mouse rests. The dot would also move, or the bar would turn green.

In the expanded state, both the charts and the number table would be "scrubbable" (except for expanded chart C).

@shawnbot You've raised an issue before regarding hover states which have no click functionality. Does this bug you, then? Should we toggle between default and expanded when the user clicks anywhere on the chart or table?

## Defaults
![artboard 21 copy 2](https://cloud.githubusercontent.com/assets/9259185/16055397/b580d45e-3236-11e6-80b7-258d839ebfa1.png)

![artboard 21](https://cloud.githubusercontent.com/assets/9259185/16055408/c11456e2-3236-11e6-96ac-c53fd8787ee7.png)

![artboard 21 copy](https://cloud.githubusercontent.com/assets/9259185/16055402/bc19a516-3236-11e6-992d-02b6d10b1ddd.png)

## Hovers
![artboard 21 copy 3](https://cloud.githubusercontent.com/assets/9259185/16056178/f4c2110c-3239-11e6-806e-55fc150f2034.png)

[...]

PR #1555, opened 2016-07-01
State pages layout cleanup

meiqimichelle:
This PR starts to get the state profile pages ever-closer to the true mock.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/megachart-tidy-eh/states/CA)

In particular:
- The megachart is much closer to the mock and has @coreycaitlin 's latest content. Still missing: solar and wind (we haven't implemented this data-wise yet so there wasn't a place to put it)
- Adds the summary sentences back to all (I believe) places they should live, and removes the detail chart view from places it shouldn't be, as per #1491.
- Starts to adjust spacing and typography, but didn't get toooo too far, ran out of time

This does *not* include the latest county view layout. I hope to get to that tomorrow.

/cc @ericronne


ericronne:
Bitchin!

PR #1565, opened 2016-07-04
Updates open/close behaviors for glossary terms

### LINKED ISSUE #1563: Design: Glossary - clicking on terms in list does not open/close definition
The glossary is an incredibly cool element on the website, but I noticed that in order to open or close a term definition in the list of terms I have to click on the chevron _next to each term_ instead of being able to click on the term itself or being able to click anywhere on the term row.

The chevron makes for a small click target and I think that the glossary would be much more useful if the same click handler were present on the row as is on the chevron.

Tested in Safari, Firefox, and Chrome on Mac.

no-onclick-on-glossary-terms

dmsnell:
Closes #1563

Previously, when opening the gallery and reading through the list of
terms, one had to click on the chevron next to a term itself in order
to open or close the definition.

This patch allows for performing the same action but also through
clicking on the term itself, expanding the click area and making for
what I feel is a more expected behavior.

**New Behavior**
_See #1563 for video of old/existing behavior_
glossary-does-toggle-on-click

**Caveats**
 - I'm not familiar with the codebase, so while this seems to _work_, I
   don't know if it's _right_ for the project.

 - I wasn't sure if I needed to submit the changes to the minified code 
   in this PR. There are some big changes that I didn't introduce into the
   minified version that nonetheless came over in rebuilding it. Maybe
   these are related to node versions? Maybe a previous commit didn't
   update the minified copies?

 - The actual work seems like it's happening inside of the `List()`
   which made it more difficult to figure out. This is also why I'm not
   convinced that I made the right changes in the right places.

 - There are lingering issues with the cursor: should it be a pointer
   when hovering over the glossary terms?

Thanks!

### DIFF
--- a/_sass/components/_glossary.scss
+++ b/_sass/components/_glossary.scss
@@ -33,6 +33,7 @@ $term-width-max: 220px;
 
 .glossary-term {
   color: $white;
+  cursor: pointer;
   display: inline-block;
   margin: 0;
   max-width: $term-width-max;
--- a/js/components/glossary.js
+++ b/js/components/glossary.js
@@ -60,6 +60,7 @@
     // Bind listeners
     self.$toggle.on('click', this.toggle.bind(this));
     self.$body.on('click', '.toggle', this.toggle.bind(this));
+    self.$body.on('click', '.glossary-term', this.toggleTermFromClick.bind(this) );
     self.$search.on('input', this.handleInput.bind(this));
 
     $(document.body).on('keyup', this.handleKeyup.bind(this));
@@ -141,6 +142,17 @@
     toggle: function() {
       var method = this.isOpen ? this.hide : this.show;
       method.apply(this);
+    },
+
+    toggleTermFromClick: function(event){
+      var $target = $(event.target);
+      var doExpand = $target.parent().attr( 'aria-expanded') === 'true' ? 'false' : 'true';
+
+      $target
+        .siblings( 'p' )
+        .attr('aria-hidden', doExpand )
+        .parent()
+        .attr('aria-expanded', doExpand );
     },
 
     show: function() {
--- a/js/lib/main.min.js
+++ b/js/lib/main.min.js
@@ -100,7 +100,7 @@
 
 	var __WEBPACK_AMD_DEFINE_FACTORY__, __WEBPACK_AMD_DEFINE_RESULT__;!function() {
 	  var d3 = {
-	    version: "3.5.17"
+	    version: "3.5.16"
 	  };
 	  var d3_arraySlice = [].slice, d3_array = function(list) {
 	    return d3_arraySlice.call(list);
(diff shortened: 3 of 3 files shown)


dmsnell:
Also, I would like to believe that I am not the reason this is failing, but if I am, I apologize for not knowing how to fix it.

_From CircleCI_
```
Gem::InstallError: jekyll requires Ruby version >= 2.0.0.
Installing nokogiri 1.6.7.1
An error occurred while installing jekyll (3.0.0.pre.beta9), and Bundler cannot
continue.
Make sure that `gem install jekyll -v '3.0.0.pre.beta9'` succeeds before
bundling.
```

gemfarmer:
Hello again @dmsnell! Great work. I think that being able to toggle glossary items via the text (or at least the title text) is a great enhancement!

The Circle CI error is not your fault. We having been working no a larger addition on a separate branch and the Circle CI tests are for the js in that branch, which builds the site with a later jekyll version. 

I just made an [issue to fix this problem](https://github.com/18F/doi-extractives-data/issues/1566), so that contributors like yourself aren't scolded for their harmless code additions.

A few comments on your "caveats"

- The minified JS should be included in the commit, so you did well here. I can take a look to make sure it is building properly and not including or excluding unwanted code.
- I'll take a deeper look in a second, but the cleanest alterations might be in `js/components/accordion.js`, not `List()` or `glossary.js`
- If it is clickable, I think it should have a cursor. Do you think _all_ the text should be clickable? As in term title text and the term definition text? Or do you think that making just the header (title text and chevron) would be a better interaction? Thoughts @ericronne?

ericronne:
Agreed, pointer cursor on hover would be great, @gemfarmer. I wouldn't make the definition text clickable. FEC is a good ref imo

gemfarmer:
Ok. Sounds good @ericronne.
For your reference, @dmsnell: [FEC site with similar glossary](https://beta.fec.gov/)

gemfarmer:
@dmsnell Ok, I took a deeper look at your changes and it still looks great. Ultimately, these changes _should_ be made in `components/accordion.js`, but for now, this is a good enhancement. I have made an issue [here]() to handle that progressive enhancement task.

A few changes that would be great to make before merging:

1. Pull in the latest changes from our `dev` branch, or add a [`circle.yml` file with the following contents](https://github.com/18F/doi-extractives-data/blob/dev/circle.yml). This should get rid of the Circle CI errors
2. Add `cursor: pointer` to the `.glossary-term` class [here](https://github.com/dmsnell/doi-extractives-data/blob/503c4c4e2387c0e8b12b68d8b6bb2f32f9802268/_sass/components/_glossary.scss#L34)
3. Update the minified files. I pulled your fork changes and built them and saw diffs, so my best guess is that you need to run an `npm install` after pulling in the latest updates to our dev `package.json`. It looks like some of the differences were updates to the `queue` and `d3` packages.

dmsnell:
@gemfarmer thanks for the feedback. I have updated this PR:

 - Rebased against current `dev` tree
 - Cursor pointer added in c74c84d
 - Minified build cleaned up in 5b300a5

It's probably worth pointing out that the issues in the minified build came from npm package differences which I was only really able to figure out by looking at the diff and trying to match versions. All of the projects I work have all basically ruled out using the npm compatibility options in favor of explicit version numbers because of this: implicit state that allows developers to have different environments without realizing it. Of course, we don't commit the minified builds into `git` either.

gemfarmer:
Ok @dmsnell. This looks great. Thanks for doing the heavy lifting and figuring out the npm versioning!

Merging 💥 🎉 !

PR #1580, opened 2016-07-07
National page

gemfarmer:
Fixes issue(s) #1521 

[:sunglasses: P R E V I E W](https://federalist.18f.gov/preview/18F/doi-extractives-data/national-page/states/)

- I didn't pay a ton of attention to formatting or getting maps on the page. This is more/less a data and basic html structure dump to get the page jump-started!
- Get content and data on to the national states page
- [x] all production
- [x] federal production
- [x] revenue
- [x] gdp
- [x] jobs
- [x] exports

@shawnbot not a blocker for this PR, but I was having so write issues for the sqlite db. I was able to reference and process data from new tables that I created (via .sql files), but was unable actually manipulate said tables from the sqlite shell because they "didn't exist"

 /cc @coreycaitlin


gemfarmer:
This PR is ready for review. It shouldn't be merged until I clean up some of the files. 

I still need to reference some tasks in the Makefile, but I will do that after lunch

meiqimichelle:
@coreycaitlin This looks like it's on what was the 'states navigation' homepage. Should we merge this there, or does it need to live somewhere else? ie, the url right now is `.../states/` so we'd loose our nav page, I think. (am I interpreting that right @gemfarmer ?)

Also, I'm not sure if that was your intention with the national page content you were working with -- I vaguely remember the discussion around what was a 'nav home' and what data might look like for the national view.

gemfarmer:
@meiqimichelle yes, I have replaced that "navigation" view with the national page. 

We could keep `states/` as the navigation page and add a `states/US/` page to handle the national page. That being said, I think I like the direction it is headed because I think people are more likely to check out the national view this way.

coreycaitlin:
I kind of *don't* want to have a separate nav vs. national page, if we can avoid it, but I may be overly optimistic there.

Long-term, I was envisioning this living at /explore/, since we'll be moving nearly all non-map sections currently listed there to the How it Works section.

gemfarmer:
@coreycaitlin do you think this is good for now then?

meiqimichelle:
I talked to @coreycaitlin and understand now :) Also, the map is just at the bottom of the page, and I didn't scroll that far. Let's move it back to the top when we have a sec.

Issue #1573, opened 2016-07-05
Content: Budget and audit visualization


coreycaitlin:
Pull content from the IA's Budget & Audit visualization into a Markdown page.

meiqimichelle:
Pinging @avarnum so he can see how notification work :)

coreycaitlin:
Started pulling in content here: https://github.com/18F/doi-extractives-data/tree/budget-audit-content

coreycaitlin:
Notes from the PR:

Sidenote: We're using the how-it-works-ia branch to collect the new content that will go into a reorg of the How it Works section.

This PR includes content from the budget and audit visualization, which we've renamed "Oversight" to be parallel with other How it Works pages/sections. There are a few principles I kept in mind in revising this content for web:

- We always want to avoid duplicative content, both for usability reasons (finding similar or parallel information in two places can undermine user trust, because they aren't sure which is correct) and for maintainability. Whenever possible, we link to existing content, rather than re-creating it. This goes for content within the USEITI site (like disbursements), deeper agency pages (like ONRR's list of relevant laws), and agency home pages (for general info).
- To avoid confusion, I tried to make sure the scope of each section was internally consistent, and that all information about a topic lives in one place (rather than having content in multiple places on the same page.
- Each link should be clearly contextualized, so that users understand the scope of the link and what to expect if they click. For both usability and accessibility reasons, we never want to have different links with identical anchor text — and all anchor text should be as descriptive as possible. As part of this, I've moved more arcane links and most PDF links to footnotes.

coreycaitlin:
Some of the text hierarchy and sidenav work we've been doing is stuck over in `state-pages`, so it's a little tricky to grok the outline of the page by looking at the [preview](https://federalist.18f.gov/preview/18F/doi-extractives-data/how-it-works-ia/how-it-works/oversight/).

Here's the outline, in case it's tricky to see:

- Data validation and auditing
  - Company standards and oversight (includes public and private company standards)
    - Federal audits of company records and systems
  - Federal standards and oversight
- Data publication
  - Company payment reporting
  - Federal revenue reporting
- Agencies that manage revenue payments
  - Non-tax payments
    - Understanding how ONRR manages revenue (includes enforcement mechanisms and disbursements)
  - Corporate income taxes
- Notes

coreycaitlin:
Current outstanding questions for you, @avarnum:

[...]

PR #1591, opened 2016-07-12
Budget and audit content (AKA How it works > Oversight)

coreycaitlin:
Fixes issue(s) #1573.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/budget-audit-content/how-it-works/oversight/)

Sidenote: We're using the `how-it-works-ia` branch to collect the new content that will go into a reorg of the How it Works section.

This PR includes content from the budget and audit visualization, which we've renamed "Oversight" to be parallel with other How it Works pages/sections. There are a few principles I kept in mind in revising this content for web:

- We always want to avoid duplicative content, both for usability reasons (finding similar or parallel information in two places can undermine user trust, because they aren't sure which is correct) and for maintainability. Whenever possible, we link to existing content, rather than re-creating it. This goes for content within the USEITI site (like disbursements), deeper agency pages (like ONRR's list of relevant laws), and agency home pages (for general info).
- To avoid confusion, I tried to make sure the scope of each section was internally consistent, and that all information about a topic lives in one place (rather than having content in multiple places on the same page.
- Each link should be clearly contextualized, so that users understand the scope of the link and what to expect if they click. For both usability and accessibility reasons, we never want to have different links with identical anchor text — and all anchor text should be as descriptive as possible. As part of this, I've moved more arcane links and most PDF links to footnotes.

/cc @meiqimichelle for initial review


meiqimichelle:
I fixed a few small things, but otherwise looks like a great first version! :)

Issue #1588, opened 2016-07-12
Code: State page year selector


gemfarmer:
Extension of #1582 

## Expected Behavior
Should be able to select a year from a year range and change _all_ bar charts in a given section to the selected year. All the years across all the sections will have the same year range available in the selector. However, if there is no data in that section for a given year (say, we don't have datayet for 2016), the year should be visible in the selector, and a new view will load that 'grays out' the bars and provides a different intro sentence that explains the situation.
## Actual Behavior
The year selector does not have any functionality
## Steps to reproduce the behavior
TBD

meiqimichelle:
Looks like

![selector](https://cloud.githubusercontent.com/assets/9259185/16739744/341d6e96-4761-11e6-95c3-8e1911ec9dc1.gif)

Issue #1582, opened 2016-07-08
State page year selector design


ericronne:
The current design of interactive-data modules on the states page includes a selector, which defaults to the most-recent year for which we have data (2013 in the example below). Since the year is always different from the current year (2016), this has caused some confusion in our [armchair usability testing](https://docs.google.com/a/gsa.gov/document/d/1IO1pYisQ6G7rk2iKQif-k8Cy52-OvWWPXGPkEiFODn4/edit?usp=sharing) sessions. 

![image](https://cloud.githubusercontent.com/assets/9259185/16697605/17cc0a3e-4510-11e6-9cf4-c322c3a2a330.png)

One potential solution is to include the non-data years in the selector, but treat them differently....
![image](https://cloud.githubusercontent.com/assets/9259185/16697697/8d104e90-4510-11e6-82e4-bbd1b90aa7c9.png)

These non-data years could perhaps be clickable. If one was selected, an explanation of why we have no data could appear in the chart area.

 

- [x] Create a prototype
- [x] Test it

ericronne:
@meiqimichelle @coreycaitlin @gemfarmer fyi

ericronne:
![selector](https://cloud.githubusercontent.com/assets/9259185/16701462/920368dc-4525-11e6-9615-f9a5ab69696a.gif)

ericronne:
@coreycaitlin @meiqimichelle 

Will this design for the "no data" years give us problems with some categories, since we may have 2012 data for federal coal in Utah, eg, but not have any 2012 data for federal biomass in Utah?

hrumph.

ericronne:
Ok after chatting with @meiqimichelle, this "no data" approach may work after all (but we will have to explain some of the oddball situations in the copy, @coreycaitlin). Here's a modified gif, where i've instead grayed out the charts for non-data years. That way we won't have to worry about the page getting longer and shorter. NOTE: this design is old, but updating for the prototype would be time consuming.

Will do a quick test or two...

cc @gemfarmer 

![selector](https://cloud.githubusercontent.com/assets/9259185/16739744/341d6e96-4761-11e6-95c3-8e1911ec9dc1.gif)

ericronne:
[...]

Issue #1518, opened 2016-06-22
State pages: production with county data, code layout


meiqimichelle:
After #1492 is done, we should code the layout and make it look like the mock.

meiqimichelle:
Looks like this is ready to go here: https://github.com/18F/doi-extractives-data/issues/1492#issuecomment-228094983

meiqimichelle:
If you haven't started yet, you can use @ericronne 's latest latest: https://github.com/18F/doi-extractives-data/issues/1492#issuecomment-228156116

gemfarmer:
@meiqimichelle I'm going to wait on this until #1495 gets merged

gemfarmer:
@ericronne what is the grid here?

gemfarmer:
Within the primary column is it 4, 5, 3 on a 12 column grid? Or still 3, 3, 2 on a 8 column grid?

ericronne:
### 12 for the content area...
![image](https://cloud.githubusercontent.com/assets/9259185/16338801/7c0f1608-39e4-11e6-8116-6ce70e02b6a1.png)

ericronne:
Note: @gemfarmer 
We can keep the units as parentheticals in the chart titles for now, if that's quicker. That would align with how the charts are already designed...

![image](https://cloud.githubusercontent.com/assets/9259185/16339195/ff4b27a8-39e6-11e6-8db8-561e6aa210c0.png)

That would make for a simpler legend layout for horizontals, too. Win win...
![image](https://cloud.githubusercontent.com/assets/9259185/16339273/532ba4ec-39e7-11e6-9977-d3b5fd9b5cc5.png)

**This makes the titles longer, natch.**
I changed the bar chart to merely "coal production (tons) to keep it on one line. Not sure what all of the final titles will be, so i took the liberty. @coreycaitlin can answer all questions, i'm sure! 🔮 

We should root out the longest non-shortenable title and test the design. If we do have some breakers, it would be great if they were coded to ensure that the horizontal rules were always vertically aligned, like so (hopefully no nasty widows like this, tho!)...

![image](https://cloud.githubusercontent.com/assets/9259185/16339446/48774456-39e8-11e6-9298-3bbe427f125b.png)

gemfarmer:
Split part of this off into #1585 (handling wider states)

ericronne:
@gemfarmer here's the latest, reflective of the updated styles...

[...]

PR #1592, opened 2016-07-13
State pages: new layout, year selectors

shawnbot:
[:sunglasses: PREVIEW Utah](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages-new-layout/states/UT)
[:sunglasses: PREVIEW National page](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages-new-layout/states)

- General layout changes for #1518 
- Year selector: #1582 (design) #1588 (implementation)
  * Introduces the `<section is="year-switcher-section">` component, which manages syncing year selections between inputs, bar charts, and maps.

- Renames `<data-map>` to `<eiti-data-map>` (more component renaming forthcoming!)

- Removes the `line-chart-data.json` include in favor of just using the new `map_hash` template filter:

  ```html
  data='{{ some_data | map_hash: 'some_key' | jsonify }}'
  ```

/cc @ericronne @gemfarmer

### DIFF
--- a/_sass/blocks/_chart-list.scss
+++ b/_sass/blocks/_chart-list.scss
@@ -2,6 +2,13 @@
   @include font-size(0.8);
   @include outer-container();
 
+  display: flex;
+  -ms-flex-wrap: wrap;
+  -webkit-flex-wrap: wrap;
+  flex-wrap: wrap;
+  -ms-justify-content: space-between;
+  -webkit-justify-content: space-between;
+  justify-content: space-between;
   padding-top: $base-padding-jumbo;
 
   .chart-list_caption {
@@ -17,7 +24,9 @@
     font-size: inherit;
     font-weight: $weight-book;
     letter-spacing: 1px;
+    line-height: $standard-padding;
     margin-bottom: $base-padding;
+    padding-bottom: $base-padding-lite;
     text-transform: uppercase;
   }
 
@@ -28,7 +37,12 @@
   .chart-item {
     @include span-columns(4);
 
+    -ms-flex-direction: column;
+    -webkit-flex-direction: column;
     border-top: none;
+    display: flex;
+    flex-direction: column;
+    margin-bottom: $base-padding;
     margin-top: 0;
     padding-top: 0;
 
--- a/_sass/blocks/_chart-list.scss
+++ b/_sass/blocks/_chart-list.scss
@@ -99,13 +99,18 @@
     background-color: $mid-gray;
     border: 0;
     font-weight: $weight-book;
-
+    margin-right: $standard-padding;
+    margin-top: $standard-padding-lite;
   }
 
-  p {
-    @include span-columns(10);
+  .chart-description {
+    float: left;
+    width: 75%;
+    padding-left: 1.25rem;
+    border-left: 1px solid #ebebeb;
   }
 }
+
 
 
 .row-container {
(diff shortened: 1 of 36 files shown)


gemfarmer:
@shawnbot thanks for adding the selectors to the rest of the pages. Also, I like the use of map hash over that cumbersome json include! 😍 

I'm making some additions now, so let me know if you are going to be working on it so we don't conflict

gemfarmer:
#### Things that need to be added to follow up PR
---
- [ ] Federal production – map legend years need to update on year switch

gemfarmer:
@shawnbot sorry for the late response. I think that sounds good. For this PR I think I'm going to hold off on adding any additional "switchable" areas, but the todo list above should be the start of a list of things that need to update (and aren't currently) when the year is switched

_sass/blocks/_chart-list.scss
```diff
@@ -28,7 +37,12 @@
   .chart-item {
     @include span-columns(4);
 
+    -ms-flex-direction: column;
+    -webkit-flex-direction: column;
```
gemfarmer: It looks like flexbox unprefixed is currently [_almost completely_](http://caniuse.com/#feat=flexbox) covered, but adding prefixes will be nice for the few browsers that it only partially supports unprefixed

_sass/blocks/_chart-list.scss
```diff
@@ -2,7 +2,13 @@
   @include font-size(0.8);
   @include outer-container();
 
-  padding-top: $base-padding-jumbo;
+  display: flex;
+  -ms-flex-wrap: wrap;
+  -webkit-flex-wrap: wrap;
```
gemfarmer: see above ^^

_sass/blocks/_chart-list.scss
```diff
@@ -2,7 +2,13 @@
   @include font-size(0.8);
   @include outer-container();
 
-  padding-top: $base-padding-jumbo;
+  display: flex;
+  -ms-flex-wrap: wrap;
```
gemfarmer: See above ^^

_sass/blocks/_chart-list.scss
```diff
@@ -28,7 +38,13 @@
   .chart-item {
     @include span-columns(4);
 
+
     border-top: none;
+    display: flex;
+    -ms-flex-direction: column;
```
gemfarmer: see above comments on vendor prefixes ^^

gemfarmer:
PR in full
---
- incorporated @ericronne's new layout
- added year switcher component 
@shawnbot since yesterday:
- added the selector to all the sections on the state pages and national page
- layout adjustments to all sections, including national page
- cleaned up related plugins
- adjusted the national page revenue data to not include instances where there is no data

gemfarmer:
@shawnbot found a bug! I'm seeing the following on the national page. And the chart for all production "Inlet Scrubber" is messed up.

screen shot 2016-07-13 at 6 10 18 pm

Looking at the code currently on `state-pages`, this bug also exists, so likely came about when #1580 was merged. It looks like it is an issue with undefined data being fed into the bar-charts. Thoughts? Could also be pushed off to an additional issue ([layout spillover issue](#1597)

coreycaitlin:
@meiqimichelle and I went over this together and made a few content changes. This is looking amazing. 🎉

Issue #871, opened 2015-11-24
Refactor explore data JavaScript


shawnbot:
Right now the three explore data sections that we've done ([Federal Revenue](), [Federal Production]() and [GDP]() ($ only)) are all 100% bespoke. There are numerous patterns to extract from these, and before we get too far down the road of copying one exploration's code to start another one I'd like to factor out some of the generic bits into one or more reusable pieces of code. Here are some of the parts to abstract:

* State management: use [immutable data](http://jlongster.com/Using-Immutable-Data-Structures-in-JavaScript) under the hood to track and mutate state so that we only re-render when needed.
* Form: add listeners to form inputs that mutate the state.
* Routing: restore state from, and save state to, the URL hash.
* Data model: loads data based on the state, and can group or aggregate data in different ways.
* Scale: derives a scale function from a set of values.
* Choropleth map: sets the fill of each feature on the map, given data grouped by region and a scale. (This might be just an API addition for the `eiti-map` component.)
* Legend: given a scale, renders discrete steps on a legend.
* Data table: takes a list of data elements and renders a sorted (and sortable) HTML table from a template.
* Timeline: renders yearly data and handles handles range selection.

shawnbot:
This will make implementing @ericronne's new design for the explored data section on the home page _much_ easier.

shawnbot:
Sketching out some ideas:

###### Model
Our model at its most fundamental level defines where to get data given a `state` object. (In the explorations we've done so far, the state is an [Immutable.Map](https://facebook.github.io/immutable-js/docs/#/Map).) The consumer of a model's data just calls `model.load(state, callback)`. Optionally, the model may filter or otherwise manipulate the data after loading, but before passing it to the callback.

```js
var model = eiti.explore.model(function(state) {
  return eiti.data.path + 'revenue/regional.tsv';
});

// describe how state filters are applied to data
// (this is only called if `state.has('region')`)
model.filter('region', function(data, state) {
  var region = state.get('region');
  return data.filter(d => d.Region == region);
});

// we can update the timeline in our "year" filter
model.filter('year', function(data, state) {
  timeline.update(data);
  var year = state.get('year');
  return data.filter(d => d.Year === year);
});
```

###### App
The app is the "controller" in MVC parlance, and coordinates state changes between the model and the view (which, ideally, lives in the DOM):

```js
var app = eiti.explore.app()
  .root('#revenue') // default: 'body'
  .model(model);

// initialize the app with the state derived from location.hash
[...]

PR #1616, opened 2016-07-16
Introduce a React version of the data explorer

dmsnell:
Idea for refactoring mentioned in #871 

This is an exploratory commit to illustrate the first steps in
refactoring the data explorer into a React component. There are many
caveats to this PR. It only provides a minimal view, but one can imagine
that with a little bit of styling and extra code it could mirror the
existing table rows.

**Testing**
 - Navigate to the GDP explorer view
 - Scroll to the bottom of the existing table, there should be an unstyled table of rows with identical data from the existing table.
 - Open the filter pane and mess around with the sort-option (is this `$`/`%` toggle labelled wrong?) and the year. The data in the new table should update. Note that the select-a-region drop-down isn't supported in this PR.

**Caveats**
 - I have tried to minimize the site-wide impact of introducing React.
   This means, for example, that I have not introduced the JSX syntax.
   While JSX significantly improves the experience using React, it also
   requires some dependencies with webpack, such as Babel.

 - This is in no way supposed to be a feature-for-feature replacement.
   If the idea is valuable enough, it should be easy enough to cover all
   of the features, but it's not worth investing the time if this is a
   non-starter.

 - Some of the semantics of how I have put this together are the result
   of fitting React into the environment already provided. Using the
   location hash the way I do to get arguments is one example. In a full
   replacement, this would likely take a more elegant form.

 - Since I am less familiar with what general needs there are, I threw
   in some code that mixes concerns: hard-coding table column names and
   the like. This could easily be remedied with some generalization and
   is in fact a common pattern in React apps. For example, the
   `DataExplorer` renders `TableRow`s directly. To generalize, we would
   create some kind of `RegionalValuePercentRow` (with a much better
   name) and pass that as a prop into the `DataExplorer` as
   `rowRenderer` or similar.

 - I have also included several helper functions just because I don't know what already exists in the code and I don't know what is appropriate to reuse. See the data loader and parser, for example.

 - Against my preferences, I have limited this code strictly to ES5 syntax. Adding in fat-arrow functions and object/list destructuring would significantly improve the readability of portions of this PR.

 - For production-level introduction of React, the common libraries (such as React) can be bundled into a vendor JS bundle and the page-specific portions can continue to be loaded separately, so that for example, a page's front-matter could specify… `react: true` and the template would load the common bundle, while the `<script src="{{ site.base_url }}/js/pages/whatever.min.js"></script>` would remain just the way it is currently used.

cc: @shawnbot


dmsnell:
Note: I have removed the changes to the minified files in order to make it easier to see the _actual_ changes this PR brings in. Because I added a new file to the webpack config, however, a bunch of unimportant constants were changed in the minified files and they were making it hard to see what was important.

Point being, if you would like to test this branch, please run `webpack` to rebuild the minified bundles.

gemfarmer:
@dmsnell thanks again for supporting the site! I checked out the table and it works well. 😄 

I'm sad to say that the `explore/gdp/` page, along with most of the `explore` pages are actually being deprecated once the `state-pages` branch goes live, which is pretty soon. There is a `/states/STATE_NAME/#gdp` section for each state on that branch. [Check out Arizona's](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/AZ/#gdp)!

I apologize for not being explicitly clear with my last comment to you what are intentions are with the `state-pages` branch. We are (and correct me where I'm wrong @meiqimichelle) going to be removing the Production (all lands and federal), Revenue (all lands and federal), GDP, Exports, and Jobs pages from the explore section in favor of the state breakdown layout. Only **Reconciliation**, **Disbursements**, **Federal Income Tax**, and **Federal revenue by company** are going be remaining.

For this reason, I am reluctant to support and merge any additions to the soon-to-be deprecated pages, no matter how good the work is. 😞 

If even after this deflating comment you are still interested in contributing to the site, there are many outstanding issues for state-pages enhancements and bugs. I haven't used React enough to know myself, but there may be room for using it to enhance our [`year-switch-section`](https://github.com/18F/doi-extractives-data/blob/state-pages/js/components/year-switcher-section.js) component, which currently controls year navigation within the state-pages.

dmsnell:
> If even after this deflating comment you are still interested in contributing to the site

@gemfarmer you have crushed my spirit and all my drive to support government and open-source endeavors - I bite my thumb at you! :smile:

I didn't expect this to go much further anyway. It was mainly a showcase and I was interested in showing it off after diving into `List.js` trying to figure out the glossary thing. If there is some piece that you know maps well to modularized components, I would be more than happy to code it up again on a spot that isn't about to die.

If this were the case, I would be strongly tempted to pull in Babel if only to be able to use JSX syntax. But, Babel does do a great job enabling newer JS features that lag behind in the browser world (such as arrow functions, destructuring, `const`, etc…) while providing good source maps too. Do y'all have any thoughts on a transpiration step like that?

dmsnell:
> these pages are going away entirely, and being replaced by mostly static content that reacts to some simple user interactions

oh yes - I forgot to mention it - the individual state pages look great! (I checked out Arizona's)

> sometimes you just have to put these things in front of people before you can know that they don't work

_c'est la vie_

gemfarmer:
@dmsnell I'm glad you weren't crushed by my frankness.

We had a discussion today about open-source contributions, and we will (when time allows) be putting together a list of issues that are flagged `help-wanted`. I will be sure to tag you when we get here.

As for Babel and enabling JSX syntax, I am all for it as long as it doesn't significantly bloat or convolute the codebase.

dmsnell:
> We had a discussion today about open-source contributions, and we will (when time allows) be putting together a list of issues that are flagged help-wanted

Awesome! I'm glad to help - more worried about being a nuisance though so the `help wanted` label will be the teh bomb

> bloat or convolute

in my experience it shouldn't, but it's easy enough to try it out and see

PR #1623, opened 2016-07-19
Redirects invalid URLs to parent page if possible

### LINKED ISSUE #1210: Invalid URL redirects
Related to the 404 Error page #1206, if a user goes to an invalid URL, they should be redirected to a valid one. Probably the parent path.

`/explore/reconc`
goes to ->
`/explore/`

dmsnell:
Resolves #1210

Previously, all 404 views showed the 404 error page.

Now, as requested in #1210, the browser will automatically be redirected
to the parent path if indeed it isn't already navigating to a root-level
page.

**Example**:
 - `/explore/reconc/` redirects to `/explore`
 - `/explorer` does not redirect
 - `/explorer/` does not redirect

This was accomplished with a simple JavaScript snippet embedded in the
404 document. It _should_ work regardless of the domain and port being used.

In the second commit in this PR I updated the redirect behavior to switch from instantly redirecting to displaying an explanation message for a few seconds before redirecting. I felt like this was more helpful to a user so they wouldn't have to see a confusing flash of a 404. The message shows the invalid path, the next path after the redirect, and the number of seconds remaining until that redirect.

> For lots of fun, navigate to `/some/very/long/url/with/lots/of/parents/that/do/not/exist`
countdown404

**Normal Behavior**
normalredirect

**Notes**
 - This code _could_ be bundled into some JS file, but actually, since it only displays on the 404 page, I figured that it wasn't too bad to embed it right there.
 - The behavior here is pretty flexible. All we're doing is looking at the given location and mogrifying it.
 - I have no design skill and take no blame for poorly designed or worded messages :wink:
 - Although I saw no problems, and this 404 is the end of the line, I did dig deep to try and find if I was causing naming clashes in the global scope - figured it was okay for the reason above, that this is the end of the line for rendering.

cc: @gemfarmer

### DIFF
--- a/404.md
+++ b/404.md
@@ -4,8 +4,17 @@ layout: error
 title: 404
 subtitle: Page not found
 ---
+<div id="additional-404-message" />
 
 <script>
+var redirectMessage = function( oldPath, newPath, pathname, remainingDelay ) {
+	return (
+		'No page found at <a href="' + oldPath.href + '">' + oldPath.pathname + '</a>.<br />' +
+		'Automatically redirecting to <a href="' + newPath + '">' + pathname + '</a> ' +
+		'in ' + remainingDelay + ' seconds.'
+	);
+};
+
 var path = document.location;
 var parentPath = path
 	.pathname
@@ -22,7 +31,19 @@ var nextLocation = [
 ].join( '' );
 
 // If we aren't already at a root-level page
+// then redirect to the parent after a short
+// explanatory delay
 if ( parentPath.length ) {
-	document.location.replace( nextLocation );
+	var additionalMessage = document.getElementById( 'additional-404-message' );
+	var countDown = function( seconds ) {
+		if ( 0 === seconds ) {
+			document.location.replace( nextLocation );
+		}
+
+		additionalMessage.innerHTML = redirectMessage( path, nextLocation, parentPath, seconds );
+		setTimeout( countDown.bind( this, seconds - 1 ), 1000 );
+	}
+
+	countDown( 5 );
 }
 </script>
--- a/404.md
+++ b/404.md
@@ -7,6 +7,7 @@ subtitle: Page not found
 <div id="additional-404-message" />
 
 <script>
+(function(){
 var redirectMessage = function( oldPath, newPath, pathname, remainingDelay ) {
 	return (
 		'No page found at <a href="' + oldPath.href + '">' + oldPath.pathname + '</a>.<br />' +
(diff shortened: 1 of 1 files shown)


dmsnell:
Also, I did not build this from `state-pages` because when I pulled the latest version, I was receiving this error on `bundle exec jekyll serve`

```
Configuration file: /Users/dmsnell/PhpstormProjects/doi-extractives-data/_config.yml
            Source: /Users/dmsnell/PhpstormProjects/doi-extractives-data
       Destination: /Users/dmsnell/PhpstormProjects/doi-extractives-data/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
  Liquid Exception: undefined method `to_h' for []:Array in _includes/location/section-gdp.html, included in _layouts/state-page.html
jekyll 3.0.5 | Error:  undefined method `to_h' for []:Array
```

gemfarmer:
@dmsnell this is awesome! It works well locally. This will give us all the more reason to bug our coworkers working on Federalist to fix the issue that is making out current 404 look like:

screen shot 2016-07-19 at 10 41 08 am

I'm curious to know more about not redirecting for first level paths? Are you thinking it would be too disorienting to redirect home if someone is trying to visit `/explore/` and mistakenly types `/explorer/`?

Other than that, I'm definitely down to merge this. It won't be seen live until Federalists fixes their 404 handling, but good to have for when it is visible.

Thoughts @meiqimichelle ?

gemfarmer:
@dmsnell as to your Jekyll error above, my local `state-pages` works without error.

My guess is that there was a new plugin added and that you need to `rm -rf _site` and restart the server. But it could also be something completely different. If this doesn't work let me know and I can try to dig deeper.

dmsnell:
> I'm curious to know more about not redirecting for first level paths? Are you thinking it would be too disorienting to redirect home if someone is trying to visit /explore/ and mistakenly types /explorer/?

@gemfarmer this was simply my interpretation of the need for the issue. I thought that if we redirect _ad infinitum_ then there won't ever really be a 404 page. On second thought, I guess it doesn't matter too much, so I updated the behavior in cea432b

Either way, it's trivial to change its behavior.

A further enhancement that could be cool for another PR would be to have Jekyll build a JSON file enumerating the valid paths. With a little more thought about it, the 404 page could load that data and take a guess at what any given mismatched path was supposed to be. For example, `/explorer/all-lands-production` could redirect to `/explore/all-lands-production` instead of walking up the parent chain to root.

dmsnell:
> you need to rm -rf _site

@gemfarmer apparently that's what I needed. thanks!

gemfarmer:
@dmsnell 
> For example, /explorer/all-lands-production could redirect to /explore/all-lands-production instead of walking up the parent chain to root.

I think this would be a great enhancement for the functionality. Having a record of valid paths would also allow us to trim down `/something/unwieldy/and/completely/incorrect/` to `/` instead of slowly backtracking via timed redirects...

For now, this is great! I hope that we can get Federalist to fix their error page handling so it can get some user accolades. 

Mergin! 💥

Issue #1602, opened 2016-07-14
Bar charts bug


gemfarmer:
## Expected Behavior
No errors to occur and bar charts render properly
## Actual Behavior
I'm seeing the following error on the national page. 

<img width="1153" alt="screen shot 2016-07-13 at 6 10 18 pm" src="https://cloud.githubusercontent.com/assets/4803473/16822847/2aa0b808-4925-11e6-8acb-8b4d820287b1.png">

And the chart for all production "Inlet Scrubber" is messed up. Notice the two bar heights on 2005 and incorrect year selection.

<img width="240" alt="screen shot 2016-07-14 at 10 42 26 am" src="https://cloud.githubusercontent.com/assets/4803473/16845763/aecc6bec-49af-11e6-9cf1-67337f4ae4ab.png">

Looking at the code currently on `state-pages`, this bug also exists, so likely came about when #1580 was merged. It looks like it is an issue with undefined data being fed into the bar-charts. Thoughts? Could also be pushed off to an additional issue ([layout spillover issue](https://github.com/18F/doi-extractives-data/issues/1597))

## Steps to reproduce the behavior

PR #1631, opened 2016-07-20
Weird FFx/Safari bars bug

gemfarmer:
Fixes issue(s) #1602

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/bars-bug/states)

- fixed bug that was preventing maps and charts to render on safari. This was the result of the `<eiti-data-map>` component script existing on the national page without any elements to reference.
- fixed bar chart bug that was causing some years to have multiple bars. This was the result of an inconsistency between the `year_range` for that chart and the years in its affiliated dataset. I filtered the data so that only years within that range were accounted for and rendered.

/cc @meiqimichelle @shawnbot

### DIFF
--- a/js/components/eiti-bar-chart.js
+++ b/js/components/eiti-bar-chart.js
@@ -71,6 +71,8 @@
         .key(function(d) { return d.x; })
         .rollup(function(d) { return d[0]; })
         .entries(data);
+
+
     } else {
       values = Object.keys(data).reduce(function(map, key) {
         map[key] = {x: +key, y: data[key]};
@@ -81,9 +83,9 @@
       });
     }
 
-    // console.log('data:', data, 'values:', values);
 
     var xrange = this.xrange;
+
     if (!xrange) {
       xrange = d3.extent(data, function(d) {
         return +d.x;
@@ -100,6 +102,12 @@
       if (!values[x]) {
         data.push({x: x, y: 0});
       }
+    });
+
+    // filter the data so that only values within the domain
+    // are included in calculations and rendering
+    data = data.filter(function(d){
+      return xdomain.indexOf(d.x) > -1;
     });
 
     var extent = d3.extent(data, function(d) { return d.y; });
@@ -136,7 +144,11 @@
     bars.exit().remove();
 
     bars.attr('transform', function(d) {
-      return 'translate(' + [x(d.x), 0] + ')';
+      if (x(d.x)) {
+        return 'translate(' + [x(d.x), 0] + ')';
+      } else {
+        console.warn(d.x, 'not in bar chart domain')
+      }
     });
 
     bars.select('.bar-value')
--- a/js/components/eiti-bar-chart.js
+++ b/js/components/eiti-bar-chart.js
@@ -106,7 +106,7 @@
 
     // filter the data so that only values within the domain
     // are included in calculations and rendering
-    data = data.filter(function(d){
+    data = data.filter(function(d) {
       return xdomain.indexOf(d.x) > -1;
     });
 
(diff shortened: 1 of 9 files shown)


gemfarmer:
@shawnbot it was [specifically the rendering of the legend](https://github.com/18F/doi-extractives-data/pull/1631/files#diff-5efbdcea04d96182d0804ce4483fe460R90)

gemfarmer:
There was a line in the `d3-svg-legend` script that was erroring out because the context that it was trying to render it in was `null`

js/components/eiti-bar-chart.js
```diff
@@ -102,6 +104,12 @@
       }
     });
 
+    // filter the data so that only values within the domain
+    // are included in calculations and rendering
+    data = data.filter(function(d){
```
shawnbot: Missing a space between the `)` and `{`. I'm wondering why Hound didn't spot this...

js/components/eiti-bar-chart.js
```diff
@@ -136,7 +144,11 @@
     bars.exit().remove();
 
     bars.attr('transform', function(d) {
-      return 'translate(' + [x(d.x), 0] + ')';
+      if (x(d.x)) {
```
shawnbot: Is this this right test? Presumably `x(d.x)` could evaluate to zero, right?

js/components/eiti-data-map.js
```diff
@@ -87,46 +87,52 @@
...
-            } else  {
-              // trim spacing between swatches that are visible
-              var translateHeight = (i * cellHeight) - (count * cellHeight);
-              cells[0][i].setAttribute('transform',
-                'translate(0,' + translateHeight + ')');
-            }
-          });
-          // end consolidation
-          // end map legend
+          if (svgLegend[0][0]) {
```
shawnbot: More succinctly: `!svgLegend.empty()`

js/components/eiti-data-map.js
```diff
@@ -87,46 +87,52 @@
...
+                // trim spacing between swatches that are visible
+                var translateHeight = (i * cellHeight) - (count * cellHeight);
+                cells[0][i].setAttribute('transform',
+                  'translate(0,' + translateHeight + ')');
+              }
+            });
+            // end consolidation
+            // end map legend
+          } else {
+            console.warn('<eiti-data-map> does not exist on this page.')
```
shawnbot: This should say that the legend doesn't exist, right?

gemfarmer:
I should rephrase that. If the `eiti-data-map` component exists, but the class that corresponds to the svg-legend, `.legend-svg`, does not exist, it was throwing an error.

js/components/eiti-data-map.js
gemfarmer: how about `this <eiti-data-map> does not have an associated svg legend.`?

js/components/eiti-bar-chart.js
gemfarmer: maybe ?

```if (x(d.x) && x(d.x) !== 0) {```

Comment (gemfarmer) [inline js/components/eiti-bar-chart.js]: Let me play around with this.

Comment (gemfarmer) [inline js/components/eiti-bar-chart.js]: Actually, it looks like this was fixed by the [filter function above](https://github.com/18F/doi-extractives-data/pull/1631#discussion-diff-71572664R109) 😄

**Comment by gemfarmer** [comment]
@shawnbot does this look better now?

File: js/components/eiti-data-map.js [hunk 5]
```diff
@@ -87,46 +87,50 @@
...
+              .ascending(true)
+              .labelDelimiter('-')
+              .shapePadding(6)
+              .scale(scale);
+
+            svgLegend.select('.legendScale')
+              .call(legend);
+
+            // reverse because the scale is in ascending order
+            var _steps = d3.range(0, 9).reverse();
```
shawnbot: Not a blocker, but why `_steps` here and not just `steps`?

gemfarmer: I think I did this to not overwrite `steps` from above, but I guess I'm not totally sure that would be problematic, come to think of it. I merged for now, because this came up in standup, but should be refactored later.

Issue #1509, opened 2016-06-20
Create print stylesheet


gemfarmer:
Sub task of #1425 

Make a print style sheet for the site that enhances the styling in print format.

gemfarmer:
@meiqimichelle do you think we should branch this off `dev` or `state-pages`? Do we want to get this released asap is what I'm really asking.

meiqimichelle:
This doesn't have to go ASAP, `state-pages` is fine!

gemfarmer:
👍

gemfarmer:
Closed with #1633

Issue #1390, opened 2016-05-09
State profile pages into sharable form


meiqimichelle:
We're pretty close. We should probably just need to move the rest of the content work from the doc to the live page. Goal: finish by Wednesday, May 9.

coreycaitlin:
✈️  [Preview link for reference](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/UT/) ✈️ 

Okay! Updates:

- I've pulled in most of the framing content from [this doc](https://docs.google.com/document/d/1_l5WbEQZAOfO-oUD3P1t06G1OjA7jeYMvf2DLqm9ZYs/edit#).
- Things that still need more work:
   - Disbursements content, which I haven't pulled in yet
   - The distinction between the two kinds of jobs data
   - Links to the datasets/documentation pages, and any notes about the scope/source of data that should be on this page (if we can keep that all on the docs pages, I'm 👍  for that)
   - The chart that shows fee rates per commodity by stage isn't in the prototype yet — not sure how we'll break this down to show only the relevant info
  - KEY FACTS: I think we can make this much, much shorter, and we'll need a contextual narrative. Hopefully usability testing can help inform what we need here!

I've been focused on building this page for non opt-in states with no offshore revenue. We'll need to do a bit more thinking about how to handle states with significantly more state data and how to talk about offshore revenue (not to mention which states that'll apply to, other than CA and AK).

With a little bit of styling and some stripped-back tables/charts, this should be ready to show to users. 🎉

meiqimichelle:
Let's break this out into separate issues and close this one.

coreycaitlin:
- Disbursements content, which I haven't pulled in yet --> #1461 
- The distinction between the two kinds of jobs data --> #1462 
- Links to the datasets/documentation pages, and any notes about the scope/source of data that should be on this page (if we can keep that all on the docs pages, I'm 👍 for that) --> #1463 
- The chart that shows fee rates per commodity by stage isn't in the prototype yet — not sure how we'll break this down to show only the relevant info --> #1464 
- KEY FACTS: I think we can make this much, much shorter, and we'll need a contextual narrative. Hopefully usability testing can help inform what we need here! --> #1465

coreycaitlin:
And:

- how to handle states with significantly more state data --> #1467
- how to talk about offshore revenue --> #1466

coreycaitlin:
Closing out this issue now that we have smaller issues to cover all the parts!

PR #1633, opened 2016-07-20
Print stylesheet

gemfarmer:
Fixes issue(s) #1509, #1390 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/print-stylesheet/states/)

To review:
---
- open link
- try to print
- check out print preview on various pages

---
- makes second print.css stylesheet that styles the site when it is printed
- I removed all exclusively interactive features of the site, except for links. 
- links: add URL as text following a link in this format `Some link (www.somelink.com)`
- left-aligned most things
- pages covered
  - all content pages
  - how it works and explore landing pages
  - state and national pages
  - home page
  - about page
- For the most part tried to keep the default spacing, font sizes and alignment
- The hardest part was making the page break in the correct location. The breaks that I have established are far from perfect – there are plenty of messy examples I have found – but this is a good place to start.

__NOTE: I used the `!important` flag quite a bit. Normally I would applaud hound for catching me, but since these are supreme overrides and I don't want to mess around with finding the appropriate specificity for each instance I am trying to override, `!important` is more helpful than harmful IMHO.__

/cc @meiqimichelle @coreycaitlin

### DIFF
--- a/_sass/print/_home.scss
+++ b/_sass/print/_home.scss
@@ -1,7 +1,8 @@
 .carousel,
 .eiti-tabs,
 .card-image_link,
-.eiti-tab-panel .explore_home-button {
+.eiti-tab-panel .explore_home-button,
+.home-bottom-links {
   display: none !important;
 }
 
--- a/_sass/print/_home.scss
+++ b/_sass/print/_home.scss
@@ -1,0 +1,41 @@
+.carousel,
+.eiti-tabs,
+.card-image_link,
+.eiti-tab-panel .explore_home-button {
+  display: none !important;
+}
+
+.card {
+
+  display: block;
+  float: none;
+  height: inherit !important;
+
+  img {
+    display: none !important;
+  }
+}
+
+.eiti-tab-panel {
+
+  display: block;
+  float: none;
+  margin-bottom: $base-padding-jumbo;
+
+  img {
+    margin: 0;
+    padding: 0;
+  }
+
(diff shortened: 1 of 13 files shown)


_sass/print/_home.scss
```diff
@@ -0,0 +1,42 @@
+.carousel,
```
gemfarmer: I'm not sure what this flag is referencing.

css/print.scss
```diff
@@ -0,0 +1,22 @@
+---
```
gemfarmer: This is because of Jekyll syntax.

ericronne:
Nice! What do you think about making all text black, and reducing its size a bit?

gemfarmer:
I think those are reasonable suggestions. I also revisited the home page and some of the first pages that I changed and realized that the page is breaking in weird ways :(

ericronne:
Is it easy to force charts and their headers to stay grouped, so they don't get split by page breaks?

gemfarmer:
@ericronne I wasn't having an easy time of it. I'll take another stab though. It doesn't look great right now. Better than before, but not great.

ericronne:
If it's not easy, my vote is _onward_ to the Next Thing.

gemfarmer:
💥 . @ericronne I ended up keeping things in color because my timebox for print stylesheeting ran out :/

PR #1645, opened 2016-07-22
Fix coal all lands production data for 2012 onward

### LINKED ISSUE #1644: Nationwide Coal production for 2012 looks off
Is it possible that there's a problem with the data from EIA? Our data tests _should_ cover this, but it needs to be looked at, because I suspect something is wrong [with our national data](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-pages/states/#all-production-coal-short-tons):

image

shawnbot:
[😎 Preview](https://federalist.18f.gov/preview/18F/doi-extractives-data/fix-coal-rollups/states/)

- I updated the 2012-2014 coal data [from EIA](http://www.eia.gov/coal/data.cfm#production) and removed the extraneous headers from `data/all-production/input/coal-????.tsv`. This appears to have fixed #1644:

  image

- We no longer need the unioned `data/all-production/input/coal.tsv`, since we transform all of the yearly data into a single intermediary `.ndjson` file then pipe _that_ through `tables`, which is faster than piping in each yearly data file one by one.
- The source data (from EIA) for all lands coal production contains state names like "Kentucky (Northern)" and "Pennsylvania (Anthracite)". This change updates the transform for coal data to strip the parentheses, and commits the newly rendered data.


gemfarmer:
@shawnbot looks good. :)  I didn't think there was a gap in the coal data!

Issue #1626, opened 2016-07-19
Remaining county chart issues


gemfarmer:
Spillover from #1604 

- [x] if scale doesn't exist in the default year that the chart/map is rendered in (2013), then no scale exists and swatches color themselves.
- [x] county charts order by value

gemfarmer:
Closed with #1649

Issue #1597, opened 2016-07-13
Layout spillover


gemfarmer:
#### Remaining work from #1592, #1518 
---
- [x] Federal production data isn't working on some states (see Arizona and Washington) (perhaps it is the states that have the random 100 years of data for a few commodities) --mhz
- [x] map legend years need to update on year switch
- [x] Map legend filtering needs to update without white space
- [ ] Color swatches on map should based on absolute (rather than relative) scale
- [ ] ~~add year switcher functionality to the mega chart (probably means making the megachart a component)~~ We don't need to look at every year here. --mhz
- [ ] ~~bar chart sentences need to handle `$0` values~~ discussed [here](https://github.com/18F/doi-extractives-data/issues/1556#issuecomment-230521577)
- [ ] ~~bar charts need to handle 0 values~~ new issue for this [here](https://github.com/18F/doi-extractives-data/issues/1675)
- [x] when sections don't have any data for any years, hide selector, bar chart and map
- [ ] fix uneven heights on `.chart-title` element (easiest solution here is to make all headers short enough to fit on one line)
- [x] get rid of "Details" toggles that are no longer needed. Remove commented out code.
- [x] wide maps should have horizontal legend [see this comment](https://github.com/18F/doi-extractives-data/issues/1492#issuecomment-228353485)
#### Things that need to update when a year is switched (and currently aren't updating)
---
- [x] legend description
- [x] Section description. Currently if the page is rendered on a year, that text is set to that year, even if another year might have different associated text
- [x] county-level bar charts

ericronne:
I'm marking up some typographic adjustments, but here are a few other tweaks for starters (mostly re bar styling). Thx @gemfarmer! ...

- [ ] Create a new section header, "Production in the entire state of Utah" (see image below).
- [ ] Extra space should go above single-line heads, not below it.
- [ ] Numbers and text in surrounding sentences seem large.
- [ ] If @coreycaitlin agrees, delete the periods after data sentences
- [ ] Can we slightly reduce the vertical space between the tops of the bars and the data sentence above them?
- [ ] ENHANCEMEHT: testing surfaced the need for a Y-axis label, which could look like the inset design below. 
- [ ] Charts are too close horizontally, see image here (padding on right)

![image](https://cloud.githubusercontent.com/assets/9259185/16876735/b15713ce-4a6a-11e6-8cc6-0758e8ebf158.png)

ericronne:
@gemfarmer 
[...]

PR #1649, opened 2016-07-25
Map and chart adjustments

gemfarmer:
For #1597 and #1626.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/remaining-chart-issues/states/UT/)

- chart and map legends update year
- chart and map legends include units
- map description displays "no data text"
- map legend adjusts without gaps in scaling
- county chart table closes when a year has no data
- change delimiter to `to` from `-` in instances where there are negative values
- when years don't have values on 2013, map and chart values for other years still populate when the year is updated

/cc @meiqimichelle @coreycaitlin @shawnbot

### DIFF
--- a/_includes/county-map.html
+++ b/_includes/county-map.html
@@ -34,16 +34,19 @@
         {% assign fips = county[0] | pad_left: 5, '0' %}
         {% if include.href %}<a xlink:href="{{ include.href | replace: '%', fips }}">{% endif %}
         {% assign value = county[1] | get: keys %}
-        {% if value != nil %}
+
+        {% assign year_values = null %}
+        {% if include.years %}
+          {% assign year_values = county[1] | get: include.years %}
+          {% if include.years_property %}
+            {% assign year_values = year_values | map_hash: include.years_property %}
+          {% endif %}
+        {% endif %}
+
+        {% if year_values != nil and year_values != null %}
         <g class="county feature"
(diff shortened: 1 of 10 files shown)


_includes/county-map.html
```diff
@@ -34,16 +34,19 @@
...
+
+        {% assign year_values = null %}
+        {% if include.years %}
+          {% assign year_values = county[1] | get: include.years %}
+          {% if include.years_property %}
+            {% assign year_values = year_values | map_hash: include.years_property %}
+          {% endif %}
+        {% endif %}
+
+        {% if year_values != nil and year_values != null %}
```
shawnbot: I'm pretty sure that `{% if year_values %}` would suffice here.

js/components/aria-toggle.js
```diff
@@ -31,7 +43,30 @@
...
+            case 'aria-expanded':
+              if (value === false) {
+                collapse(this);
+              } else if (value === true) {
+                expand(this);
+              }
+          }
+        }},
+
+        'aria-expanded': {
```
shawnbot: I think the property should just be called `expanded` or `ariaExpanded` so that you can access it without brackets (`element.expanded` vs. `element['aria-expanded']`).

gemfarmer:
@shawnbot I addressed your comments in 8927f62

PR #1663, opened 2016-07-26
update kern county text and footnotes

coreycaitlin:
Update Kern County case study text and data.


coreycaitlin:
Just kidding — forgot to change the base branch.

PR #1687, opened 2016-07-31
Adjust mobile display styling for "diving previews"

### LINKED ISSUE #988: Visual Bug: Mobile view of diving previews on how it works landing
The four preview images don't work very well on mobile on the [how it works page](https://useiti.doi.gov/how-it-works/#process)...
image

What do you think of this workaround, @meiqimichelle?...
image

dmsnell:
Resolves #988

Previously it was noted that the graphics for the different resource
types on the [How It Works page](http://localhost:4000/how-it-works/)
look funny on narrow screens. Because of the width of the displays, the
text column was too narrow.

This patch incorporates @ericonne's suggestion for an adjustment on
screens smaller than a certain width (in this case, `tiny-down`).
Included here are cropped icons from the corresponding "diving"
background images.

**Before**
howitworkedonmobile

**After**
howitworksonmobile

**Caveats**
 - The CSS changes I introduced appear to get-the-job-done but someone
   with real skillz should probaby look over the changes and suggest a
   better way to accomplish it. I'm not confident that this is good
   styling the way I have written it.

 - Because the mockup showed the landing header beneath the icons but
   the header HTML is outside of the scope that the icon lives in, I
   have used _creative margins_ to swap their spots vertically instead
   of doing something like rendering the icon as an `` lexically
   above the heading while dropping the existing background image
   reference entirely (for small screens).

 - The text in all of the mockups appears to be of a lighter-gray color
   while the current text in `dev` is black. I have not addressed this
   since it appeared ancillary to the actual issue.

 - For what it's worth, I'm not sure I understand what exactly was bad on mobile
   screens before, as it looked fine to me `¯\_(ツ)_/¯`

cc: @ericronne


gemfarmer:
Reviewed. This looks great, @dmsnell! As far as I'm concerned, it is ready to merge once those hound issues above are addressed. It's persnickety, I know, but gotta please our linting overlords 🙇 

@ericronne, what do you say about the text color? Should the the text be a lighter gray, or do we want to keep as is for now?

dmsnell:
thanks @gemfarmer - the hound overlords have been pleased. Woof!

gemfarmer:
Perfect! @ericronne is fine with the text color as is.

PR #1688, opened 2016-07-31
Add pagination to search results

### LINKED ISSUE #1172: Add pagination to curated search
Currently, we're arbitrarily cutting off the number of [search results](https://useiti.doi.gov/search-results/?q=oil) since we don't have pagination. We should fix this.

dmsnell:
**Not ready for merging!**

Resolves #1172

Modifies [search.js](https://github.com/18F/doi-extractives-data/blob/5692a68023b921dcb82fc22ba1ea690acc8db5ea/js/components/search.js) to add previous and next buttons to search results when there are
sufficient results to allow it.

While working on the code to add the previous and next links, I took
considerable liberty in refactoring the existing code. Since I have only
begun to jump into contributions of any scale here, I would really
appreciate feedback on which such liberties I can take and how my
personal style choices mesh with the project/team. It is not my
intention to overrule existing style guidelines, but I hope to learn
them better through this PR.

**The API call is a blocker to this PR**
The search at api.data.gov was consistently returning a `504 Gateway
Timeout` while I was working. Since I could not see the actual search
results, I have guessed at certain elements when writing my code.

For example, with `size=10&from=13`, the code expects to get the 14th
through 24th elements from the search results (14 vs. 13 due to
zero-based-indexing).

Also, with no results, there is no way to see the prev and next links
that I am adding without fidgeting with the code to fib the results. To
test, I fibbed the results by settings `timeout: 200` on the `$.ajax()`
query so that it would quickly fail. Then, I placed a `.fail()` handler
on the query with my paging code and a made up value for the
`json.hits.total` value. The links appeared to work fine under these
conditions.

## Previews

You might notice some oddities in these previews. First of all, Chrome has been started with `web-security` disabled, allowing unauthorized CORS requests (had to do this to ping api.data.gov from `localhost`). Next, there are no search results yet there are links; this is due to the "fibbing" I described above. Further, the tooltips are active because I was hovering over the links when I took the screenies. Finally, you can see the new parameters in the query string: `size` and `from`, which coordinate the paging.

**On first page**
nextlink

**On a "middle" page**
bothlinks

**On last page**
prevlink

cc: @meiqimichelle @gemfarmer

### DIFF
--- a/js/components/search.js
+++ b/js/components/search.js
@@ -118,7 +118,7 @@ $( document ).ready(function() {
         $('.search-no-results').remove();
       }
 
-      $.each(json.hits.hits, function(i, hit){
+      var hits = json.hits.hits.map(function(hit) {
         var tags = (hit._source.tag || [])
           .map(function(tag) {
             return (
@@ -128,21 +128,22 @@ $( document ).ready(function() {
                 '</a>&nbsp;' +
               '</span>'
             );
-          })
-          .join('');
+          });
 
         var contentType = getContentTypeGlyph(hit._source.content_type);
 
-        $('.search-results-container').append(
+        return (
           '<article class="search-result-list">' +
             '<h1><a href="' + hit._source.url + '" target="_blank">' +
               hit._source.title +
             '</a></h1>' +
             '<p>' + hit._source.description + '</p>' +
-            '<p>Tagged&nbsp;/' + tags + '</p>' +
+            '<p>Tagged&nbsp;/' + tags.join('') + '</p>' +
           '</article>'
         );
       });
+
+      $('.search-results-container').append( hits.join('') );
 
       var paging = {
         needsPrevLink: from > 0,
--- a/js/components/search.js
+++ b/js/components/search.js
@@ -5,85 +5,165 @@ $( document ).ready(function() {
   //
   // By Sean Herron <sean@herron.io>
   //
-  function GetURLParameter(sParam) {
-    var sPageURL = window.location.search.substring(1);
-    var sURLVariables = sPageURL.split('&');
-    for (var i = 0; i < sURLVariables.length; i++) {
-      var sParameterName = sURLVariables[i].split('=');
-      if (sParameterName[0] == sParam) {
-        return sParameterName[1];
-      }
-    }
+
+  /**
+   * Returns parameter from GET query string
+   *
+   * Note that in application/x-www-form-urlencoded strings
+   * (such as the GET query), a `+` will replace `%20` so we
+   * need to _untranslate_ that back into `%20` before decoding.
+   * The _actual_ `+` is safely encoded as `%2B`, however, so
+   * we do not have to worry about losing `+`s
+   *
+   * @param {string} param - Which value to extract by name in name=value
+   * @returns {string} Value of first requested param else ''
+   */
+  function getQueryParam(param) {
+    return decodeURIComponent(window.location.search
+      .slice(1) // drop the leading '?'
+      .split('&') // split into name=value pairs
+      .map(function(pair) { return pair.split('=') }) // split pairs into [name, value]
+      .filter(function(pair) { return pair[0] === param }) // return matching pairs
+      .map(function(pair) { return pair[1] }) // but only return the value
+      .concat( '' ) // add a default in case didn't find any
+      .shift() // return first match or empty string if none available
+      .replace(/[+]/g, '%20') // untranslate `+` into `%20` (see above)
+    );
   }
 
-  var query = GetURLParameter('q') || '',
-    apiKEY = eiti.beckleyApiKey;
+  /**
+   * Returns URL for search request
+   *
+   * @param {string} query - Raw string holding search terms/query
+   * @param {string} apiKey - Raw string holding API key for api.data.gov
+   * @param {object} [options] - Optional search parameters
+   * @param {number} [options.size=10] - How many results to fetch
+   * @param {number} [options.from=0] - Offset to start from for paging
+   * @returns {string} URL for desired search
+   */
+  function apiSearchUrl( query, apiKey, options ) {
+    var q = encodeURIComponent( query );
+    var size = (options && options.size) || 10;
+    var from = (options && options.from) || 0;
 
-  $('.search-string').append(query.replace(/%20/g,' '));
-  $('.site-search-text').attr('value',query.replace(/%20/g,' '));
-  if (query) {
-    $('input.q').val(query);
-    $('.search-result-list').append('<div class="loading"><span class="glyphicon glyphicon-refresh"></span> Loading</div>');
-    var url = 'https://api.data.gov/beckley-federalist/v0/resources/eiti/?q=' + query + '&size=50&from=0&api_key=' + apiKEY;
-    $.ajax({
-      url: url,
-      cache: false,
-      dataType: 'json'
-    })
-      .done(function(json) {
-        $('.loading').remove();
-        $('.search-results-count').append( json.hits.total + ' search results');
-        if (json.hits.total == 0){
-          $('.search-no-results').show();
-        }
-        else{
-          $('.search-no-results').remove();
-        }
-
-        $.each(json.hits.hits, function(i, hit){
-
-        var result_description = hit._source.description;
-
-          var tags = '';
-          if(hit._source.tag) {
-
-            $.each(hit._source.tag, function(i, tag) {
-              tags += '<span class="search-result-list-tag">&nbsp;<a href="../search-results/?q='
-              + tag + '" title="Search for '
-              + tag +'">'
-              + tag + '</a>&nbsp;/</span>';
-            });
-          }
-
-          var content_type = '';
-          if(hit._source.content_type == 'text/html') {
-            content_type = '<span class="glyphicon glyphicon-link"></span> Website';
-          }
-          else if(hit._source.content_type == 'application/pdf') {
-            content_type = '<span class="glyphicon glyphicon-file"></span> PDF';
-          }
-          else {
-            content_type = '<span class="glyphicon glyphicon-file"></span> ' + hit._source.content_type + '';
-          }
-          $('.search-results-container').append('<article class="search-result-list"><h1><a href="'
-            + hit._source.url
-            + '" target="_blank">'
-            + hit._source.title
-            + '</a></h1>'
-            + '<p>'
-            + result_description
-            + '</p>'
-            +'<p>Tagged&nbsp;/'
-            + tags
-            + '</p>'
-            + '</article>');
-        });
-      });
+    return (
+      'https://api.data.gov/beckley-federalist/v0/resources/eiti/?' +
+      'q=' + q + '&size=' + size + '&from=' + from + '&api_key=' + apiKey
+    );
   }
-  else {
+
+  function localSearchUrl( query, options ) {
+    var q = encodeURIComponent( query );
+    var size = (options && options.size) || 10;
+    var from = (options && options.from) || 0;
+
+    return '/search-results/?q=' + q + '&size=' + size + '&from=' + from;
+  }
+
+  /**
+   * Returns HTML string of Glyph for given content type
+   *
+   * @param {string} contentType - given document content type
+   * @returns {string} glyph and label for content type or generic default
+   */
+  function getContentTypeGlyph(contentType) {
+    var contentTypes = {
+      'text/html':       '<span class="glyphicon glyphicon-link"></span> Website',
+      'application/pdf': '<span class="glyphicon glyphicon-file"></span> PDF',
+      'default':         '<span class="glyphicon glyphicon-file"></span> ' + contentType
+    };
+
+    return contentTypes[contentType]
+      ? contentTypes[contentType]
+      : contentTypes['default']
+  }
+
+  var query = getQueryParam('q'),
+      size  = parseInt(getQueryParam('size'), 10) || 10,
+      from  = parseInt(getQueryParam('from'), 10) || 0;
+
+  $('.search-string').text( query );
+  $('.site-search-text').attr('value', query );
+
+  if ( ! query ) {
     $('.search-no-results').show();
     $('.loading').remove();
     $('.search-results-count').append( 0 + ' search results');
-
+    return;
   }
+
+  $('input.q').val( query );
+  $('.search-result-list').html(
+    '<div class="loading">' +
+      '<span class="glyphicon glyphicon-refresh"></span> Loading' +
+    '</div>'
+  );
+
+  $.ajax({
+    url: apiSearchUrl( query, eiti.beckleyApiKey, {
+      size: size,
+      from: from
+    }),
+    cache: false,
+    dataType: 'json'
+  })
+    .done(function(json) {
+      $('.loading').remove();
+      $('.search-results-count').append( json.hits.total + ' search results');
+
+      if (json.hits.total == 0){
+        $('.search-no-results').show();
+      }
+      else{
+        $('.search-no-results').remove();
+      }
+
+      $.each(json.hits.hits, function(i, hit){
+        var tags = (hit._source.tag || [])
+          .map(function(tag) {
+            return (
+              '<span class="search-result-list-tag">&nbsp;' +
+                '<a href="' + localSearchUrl(tag, {size: size, from: from}) + '" title="Search for ' + tag + '">' +
+                  tag +
+                '</a>&nbsp;' +
+              '</span>'
+            );
+          })
+          .join('');
+
+        var contentType = getContentTypeGlyph(hit._source.content_type);
+
+        $('.search-results-container').append(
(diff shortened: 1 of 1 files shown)


js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
   //
-  function GetURLParameter(sParam) {
-    var sPageURL = window.location.search.substring(1);
-    var sURLVariables = sPageURL.split('&');
-    for (var i = 0; i < sURLVariables.length; i++) {
-      var sParameterName = sURLVariables[i].split('=');
-      if (sParameterName[0] == sParam) {
-        return sParameterName[1];
-      }
-    }
```
dmsnell: The reason I rewrote this function, even though it worked fine, was because it was harder for me to see what it was actually doing compared to the one I wrote. Granted, some of this is personal style; I am interested in learning how y'all feel about the kind of code I replaced it with.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+      .slice(1) // drop the leading '?'
+      .split('&') // split into name=value pairs
+      .map(function(pair) { return pair.split('=') }) // split pairs into [name, value]
+      .filter(function(pair) { return pair[0] === param }) // return matching pairs
+      .map(function(pair) { return pair[1] }) // but only return the value
+      .concat( '' ) // add a default in case didn't find any
+      .shift() // return first match or empty string if none available
+      .replace(/[+]/g, '%20') // untranslate `+` into `%20` (see above)
+    );
   }
```
dmsnell: This new version always returns strings (instead of `undefined`/`null`) and also _should_ properly decode/unescape the query parameters, meaning anything that calls this function should be able to rely on it producing usable data that needs no further processing.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
 
-  var query = GetURLParameter('q') || '',
-    apiKEY = eiti.beckleyApiKey;
-
-  $('.search-string').append(query.replace(/%20/g,' '));
-  $('.site-search-text').attr('value',query.replace(/%20/g,' '));
-  if (query) {
```
dmsnell: This line right here is why it appears like I changed the whole file. Instead of wrapping everything inside a conditional, I checked for the non-existence of the query and early-aborted, leaving everything else un-nested.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+    var contentTypes = {
+      'text/html':       '<span class="glyphicon glyphicon-link"></span> Website',
+      'application/pdf': '<span class="glyphicon glyphicon-file"></span> PDF',
+      'default':         '<span class="glyphicon glyphicon-file"></span> ' + contentType
+    };
+
+    return contentTypes[contentType]
+      ? contentTypes[contentType]
+      : contentTypes['default']
   }
```
dmsnell: Not sure why this exists. It wasn't being used before but I went ahead and refactored it into a function just in case there were plans to start adding the glyphs in the future. If the glyph isn't going to be added, this can be zapped entirely.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+  $('.search-string').text( query );
+  $('.site-search-text').attr('value', query );
+
+  if ( ! query ) {
     $('.search-no-results').show();
     $('.loading').remove();
     $('.search-results-count').append( 0 + ' search results');
-
+    return;
   }
```
dmsnell: ^^^ this is the aforementioned early-abort

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+          .map(function(tag) {
+            return (
+              '<span class="search-result-list-tag">&nbsp;' +
+                '<a href="' + localSearchUrl(tag, {size: size, from: from}) + '" title="Search for ' + tag + '">' +
+                  tag +
+                '</a>&nbsp;' +
+              '</span>'
+            );
+          })
+          .join('');
```
dmsnell: When I work with arrays of things, I tend to prefer immutable maps over the arrays instead of string-building or array-building loops. Here we're actually preserving the map from tag to tagHTML on each item until the very end with the `join( '' )`.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+
+        var contentType = getContentTypeGlyph(hit._source.content_type);
```
dmsnell: This is the unused counterpart to the unused function above. I did not remove it merely to ask the question of whether it should be removed.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+
+        $('.search-results-container').append(
+          '<article class="search-result-list">' +
+            '<h1><a href="' + hit._source.url + '" target="_blank">' +
+              hit._source.title +
+            '</a></h1>' +
+            '<p>' + hit._source.description + '</p>' +
+            '<p>Tagged&nbsp;/' + tags + '</p>' +
+          '</article>'
```
dmsnell: While some may understandably prefer the `+` at the beginning, I find that when marking up HTML strings while preserving indentation that leaving the `+` at the end is visually clearer.

js/components/search.js
```diff
@@ -5,85 +5,189 @@ $( document ).ready(function() {
...
+      }
+
+      $.each(json.hits.hits, function(i, hit){
+        var tags = (hit._source.tag || [])
+          .map(function(tag) {
+            return (
+              '<span class="search-result-list-tag">&nbsp;' +
+                '<a href="' + localSearchUrl(tag, {size: size, from: from}) + '" title="Search for ' + tag + '">' +
+                  tag +
+                '</a>&nbsp;' +
```
dmsnell: What's up with these `&nbsp;`s? I left it in there, but I'm not sure why it's there.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+      var paging = {
+        needsPrevLink: from > 0,
+        needsNextLink: (from + size) < json.hits.total,
+        urlPrev: localSearchUrl(query, { size: size, from: Math.max(0, from - size - 1) }),
+        urlNext: localSearchUrl(query, { size: size, from: from + size + 1 }),
+        titlePrev: 'Previous page of search results',
+        titleNext: 'Next page of search results'
+      };
+
+      $('.search-results-container .search-results-container').append(
```
dmsnell: I'm not sure how this isn't affecting the tag list above, but there are two containers with this same class. To prevent adding links twice, I appended to the innermost of the two.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+
+  $('input.q').val( query );
+  $('.search-result-list').html(
+    '<div class="loading">' +
+      '<span class="glyphicon glyphicon-refresh"></span> Loading' +
+    '</div>'
+  );
+
+  $.ajax({
+    url: apiSearchUrl( query, eiti.beckleyApiKey, {
```
dmsnell: we didn't appear to use the API key anywhere but here, and I don't like leaving around variables for the sake of having variables, so I inlined this here. I did this a few other places too

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+    };
+
+    return contentTypes[contentType]
+      ? contentTypes[contentType]
+      : contentTypes['default']
+  }
+
+  var query = getQueryParam('q'),
+      size  = parseInt(getQueryParam('size'), 10) || 10,
+      from  = parseInt(getQueryParam('from'), 10) || 0;
```
dmsnell: yep. this is inefficient calling this for each parameter (vs caching or grabbing them all at once). however…
 - the query string should be small enough that it doesn't matter
 - optimization can be easily added; I wrote this for clarity first

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+    dataType: 'json'
+  })
+    .done(function(json) {
+      $('.loading').remove();
+      $('.search-results-count').append( json.hits.total + ' search results');
+
+      if (json.hits.total == 0){
+        $('.search-no-results').show();
+      }
+      else{
```
dmsnell: though I did not change this here, I'm not a fan of splitting the `else` onto a new line. is this part of the project/18F style?

dmsnell:
ES2015? Is this a thing here? When I started working on this PR, I was using some fat-arrow functions but then realized by inspecting the JS output that those weren't being transformed into ES5 functions nor were other things such as template literals.

My reason for asking is that if the newer features of the language are permitted here, then big chunks of this PR could be condensed and made much clearer. This is from a combination of things like the fat-arrow syntax, list destructuring, and template literals.

Thanks!

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+      'q=' + q + '&size=' + size + '&from=' + from + '&api_key=' + apiKey
+    );
+  }
+
+  function localSearchUrl( query, options ) {
+    var q = encodeURIComponent( query );
+    var size = (options && options.size) || 10;
+    var from = (options && options.from) || 0;
+
+    return '/search-results/?q=' + q + '&size=' + size + '&from=' + from;
```
dmsnell: The original code had `../search-results/` and I wondered why the `../` was there instead of using an absolute reference. Obviously, I took it out but could happily put it back.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+      ? contentTypes[contentType]
+      : contentTypes['default']
+  }
+
+  var query = getQueryParam('q'),
+      size  = parseInt(getQueryParam('size'), 10) || 10,
+      from  = parseInt(getQueryParam('from'), 10) || 0;
+
+  $('.search-string').text( query );
+  $('.site-search-text').attr('value', query );
```
dmsnell: What's the rule on spaces around parentheses? I'm a huge fan of including them, but I see a lack of them here. Of course, I'm also usually assuming that they will be stripped in the minified build too, so I don't care about code size. Are the extra spaces adding a significant amount of space?

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+    return contentTypes[contentType]
+      ? contentTypes[contentType]
+      : contentTypes['default']
+  }
+
+  var query = getQueryParam('q'),
+      size  = parseInt(getQueryParam('size'), 10) || 10,
+      from  = parseInt(getQueryParam('from'), 10) || 0;
+
+  $('.search-string').text( query );
```
dmsnell: The change to `text()` here was deliberate in order to prevent injecting HTML from the `GET` string or from the search box. The difference in how they render is as follows: &lt;s&gt;text&lt;/s&gt; verses <s>text</s>

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+  if ( ! query ) {
     $('.search-no-results').show();
     $('.loading').remove();
     $('.search-results-count').append( 0 + ' search results');
-
+    return;
   }
+
+  $('input.q').val( query );
+  $('.search-result-list').html(
```
dmsnell: In most places I have seen `append()` and I'm not sure if that is habit or preferred style. When I come across a snippet like this and I realize that whatever is being inserted should cover the entire content of the matched tag, I would lean towards using `html()` to _set_ the whole darn thing instead of taking a chance that there could be something in there already and merely `append()`ing onto it.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+      });
+
+      var paging = {
+        needsPrevLink: from > 0,
+        needsNextLink: (from + size) < json.hits.total,
+        urlPrev: localSearchUrl(query, { size: size, from: Math.max(0, from - size - 1) }),
+        urlNext: localSearchUrl(query, { size: size, from: Math.min(json.hits.total - 1, from + size + 1) }),
+        titlePrev: 'Previous page of search results',
+        titleNext: 'Next page of search results'
+      };
```
dmsnell: Could have used different data structures here. This is primarily here for namespacing within the current function and to allow easy programmatic access (see below inside the map). Honestly some of this paging stuff should probably be moved into its own function anyway.

I thought about something like this too…
```js
var paging = {
	next: {
		needsLink: from > 0,
		url: …
		title: 'Next page…',
		label: 'Next'
	},
	prev: {
		…
	}
};
```

In the end, I chose what I did since it was more terse in the map below and also because I did not anticipate needing more than `prev`/`next`. If we were to add page crumbs then I would probably want to reconsider this whole structure.

js/components/search.js
```diff
@@ -5,85 +5,165 @@ $( document ).ready(function() {
...
+    var contentTypes = {
+      'text/html':       '<span class="glyphicon glyphicon-link"></span> Website',
+      'application/pdf': '<span class="glyphicon glyphicon-file"></span> PDF',
+      'default':         '<span class="glyphicon glyphicon-file"></span> ' + contentType
+    };
+
+    return contentTypes[contentType]
+      ? contentTypes[contentType]
+      : contentTypes['default']
+  }
```
dmsnell: Not sure why this exists. It wasn't being used before but I went ahead and refactored it into a function just in case there were plans to start adding the glyphs in the future. If the glyph isn't going to be added, this can be zapped entirely.

js/components/search.js
```diff
@@ -5,85 +5,166 @@ $( document ).ready(function() {
...
+        var tags = (hit._source.tag || [])
+          .map(function(tag) {
+            return (
+              '<span class="search-result-list-tag">&nbsp;' +
+                '<a href="' + localSearchUrl(tag, {size: size, from: from}) + '" title="Search for ' + tag + '">' +
+                  tag +
+                '</a>&nbsp;' +
+              '</span>'
+            );
+          });
```
dmsnell: When I work with arrays of things, I tend to prefer immutable maps over the arrays instead of string-building or array-building loops. Here we're actually preserving the map from tag to tagHTML on each item until the very end with the join( '' ).

in this case, we have a single constant binding between the source list and the output instead of a multi-step process of altering a given variable.

js/components/search.js
```diff
@@ -5,85 +5,166 @@ $( document ).ready(function() {
...
+
+        var contentType = getContentTypeGlyph(hit._source.content_type);
```
dmsnell: This is the unused counterpart to the unused function above. I did not remove it merely to ask the question of whether it should be removed.

js/components/search.js
```diff
@@ -5,85 +5,166 @@ $( document ).ready(function() {
...
+
+        return (
+          '<article class="search-result-list">' +
+            '<h1><a href="' + hit._source.url + '" target="_blank">' +
+              hit._source.title +
+            '</a></h1>' +
+            '<p>' + hit._source.description + '</p>' +
+            '<p>Tagged&nbsp;/' + tags.join('') + '</p>' +
+          '</article>'
```
dmsnell: While some may understandably prefer the + at the beginning, I find that when marking up HTML strings while preserving indentation that leaving the + at the end is visually clearer.

gemfarmer:
@dmsnell I haven't had a chance to dive into these changes, but thanks for the monster PR!

dmsnell:
Thanks @gemfarmer - one question I was wondering was whether the search has been working before at all. The API still appears to hang when I make a search live on [US EITI](https://useiti.doi.gov/search-results/?q=oil), eventually timing out. I thought that maybe this was something still in development or waiting to finish.

gemfarmer:
@dmsnell the search has been working, but it has gone down more than ideal. I thought I fixed the issue when I added an additional Beckley app instance, but maybe I have to add more disk space? Whatever the solution, the request hanging is beyond the scope of this repo.

A change that we could make on this repo is a display for when the request hangs for an extended (3s, 5s?) period of time.

dmsnell:
> A change that we could make on this repo is a display for when the request hangs for an extended (3s, 5s?) period of time.

Sounds like another simple PR. There's currently no `.fail()` handler and no `timeout` for the search API call, both of which would be trivial to add.

dmsnell:
hey @gemfarmer - is this still a thing worth considering? I was just clearing out some old PRs in my list and wondered if it needs further work or should simply be closed.

gemfarmer:
@dmsnell yes, it is! For some reason I had it in my head that this was WIP, but it looks like we just have an unrelated issue. I will review again and merge asap! Thanks reminding me :)

dmsnell:
> @dmsnell yes, it is! For some reason I had it in my head that this was WIP, but it looks like we just have an unrelated issue. I will review again and merge asap! Thanks reminding me :)

no problem - I realize that the title is will WIP, but I probably did that since I was uncertain at the time if this was mergeable to your standards. I think now I understand better the norms of your JS process and wouldn't have labeled it as such had I done it today

thanks!

gemfarmer:
Looks good 🎉 . Thanks again @dmsnell!

PR #1023, opened 2015-12-08
Add offshore regions

shawnbot:
This is meant to address #933. Includes:

- [x] create regional boundaries in offshore TopoJSON
- [x] implement region grouping for [federal revenue](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-regions/explore/federal-revenue-by-location/#region=gulf&year=2013)
- [x] implement region grouping for [federal production](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-regions/explore/federal-production/#region=gulf)
- [x] update regional data to include offshore _region names_, not area ids

PR #1690, opened 2016-08-01
Add Offshore pages, data

gemfarmer:
Fixes issue(s) #1023, #1693 

[Offshore alaska :sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore/offshore/alaska/ALA)

## Data:
- [x] changed config file to add offshore routes (e.g. `offshore/alaska`)
- [x] added federal production offshore data
- [x] added revenue offshore data

@shawnbot, for both revenue and federal production I created a single regional yml output and a directory with yml files for each area.

## Maps (from #1698 
- [x] Adds new SVG map data for offshore regions and planning areas.
- [x] Reorganizes map includes to separate out states, offshore areas, and counties so they're more easily composable.
- [x] Changes the URL format for data map hrefs from simply replacing % with the input value to a more useful some/:key/blah format that interpolates hash keys. The old % behavior still works if the "data" is a string, and if it's a hash the data's id is substituted.


gemfarmer:
@shawnbot back atcha. I think this PR is ready to ship. Do you see any adjustments to the data that need to be made before merging to dev?

meiqimichelle:
Hmmm, I'm thinking this one is primarily data and structure (ie, I don't see anything at the preview link) -- I will leave it to @shawnbot to review :)

gemfarmer:
@shawnbot this LGTM. Mergeable from your end?

gemfarmer:
Oh oops! Sounds good.

_includes/county-map.html
```diff
@@ -12,9 +12,9 @@
 {% endif %}
 
 {% if _viewbox %}
-  {% assign _viewbox_list = _viewbox | split:' ' | to_f %}
-  {% assign breakpoint_width = _viewbox_list[2] %}
-  {% assign breakpoint_height = _viewbox_list[3] | times:1.5 %}
+  {% assign _viewbox_list = _viewbox | split: ' ' %}
+  {% assign breakpoint_width = _viewbox_list[2].to_f %}
+  {% assign breakpoint_height = _viewbox_list[3].to_f | times:1.5 %}
```
shawnbot: This is really weird. @gemfarmer I wasn't having problems with this before, but for some reason one of my changes (or a bad merge, maybe?) caused the `| to_f` bit to fail. I spent some time debugging this with `puts` calls in the template filters, and discovered that the `to_f` filter was never being called, which caused the `_viewbox_list` variable to contain an array of strings, which caused Liquid to barf on comparing `breakpoint_width` with `breakpoint_height` because the former is a string but the latter is a float (after being multiplied by 1.5). This syntax works, though, and I didn't see any other references to `to_f` in our layouts or includes, so.... ¯\\\_(ツ)_/¯

shawnbot: See: #1731

gemfarmer:
LGTM ‼️

Issue #1697, opened 2016-08-01
Create offshore region and area map data


shawnbot:
This is the mapping counterpart issue to #1693. We need SVG map data for:

- [x] Offshore regions: Alaska, Atlantic, Gulf, Pacific
- [x] Offshore _planning areas_ (Western Gulf of Mexico, Beaufort Sea, etc.) within each region
- [ ] Maybe not for this sprint, but eventually: offshore _protraction areas_ (subdivisions of planning areas), found [here](https://www.data.boem.gov/homepg/data_center/mapping/geographic_mapping.asp), and which [look like this](https://www.data.boem.gov/images_opt/graphics/gis/GOMR_Protrac.png).

Then we need to make those regions visible in our national page maps. Rather than shoehorning these into the `state-map` include, I'm going to reorganize the map includes to pull from other subsets of shapes so that they're easier to pick and choose from.

shawnbot:
Done in #1698!

PR #1698, opened 2016-08-01
[WIP] Offshore maps

shawnbot:
@gemfarmer this is a request to merge my map work into your offshore branch. Addresses #1697.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-maps/states/)

- Adds new SVG map data for offshore regions and planning areas.
- Reorganizes map includes to separate out states, offshore areas, and counties so they're more easily composable.
- Changes the URL format for data map hrefs from simply replacing `%` with the input value to a more useful `some/:key/blah` format that interpolates hash keys. The old `%` behavior still works if the "data" is a string, and if it's a hash the data's `id` is substituted.


_includes/maps/state-areas.html
```diff
@@ -0,0 +1,21 @@
+    <g class="states features">
+      <use xlink:href="{{ include.svg }}#states"></use>
+    </g>
+    {% assign keys = include.value %}
+    {% for state in include.states %}
+    {% assign _state = site.states | where: 'id', state[0] | first %}
+    {% if include.href %}<a xlink:href="{{ include.href | format_url: _state }}">{% endif %}
```
gemfarmer: The link to individual states isn't working on Federalist. I think we need to add a `{{ site.baseurl }}` to the href here

gemfarmer:
@shawnbot I made one comment, other than that, looks good to merge!

_includes/maps/state-areas.html
shawnbot: Yep, makes sense. Fixing!

gemfarmer:
@shawnbot this works for my locally, but not on Federalist (see below). It looks like the paths following the domain, but prior to the intended route are being stripped so that

`https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-maps/states/` 
becomes 
`https://federalist.18f.gov/states/NV/`
instead of 
`https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-maps/states/NV/`
 
I haven't investigated, but I'm assuming this has something to do with the `format_url` filter as we weren't previously seeing this kind of routing issue.

screen shot 2016-08-02 at 10 18 59 am

gemfarmer:
Nice! Merging :)

PR #1712, opened 2016-08-06
Add minimum Ruby version requirement

dmsnell:
Previously, while attempting to build the site, I could not get the
build to complete without error. It complained of `to_h` not existing.

After tracking down this problem, I realized that my system was running
a late Ruby 2.0-release and `to_h` requires `2.1`

This patch adds a specifier in the Gemfile to require a minimum version.

cc: @shawnbot @gemfarmer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,8 @@
 # A sample Gemfile
 source "https://rubygems.org"
 
+ruby "> 2.1.0"
+
 gem 'jekyll', '3.0.5'
 gem 'scss_lint', require: false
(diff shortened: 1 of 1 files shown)


dmsnell:
right @shawnbot it confused me too, which is why it's not fixed. ruby is such a 💎

dmsnell:
@gemfarmer I think it should work this time. took out the patch version number and it appeared to work locally; also I think I needed a `~` in there

gemfarmer:
No issues over here. Merging!

Issue #1710, opened 2016-08-05
State page without Javascript


gemfarmer:
When a user doesn't have JS enabled, display content in a way that still delivers the primary focus of the page. For the state pages, this means 
- [ ] showing the informational bar chart text
- [ ] displaying the county map tables
- [ ] ensuring that all tabs work with CSS only
- [ ] hiding buttons that are only JS accessible
- [ ] hiding maps elegantly or displaying them in a way that demonstrates the need to turn on JS
- [ ] other things

coreycaitlin:
Closing in favor of #2606

Issue #1590, opened 2016-07-12
Locale currency support


gemfarmer:
Javascript supports locale currencies! Let's add them to the site so that other countries can view our project more seamlessly!

Refs:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

![pasted image at 2016_07_11 12_51 pm](https://pbs.twimg.com/media/CmoKS0JVUAA8sEI.jpg)

gemfarmer:
A corollary to this – we might also consider allowing people to convert units to something more accessible/relatable. It might be nice to allow people to convert say `short tons` to `tons`, `kilos`, `lbs`, etc.

dmsnell:
@gemfarmer it's worth noting that all this does is replace the number format and the currency symbol. It does not, for instance, convert the _value_ of the currencies. That is to say that if we give it $100 and ask for JPY it will give us 100¥ (or 百円 maybe) instead of something like 10,000¥ as is the exchange value.

We _could_ hook this up to some API to convert the values, but I imagine that most values here will be historical and the exchange rates fluctuate with time, so some decisions would need to be made on how to make that historical conversion too.

At this point, short of hard-coding the properly-converted values into a variety of currencies in the data itself, is it feasible/practical to enhance the site in this way?

meiqimichelle:
We probably don't need to do this, not a strong user need :) It is cool though! Closing for now.

Issue #1541, opened 2016-06-29
Add adjustable scale to megachart


ericronne:
Let's discuss, and then possibly test, and then maybe actually create a tool enabling users to adjust the scale of the megachart. This will enable them to visually compare differences between smaller numbers, in cases where an outlier (such as oil and gas) completely throws off the scale.

(Just making a backlog issue for this, so it doesn't get forgotten.)

ericronne:
If it's deemed useful, it might be handy on other charts, as well.

Perhaps similar to this design, which was concepted for revenues (but never implemented)...

![image](https://cloud.githubusercontent.com/assets/9259185/16459170/8e83dc86-3de7-11e6-9450-e7cff0dce98d.png)

Eric note to self: see "EITI Explore Data - Desktop v2.sketch"

@coreycaitlin @meiqimichelle @gemfarmer @shawnbot

ericronne:
First envisioned in #962

ericronne:
One simple(?) solve(?) would be to offer the option to hide the largest commodity, like so ...

![image](https://cloud.githubusercontent.com/assets/9259185/17453589/0baed48c-5b33-11e6-8bd3-1165dc68d115.png)

coreycaitlin:
Noted this idea for posterity; closing this issue — will reopen if this emerges as a user need.

Issue #991, opened 2015-12-07
Explore > Revenue by Company: show total in revenue type chart


meiqimichelle:
There should be a 'total' line in the revenue type chart. Quick link to updated description [in this comment](https://github.com/18F/doi-extractives-data/issues/991#issuecomment-234407503) ...after all the GitTracks of us not doing this 😁

gemfarmer:
A bar for "Total" needs to be added to this chart on the [explore/federal-revenue-by-company](https://useiti.doi.gov/explore/federal-revenue-by-company/2015/) page

<img width="698" alt="screen shot 2016-07-21 at 5 47 23 pm" src="https://cloud.githubusercontent.com/assets/4803473/17041362/43bb9b54-4f6b-11e6-8a8e-012c161602b2.png">

ericronne:
Moot, @meiqimichelle?

meiqimichelle:
Moot.

PR #1715, opened 2016-08-07
WIP: Reactify revenue table

dmsnell:
This is a **work-in-progress** PR exploring the concept of replacing the Revenue table with a React component hierarchy. It shouldn't be considered for merging until it's complete.

## Why do this in the first place?

When I started trying to work on #1541 I realized that there was so much complexity in understanding how the revenue tables were generated and how the JavaScript interacted with them that I could not reasonably begin to address it without spending too much time getting myself acquainted with the code.

Using React in this case is far more about building a declarative and traceable chain of rendering and behavior than it is about using React. Presumably this could be done with web components just the same, but I am fluent in React and ignorant in web components. Additionally, I personally like the programming model in React more as well `¯\_(ツ)_/¯`

## What problem is this really solving anyway?

The revenue table seems to be a big and important piece of the state pages. It also feels very difficult to jump into and modify or extend. I have hit a barrier when trying to work on #991 and #1541 because of the complexity.

Ideally, this PR would make it easier to get acquainted with the revenue table, reduce the barriers to modifying and extending it, further to make _testing_ the revenue behaviors possible.

## Other Notes

 - By creating a `localize()` wrapper component, it becomes trivial to provide utilities for converting number/string formats (such as requested in #1590).

 - Using React to render makes #1710 far more difficult because nothing renders without JavaScript enabled. It is not a deal-breaker, however, as the React-based pages can be pre rendered into static HTML and hooked up with relative ease. Granted, that PR is challenging overall for any construction since the current page is so dynamic.

 - As in #1616, this introduces quite a few new dependencies in the build toolchain. It opens up the easy opportunity to use more modern JavaScript syntax and ES6 modules throughout other existing source files.

> Note: This PR is for tracking the development during the process. Unfortunately it is going to be a huge branch because it introduces the entire React stack and new components in a way that isn't really possible to do incrementally.

## Known bug-fixes as a side-effect

 - Negative revenue values previously appeared at `$-7,139` but now appear properly formatted (thanks to `toLocaleString()`) as `-$7,139`

 - Negative revenue values now appear visually in the "filled-bar" component. Previously they rendered as a black visual glitch slightly offset from the surrounding bar and with a black outline around the whole box. Now they take on a red color and grow from the right side of the box instead of from the left.

 - The "All Commodities" row now only includes icons for the commodities with actual values for the state. Previously all icons were visible on the "All Commodities" row

 - When both the oil _and_ gas rows have zero revenue, the table now shows a single `$0.00` for the cell instead of two lines with `Gas $0` and `Oil $0`, which looked kind of odd. See the Arizona state page for an example.

 - Similarly, when only a single revenue for either oil _or_ gas is available, there is no need to show a zero value for the empty commodity. This has been slimmed to only show the commodity with revenue. See the Indiana state page for an example.

 - When the sum of the revenue is less than the maximum revenue value, the bar graphs were visually glitching. Now, the maximum revenue component is used in these cases which fixes the glitches. See the state page for North Carolina for an example.

 - For some state, the all commodities summary table was showing up even when there was only a single type of commodity. This has been fixed. See the state page for North Carolina for an example.

 - Previously, some commodities (such as gas and oil) may have shown even if there was no revenue for the commodity. The list of available commodities has been filtered to prevent this. See the state page for Florida for an example.

## Possible bug-fixes as a side-effect

 - It seems like the table doesn't render all of the data, likely due to misdirection in the revenue data structure. For example, see **Arizona** and **Geothermal** or **Potassium**. I will probably work this into the current PR because I feel like things aren't right in `dev`

 - The "All Commodities" value in the Jekyll-build is slightly smaller than in the React build. I believe that the Jekyll build is skipping some very small data sources but haven't figured out which ones yet. No differences that I have seen have been significant.

## Possible bugs

 - On the Alabama state page, there is a missing `-$615` revenue for `Hardrock Minerals` > `Hardrock Minerals`

## Todo before final ready-for-merge review

 - [ ] Incorporate new tables into **Revenue details by phase**
 - [ ] Remove existing Jekyll page rendering templates
 - [ ] Refactor oil/gas rows to eliminate separate `oil-gas-table.jsx`
 - [ ] Try and cleanup logic in `state-revenue-table.jsx`
 - [ ] Confirm possible missing commodity views bug (see above, not caused by this PR)
 - [ ] Reorganize master data list to be more conducive to the page at hand and simplify the data-mangling in this PR.

cc: @shawnbot @gemfarmer


gemfarmer:
@dmsnell: Thanks again for the contribution!

@shawnbot and reviewed this, and while it looks great, I don't think we are ready to incorporate javascript frameworks into the site at this juncture. 

We are closing the PR to clear our backlog.

dmsnell:
> Negative revenue values previously appeared at $-7,139 but now appear properly formatted (thanks to toLocaleString()) as -$7,139

This can easily be solved using a small JavaScript function but I'm not sure how easy it would be to implement in Jekyll without installing moar gems

> Negative revenue values now appear visually in the "filled-bar" component. Previously they rendered as a black visual glitch slightly offset from the surrounding bar and with a black outline around the whole box. Now they take on a red color and grow from the right side of the box instead of from the left.

Notice how in the before shot, the "progress" part of the graphic slumps below the line, as if it were a descender.

**Before**
screen shot 2016-09-02 at 2 29 39 pm

**After**
screen shot 2016-09-02 at 8 52 54 am

cc: @meiqimichelle

dmsnell:
> The "All Commodities" row now only includes icons for the commodities with actual values for the state. Previously all icons were visible on the "All Commodities" row

Here we can look at Arizona which only has a limited subset of commodity types. This isn't a bug _per-se_, but it seems like the icons should correspond to the data more. In the before shot we see all four icons but in the after only the two relevant types are listed.

> The "All Commodities" row now only includes icons for the commodities with actual values for the state. Previously all icons were visible on the "All Commodities" row

While there are still underlying oddities causing this to happen, you can see in the before shot that the geothermal header is still there with no revenue rows listed. I believe that this has to do with the way the data is passed and filtered. Regardless, that header shouldn't be there _or_ there should be rows that we're missing.

**Before**
screen shot 2016-09-02 at 2 34 04 pm

**After**
screen shot 2016-09-02 at 2 33 01 pm

cc: @meiqimichelle

dmsnell:
> When both the oil and gas rows have zero revenue, the table now shows a single $0.00 for the cell instead of two lines with Gas $0 and Oil $0, which looked kind of odd. See the Arizona state page for an example.

**Before**
screen shot 2016-09-02 at 2 36 55 pm

**After**
screen shot 2016-09-02 at 2 37 34 pm

> Similarly, when only a single revenue for either oil or gas is available, there is no need to show a zero value for the empty commodity. This has been slimmed to only show the commodity with revenue. See the Indiana state page for an example.

**Before**
screen shot 2016-09-02 at 1 46 19 pm

**After**
screen shot 2016-09-02 at 1 41 23 pm

cc: @meiqimichelle

dmsnell:
> When the sum of the revenue is less than the maximum revenue value, the bar graphs were visually glitching. Now, the maximum revenue component is used in these cases which fixes the glitches. See the state page for North Carolina for an example.

In this case, we had very few resources and the max revenue value was only $474. In most cases, the sum of the individual revenues will be larger than the single largest revenue value. Here, however, the sum of all the revenue values (due to the -$316 and lack of other revenue sources) means that the sum is less than the largest single source and therefore trying to scale off of that was failing. Glitch. By scaling off of the largest single value we can fix it.

**Before**
screen shot 2016-09-02 at 2 42 31 pm

**After**
screen shot 2016-09-02 at 2 43 10 pm

cc: @meiqimichelle

dmsnell:
> Previously, some commodities (such as gas and oil) may have shown even if there was no revenue for the commodity. The list of available commodities has been filtered to prevent this. See the state page for Florida for an example.

**Before**
screen shot 2016-09-02 at 2 49 44 pm

**After**
screen shot 2016-09-02 at 2 50 11 pm

cc: @meiqimichelle

Issue #1723, opened 2016-08-09
national_revenue_by_type data bug


gemfarmer:
Each commodity in `national_revenue_by_type.yml` is missing the `All` key. 

This is making it so each commodity doesn't have an extent, which is preventing the bars from being drawn.

Update the data transforms to include an all key, probably via a rollup.

PR #1724, opened 2016-08-09
Fixes National revenue by type data bug

gemfarmer:
Fixes issue(s) #1723 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/national-revenue-by-type/states/)

- updated Makefile to include commodity-level rollups for the national page
- updated the data in `national_revenue_by_type.yml`

/cc @coreycaitlin @shawnbot


gemfarmer:
@shawnbot this shouldn't fix the table yet, because it doesn't have the template changes. When it is combined with the template change that @coreycaitlin made, the megachart issue should disappear.

Is that what you are getting at?

gemfarmer:
If that is the case, I'm not sure it needs to be blocked, because the PR is just to add the data.

Issue #1637, opened 2016-07-21
How it works > offshore oil & gas "did you know" bug


gemfarmer:
## Expected Behavior
- when you visit the page, the "did you know" section should be small with a button that says "more". - Clicking on the "more" button should toggle the remaining content

## Actual Behavior
- when you visit the page, the "did you know" section is fully expanded
- Clicking on the "more" button toggles the content

## Steps to reproduce the behavior

PR #1726, opened 2016-08-09
How it works bug

gemfarmer:
Fixes issue(s) #1637

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/how-it-works-bug/how-it-works/offshore-oil-gas/)

There were two changes that were addressed in this pull request:
- accordion.js wasn't properly handling event bubbling, so clicking on the `+more`/`- less` buttons was targeting the span and erroring
- on how-it-works/offshore-oil-gas, the accordion content wasn't nested properly, so __nothing__ was being toggled. 

/cc @meiqimichelle


coreycaitlin:
Yay! More makes it show more, and less makes it show less! 🎉

Issue #1731, opened 2016-08-10
Federalist builds are failing


shawnbot:
Something changed on dev about two hours ago that is causing all builds on dev and any branch that has dev merged since then to fail. I see the following in the [Federalist logs](https://federalist.18f.gov/#site/73/logs):

```
Configuration file: ./_config.yml
Source: . Destination: ./_site
Incremental build: disabled. Enable with --incremental
Generating...
Liquid Exception: Liquid error (line 24): comparison of String with 172.1954818519226 failed in _includes/location/section-ownership.html, included in _layouts/state-page.html
jekyll 3.0.5 | Error: Liquid error (line 24): comparison of String with 172.1954818519226 failed
```

@gemfarmer It looks like it was a merge to dev "about 2 hours" ago, which, judging from the commits, appears to be #1703. We need to figure this out ASAP so that we can get Federalist previews working again.

gemfarmer:
Ugh. So took a brief look tonight and I haven't fixed anything, but I have discovered a few things:

1. The following snippet from `county-map.html` is causing issues. The liquid error disappears when I explicitly set the breakpoint_width and breakpoint_height to numbers. This is likely because the chained `split` and `to_f` filters aren't doing what they are supposed to
 
```
{% assign _viewbox_list = _viewbox | split:' ' | to_f %}
  {% assign breakpoint_width = _viewbox_list[2] %}
  {% assign breakpoint_height = _viewbox_list[3] | times:1.5 %}
  {% capture is_wide %}
    {% if breakpoint_width > breakpoint_height %}wide{% endif %}
  {% endcapture %}
```

2. The paths to the state-specific svgs aren't correct. I am seeing the following:

```
http://localhost:4000/maps/states//states/UT.svg#counties
```

It has this problem even if I change the href in county-map.html :scratch-head:

3. The bar charts are broken


These issues seem unrelated, but they likely are. How? rrr

gemfarmer:
@shawnbot good news/bad news.

Good news:  I fixed everything. 
Bad news: I had to remove `module_function :yada` from all of the methods to do so. So now our tests are broken :(

gemfarmer:
I pushed it up to the `tests-hotfix` branch if you feel like taking a look

shawnbot:
This is fixed in dev, but dev needs to be merged to any other branches before they'll build.

PR #1732, opened 2016-08-10
Site fixed, tests broken

gemfarmer:
Fixes issue(s) #1731 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/tess-hotfix/)

@shawnbot good news/bad news.

Good news:  I fixed everything. 
Bad news: I had to remove `module_function :yada` from all of the methods to do so. So now our tests are broken :(

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -70,9 +70,9 @@ collections:
     output: true
     permalink: /how-it-works/:path/
   offshore_areas:
-    output: true
+    output: false
   offshore_regions:
-    output: true
+    output: false
     permalink: /offshore/:path/
   states:
     output: true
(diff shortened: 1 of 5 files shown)


gemfarmer:
@shawnbot I decided the simplest fix for now is to disable automatic tests via CircleCI. 

This doesn't really solve our problem, but I couldn't figure out a way to run our tests AND have the filters work, unfortunately. This might be a case for adding minitest or rspec. Since this isn't a super high priority for us, maybe one of the folks from the ruby group would be willing to lend their help with the config??

So, until the testing suite is updated or changed, we will need to uncomment the `module_function` declaration at the top of both `eiti` prefixed filter files to run the tests. Since we aren't likely to be fiddling around with existing filters much, this seems sufficient for the time being.

Issue #1695, opened 2016-08-01
State pages: connecting a county in drop-down list to map


meiqimichelle:
Some users know a county name, some know its location. We should have a design for connecting a name on a county expanda-list to the map, and maybe the converse (perhaps a tool-tip with revenue info on click or hover?)

![screenshot 2016-08-01 10 37 14](https://cloud.githubusercontent.com/assets/4827522/17303072/ee7f4be2-57d3-11e6-8f68-e9ce0b3d39aa.png)

gemfarmer:
_ideas_
**Map --> Chart**: click on map and the table opens with that county's row bolded?
**Chart --> Map**: We probably don't want to change the county's fill color (because then it wouldn't correspond to the legend), but maybe we could change it's stroke color and have a tooltip with name and corresponding data as @meiqimichelle suggested.

ericronne:
+1 re outlining vs altering the county's color on hover, and to your suggestion for behavior. 

### That's what i've done in the [invision prototype here, which y'all should test and comment upon!] (https://invis.io/DP874D145#/179355225_Production_-_Default). ###

 I think a traditional tool tip might be the best route to take for the hover, because it works well for both vertical and horizontal maps.

My topline thoughts re behavior:

 - Hover outlines county with a **two-pixel rule** `#587F4C` bar-chart green **at 50% opacity** and shows tool tip with sentence-style county data:
![image](https://cloud.githubusercontent.com/assets/9259185/17449910/58f07870-5b12-11e6-9b02-48f71069b1ec.png)

 - County click darkens its green outline to full opacity, opens the table (assuming it isn't already open), expands that county's row, adds the sentence-style data, and fills the background of that row with `#F7F7F7`, our $light-gray (an accordion, basically)
![image](https://cloud.githubusercontent.com/assets/9259185/17450067/15805046-5b13-11e6-8168-cfd03e0df680.png)

Again, please leave comments in invision, and we can chat, as well.
Thx!

ericronne:
I removed the green and boldfacing from the table design; i don't think we need it there. On the bars, the green is meant to reinforce the connection between the figure and the selected (green) bar, since they're not adjacent. But here, everything is boxed together ...

![image](https://cloud.githubusercontent.com/assets/9259185/17486262/0b93b634-5d45-11e6-91bb-2249989e6737.png)

coreycaitlin:
I like this a lot, @ericronne — seems super clear to me.

[...]

Issue #1673, opened 2016-07-27
State jobs data is not accurate


gemfarmer:
## Expected Behavior
The bar chart of state wage and salary data should be a sum of all the counties.

## Actual Behavior
Instead appears to be using the first instance of the county data. Fortunately (or unfortunately), this appears to be caused by a problem with the data. `state_jobs.yml` is not a sum :(
<img width="748" alt="screen shot 2016-07-27 at 4 35 53 pm" src="https://cloud.githubusercontent.com/assets/4803473/17193356/45a79b28-5418-11e6-8fd4-ff8dc8e69425.png">
## Steps to reproduce the behavior
Update the Makefile to sum the data properly and rebuild the yml file.

PR #1744, opened 2016-08-15
Interactive chart map

gemfarmer:
Fixes issue(s) #1695

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/interactive-chart-map/states/UT)

- Interactively connects the county-level chart  with the county-level map and year selectors for the following sets of data:
  - [x] federal production
  - [x] revenue
  - [x] jobs

NOTES: 
- there is a visible bug in the jobs data, but it is related to #1673, not the result of this issue
- The county outlines don't always look great. This is caused by the SVGs overlapping and will be pushed into a spillover issue.

/cc @ericronne @meiqimichelle

### DIFF
--- a/_includes/county-map.html
+++ b/_includes/county-map.html
@@ -35,6 +35,9 @@
     <g class="counties features">
       <use xlink:href="{{ state_svg }}#counties"></use>
     </g>
+    <g class="counties mesh">
+      <use xlink:href="{{ state_svg }}#counties-mesh"></use>
+    </g>
 
     {% if include.counties %}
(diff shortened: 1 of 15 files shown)


gemfarmer:
@meiqimichelle @ericronne ready for review! 🎉

ericronne:
Coolio!

Do you want my trademark annoying nitpicks now, or should we pull it and do a sep issue?

gemfarmer:
WOO!

PR #1758, opened 2016-08-18
No data/withheld UI

gemfarmer:
Fixes issue(s) #1704 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/no-data-ui/)

DIFFs: https://gist.github.com/gemfarmer/20a064bb8921fb3814f0b5ad93ea116a

Bar chart
---

For a given bar, bar chat caption: 
- [x] there is data that is the sum of all the counties for that year (no withheld flag)
- [x] there is no data for any counties (no data flag)
- [x] there is no data for any counties EXCEPT counties that have withheld data (withheld flag)

Map
---
- [x] if there is data with values or withheld data, show county 

Map legend
---
- [x] if there is only withheld data, show withheld figcaption
- [x] if there is no data, show no data figcaption
- [x] if there is data, show data figcaption

Table toggle, county tables
---
- [x] show if there is data or withheld data in any county for a given year
- [x] should disappear if there is no data

County Table

- [x] never show counties that don't have data
- [x] show withheld counties and data counties
- [x] withheld underlying caption needs to be specific to withheld

TODO still. Break off into new branches??
---
- [ ] Decide how we are going to filter bar charts. Is it only when there is no data for every year? Or is it when there is no data or withheld data? Do we filter if every year is withheld?
- [ ] Select box show "no data" only when there is no data
- [x] 0 data x axis should be dark, not gray

/cc @ericronne


gemfarmer:
Woo! 🎉

Issue #1740, opened 2016-08-15
Offshore data regions/areas


meiqimichelle:
Dump existing data onto a layout that uses our state profiles pages as a start.

gemfarmer:
@coreycaitlin @shawnbot @meiqimichelle to clarify, we left the conversation that we had last week with the resolution that we would (for now):
- treat offshore regions as "states" and offshore areas as "counties"
- add content on the offshore-region level
- remove offshore area pages
- _not_ cover protraction data?
- _not_ have a offshore landing page

PR #1769, opened 2016-08-22
Offshore prototype

gemfarmer:
Fixes issue(s) #1740 

[Central Gulf of Mexico :sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-prototype/offshore/gulf/CGM/)

[Gulf of Mexico :sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-prototype/offshore/gulf/)

- Added stubs for all of the offshore regions (gulf, atlantic, pacific, alaska)
- Added federal production and federal revenue data for offshore
- Used templates from the national page without copy (outside of headings)

Ready for review! 🎉 

/cc @meiqimichelle @coreycaitlin

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -73,7 +73,7 @@ collections:
     output: true
     # permalinks are defined in _offshore_areas/???.md
   offshore_regions:
-    output: false
+    output: true
     permalink: /offshore/:path/
   states:
     output: true
@@ -91,6 +91,12 @@ defaults:
       type: states
     values:
       layout: state-page
+
+  - scope:
+      path: ''
+      type: offshore_regions
+    values:
+      layout: offshore-region
 
   - scope:
       path: ''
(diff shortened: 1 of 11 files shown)


meiqimichelle:
@gemfarmer @coreycaitlin I can get to all the like 'central gulf of mexico' and similar levels, but larger areas don't seem to be working for me. However! Maybe this is what we want in the end? It seems pretty intuitive to click on the map and get to a thing this way.

gemfarmer:
@meiqimichelle I didn't add content to the larger areas (aka regions). I just spoke with @coreycaitlin and was reminded that we also have data for all lands & waters production, but only on the "regional" level. This made us think that instead of showing data at the area level, we should show all data only at the regional level, and display area-level data like we do for counties in the state pages.

meiqimichelle:
@gemfarmer Given the conversation here, should I merge this for now, or were you going to adjust to the latest thinking on offshore?

PR #1779, opened 2016-08-23
state jobs data, data tests

gemfarmer:
Fixes issue(s) #1673 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/jobs-data/)

- [x] added salary and wage jobs data for the state level
- [x] fixed an issue that was hiding the county table dropdown if there was withheld data in the first item of the county data array (slightly unrelated, but noticed while I was looking at the jobs dropdown)
- [x] adding data tests for self employment. This is a tricky one because the data goes through a few transformations. For now, I am going to focus on checking that the data remains consistent in its transformation from `joined.tsv` files to the yml.

/cc @shawnbot


test/data/state-self-employment.js
```diff
@@ -0,0 +1,176 @@
...
+/* jshint -W110 */
+var tito = require('tito');
+var fs = require('fs');
+var path = require('path');
+var yaml = require('js-yaml');
+var assert = require('assert');
+var parse = require('csv-parse');
+var async = require('async')
+
+var _ = require('lodash');
```
shawnbot: It doesn't look like this is being used anywhere. Can we remove this dependency?

test/data/state-self-employment.js
```diff
@@ -0,0 +1,250 @@
...
+            }
+
+          done();
+        });
+      });
+
+    }
+  });
+
+    // it("doesn't contain values that aren't in the pivot table", function(done) {
```
shawnbot: If you want to mark a test as pending you can use `xit()` instead of `it()`. Those tests won't be run, but will show up in the output as pending.

test/data/state-self-employment.js
gemfarmer: @shawnbot I used it in a later commit, so unless you think we should avoid having it as a dev dependency, is it ok if we keep it in?

gemfarmer:
@shawnbot I think the tests are in a good place for now.

tests cover:

- state rollups
- county data
- checks if all the county data present in the site data is in the source files

* note: the only hiccup here is that there were some rows in the source data that have fips, but no county name. These rows have been excluded in the data. Should we try to include them?

gemfarmer:
Close in favor of #1789

PR #1782, opened 2016-08-23
More national profile page clean up, and fancy new sticky subnav

meiqimichelle:
[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/national-page-ribbon/states)

- New sticky nav that isn't broken
- Tighter viewbox (less padding) around national map
- Properly organized right column on the top of the national page (though content is not final)

/cc @gemfarmer


_sass/components/_ribbon.scss
```diff
@@ -103,6 +107,10 @@
   }
 }
 
+.ribbon-card-top {
```
gemfarmer: The extra specificity is necessary in this case

_sass/components/_sticky.scss
```diff
@@ -0,0 +1,19 @@
+.sticky {
+  position: -webkit-sticky;
```
gemfarmer: need it for safari

Issue #1612, opened 2016-07-15
State page overview design: Make more interesting


ericronne:
Currently the state-page "overview" (intro) sections are a little drab, consisting simply of text in a gray box...
![image](https://cloud.githubusercontent.com/assets/9259185/16887613/c3bca552-4aa0-11e6-98cc-5fdae7c2acfc.png)

Adding a very-simple map is one way to easily add interest and visually anchor the page. Perhaps it doesn't need to be anything more than a map of the state (at least for now). This will at least serve to instantly reinforce to users that they've arrived at the right place (since they may well have clicked on that state's shape on the national map in order to reach this page).
![image](https://cloud.githubusercontent.com/assets/9259185/16887578/98edc61c-4aa0-11e6-8125-40a715194818.png)

ericronne:
@meiqimichelle @coreycaitlin @gemfarmer 👍 ?

coreycaitlin:
👍

meiqimichelle:
@ericronne Any thoughts on what this might look like for horizontal states?

coreycaitlin:
Here's our new/improved/revised intro content for a few nice, meaty states:

> Natural resource extraction varies widely from state to state. In Texas, extractive industries accounted for 13.6% of gross domestic product (GDP) in 2013, and jobs in the extractive industries made up 3.8% of statewide employment.
> 
> Texas leads the nation in production of:
> 
> * Oil: 31% of U.S. production
> * Natural Gas: 28% of U.S. production
> * Wind: 21% of U.S. production
> 
> Natural resource ownership in the U.S. is closely tied to land ownership. Land can be owned by citizens, corporations, Indian tribes or individuals, or governments (for instance, federal, state, or local governments). Many USEITI datasets only cover natural resource extraction on federal land, which represents 1.8% of all land in Texas.
> 
> Texas also borders an offshore area with significant natural resource extraction, which may contribute to the state's economy. For production and revenue data about offshore extraction near Texas, see the [Western Gulf of Mexico]().
> 
> For a detailed view of how the extractive industries affect communities in Texas, read more about [Tarrant and Johnson Counties]().
> 
> -------
> 
> Natural resource extraction varies widely from state to state. In Montana, extractive industries accounted for 5.7% of gross domestic product (GDP) in 2013, and jobs in the extractive industries made up 12.4% of statewide employment.
> 
[...]

PR #1785, opened 2016-08-24
Makes top of state profile pages more interesting

meiqimichelle:
Addresses #1612

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/state-page-top/states/MT/)

- Moves land ownership section into introduction in heading
- Sets unique viewboxes for every state so that little ones aren't HUGE and vertically-oriented states are nicely centered and other goodness

/cc @gemfarmer @ericronne


gemfarmer:
@meiqimichelle LGTM!

PR #1789, opened 2016-08-26
Update state jobs data, jobs data tests

gemfarmer:
Fixes issue(s) #1673 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/jobs-data/)

- [x] added salary and wage jobs data for the state level
- [x] fixed an issue that was hiding the county table dropdown if there was withheld data in the first item of the county data array (slightly unrelated, but noticed while I was looking at the jobs dropdown)
- [x] adding data tests for self employment. This is a tricky one because the data goes through a few transformations. For now, I am going to focus on checking that the data remains consistent in its transformation from `joined.tsv` files to the yml.

@shawnbot I think the tests are in a good place for now.

tests cover:

- state rollups
- county data
- checks if all the county data present in the site data is in the source files

* note: the only hiccup here is that there were some rows in the source data that have fips, but no county name. These rows have been excluded in the data. Should we try to include them?


gemfarmer:
Great! Thanks @shawnbot! Mergin..

PR #1810, opened 2016-08-30
Clickable states

### LINKED ISSUE #1778: Bug: Can't click on Iowa or Wisconsin in the national map
I can't seem to get to Iowa or Wisconsin from the national map on the `/states/` page, no matter how carefully I hover.

shawnbot:
Fixes #1778.

## :sunglasses: Previews
* [explore page](https://federalist.18f.gov/preview/18F/doi-extractives-data/clickable-states/explore/)
* [case studies](https://federalist.18f.gov/preview/18F/doi-extractives-data/clickable-states/case-studies/)

- Moves the `/states/` directory to `/explore/`.
- Changes the logic for which areas to render in the `state-map.html` include:
  * If `include.states == false`, then states are not rendered. Otherwise, `include.states` is assumed to be a hash with a value for each state id, and all states (the `site.states` collection) are rendered, with a `data-value` attribute for each with a corresponding value. The `include.value` expression is used to access nested values, as before.
  * The same logic holds for offshore areas, but this will be scrapped in a future update when we change these to offshore _regions_.
- Updates the parameters passed to `state-map.html` on the explore landing page to exclude `states` and `value`, and only include `href=':url'` so that each one is wrapped in a link to the respective state page.
- Updates the parameters passed to `state-map.html` on the case studies landing page to exclude offshore areas and not wrap states in links.
- Removes the `/maps/` page, which no longer works.

/cc @gemfarmer @meiqimichelle


gemfarmer:
@shawnbot Much better! LGTM

Issue #1773, opened 2016-08-22
Withheld data spillover


gemfarmer:
Spillover issue from #1758 

- [ ] Decide how we are going to filter bar charts. Is it only when there is no data for every year? Or is it when there is no data or withheld data? Do we filter if every year is withheld?
- [ ] Select box show "no data" only when there is no data

New discovery:
- [ ] there is withheld data showing up in maps and charts outside of federal production. This is likely caused by `eiti-data-map.js` mistakenly defaulting values to `Withheld` instead of `no-data`, or in other words `null` instead of `undefined`. This will likely clear up many of the quirks I was noticing, primarily the select box no data bug mentioned above. Also, the `county-map.html` template _could_ be the culprit, setting `undefined` values to `null`

@coreycaitlin @ericronne we should set a time to decide which bar charts are filtered!

ericronne:
I feel like we shouldn't ignore when data has been withheld, though it would be awkward to show an entire (blank) chart if that's all of the info we have. Sure, let's chat!

gemfarmer:
@ericronne agreed. And that is why I left them unfiltered for now. It seemed a little strange to remove them, because `Withheld` data is interesting/important, but it also looks crummy at the moment.

ericronne:
If it doesn't mess with the sorting logic, what if we ran such charts at the end, like Oil & Gas here?...

![image](https://cloud.githubusercontent.com/assets/9259185/17902840/934e4f90-692e-11e6-80c4-4cd48981fb21.png)

gemfarmer:
Hmm yeah, I like that. We could also have those hidden in some kind of accordion?

ericronne:
Do we know about how many we're talking about?

ericronne:
Created a new issue to deal with no-data charts: #1851

Issue #1761, opened 2016-08-18
Bug: Revenue in CA


gemfarmer:
Bug caused by a change in the no data/withheld PR #1758 

Clicking on bar charts with data hides the chart toggle when there is data. When there is no data it shows the toggle. 

There was likely some logic that was reversed

PR #1861, opened 2016-09-07
Withheld bugs

gemfarmer:
Fixes issue(s) #1773, #1761

[:sunglasses: PREVIEW UT 🗻 ](https://federalist.18f.gov/preview/18F/doi-extractives-data/withheld-bug/explore/UT)

- updated filter logic in `eiti-data-map.js` to properly evaluate whether or not the map has values, which allows the toggle to show/hide as expected.
- Discovered that the `Withheld` values that were showing up in the county-level charts, was the product of `bar-chart-table.js` not distinguishing between truly withheld values and instances in which a county has values one year, but not another. I updated the logic to reflect this distinction.
- there should no longer be any data that is flagged as `Withheld` outside of federal production.

/cc @shawnbot @meiqimichelle

### DIFF
--- a/js/components/bar-chart-table.js
+++ b/js/components/bar-chart-table.js
@@ -12,6 +12,7 @@
   };
 
   var WITHHELD_FLAG = 'Withheld';
+  var NO_DATA_FLAG = undefined;
 
   var setYear = function(year) {
     var root = d3.select(this);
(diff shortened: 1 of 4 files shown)


js/components/bar-chart-table.js
```diff
@@ -12,6 +12,7 @@
   };
 
   var WITHHELD_FLAG = 'Withheld';
+  var NO_DATA_FLAG = undefined;
```
gemfarmer: not necessary, but imo, helpful.

meiqimichelle:
WOOOOOO I think this is looking good! 💥 🐭 🎉

Issue #1822, opened 2016-09-01
Maps with only one region have no legend


shawnbot:
I noticed this in the [Alaska offshore region production](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-regions/offshore/alaska/#federal-production) and [Pacific](https://federalist.18f.gov/preview/18F/doi-extractives-data/offshore-regions/offshore/pacific/#federal-production) maps:

![image](https://cloud.githubusercontent.com/assets/113896/18179798/fd1db0bc-7038-11e6-8283-a4b817088a6a.png)

Looks like it only happens when there's one value in the data (one region with data).

(originally mentioned in [this comment](https://github.com/18F/doi-extractives-data/pull/1812#issuecomment-244172633))

shawnbot:
Fixed in #1865!

PR #1865, opened 2016-09-07
Legend bugs

gemfarmer:
Fixes issue(s) #1822 

### Previews

[:sunglasses: offshore alaska – NOW](https://federalist.18f.gov/preview/18F/doi-extractives-data/legend-work/offshore/alaska)

[😢  offshore alaska – before ](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/offshore/alaska)

[:sunglasses: NV – NOW](https://federalist.18f.gov/preview/18F/doi-extractives-data/legend-work/explore/NV)

[😢  before NV - before ](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/NV)

### Changes proposed in this pull request:
- make color domain 5 instead of 9
- make legend show and hide depending on whether or not it has data
- fix bug that was breaking the legend when there were less than two counties and one of them was `Withheld`

/cc @shawnbot

### DIFF
--- a/_includes/location/offshore-area-federal-production-table.html
+++ b/_includes/location/offshore-area-federal-production-table.html
@@ -1,5 +1,6 @@
-{% assign product_ref = include.product_ref %}
-{% assign caption = include.caption %}
+{% assign product_ref = include.product_ref | default: caption %}
+{% assign caption = include.caption | default: caption %}
+{% assign product_name = include.product_name | default: product_name %}
 
 {% assign units = county[1].products[1].units %}
 {% if include.long_units %}
@@ -12,7 +13,7 @@
        class='county-table'>
   <thead>
     <tr>
-      <th>{{ region_name }}</th>
+      <th>{{ locality_name }}</th>
       <th colspan='2' class='numeric' data-series='volume'>{{ units | capitalize }} of {{ product_name | downcase | suffix:units_suffix }}</th>
     </tr>
   </thead>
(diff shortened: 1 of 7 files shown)


js/components/eiti-data-map.js
```diff
@@ -4,6 +4,7 @@
 
   var eiti = require('./../eiti');
   var WITHHELD_FLAG = 'Withheld';
+  var NO_DATA_FLAG = undefined;
```
gemfarmer: not necessary, but helpful, imo

_includes/location/offshore-area-federal-production-table.html
```diff
@@ -12,7 +13,7 @@
        class='county-table'>
   <thead>
     <tr>
-      <th>{{ region_name }}</th>
+      <th>{{ locality_name }}</th>
```
gemfarmer: Makes the column header display as `Region` instead of the particular region's name, in accordance with how we are approaching this on other pages.

gemfarmer:
@shawnbot ready for review!

js/components/eiti-data-map.js
```diff
@@ -173,9 +174,19 @@
 
           var marks = this.marks;
 
+          var stripWithheld = function(marks) {
+            marks = marks.filter(function(mark) {
+              return mark !== WITHHELD_FLAG && mark !== NO_DATA_FLAG;
+            });
+            if (!marks[1]) {
+              marks.unshift(0);
```
shawnbot: What is this doing? Making sure that the list of marks always has more than one value?

meiqimichelle:
Lookin good! The 5 bucket seems to solve several problems at once, at looks pretty elegant, too.

js/components/eiti-data-map.js
gemfarmer: Yes, probably not in the prettiest way. How would you improve this?

gemfarmer:
```if (marks.length < 2) {
         marks.unshift(0);
```
??

shawnbot: That works, and maybe a comment describing why it's necessary?

gemfarmer: for sure!

Issue #1850, opened 2016-09-06
Update self-employment jobs data to match BLS


shawnbot:
In #1833 I updated the BLS data for jobs, but the self-employment data comes from a separate source. We need to update that data to match the 2006-2015 year range.

shawnbot:
This is done.

PR #1867, opened 2016-09-08
Update self-employment data

shawnbot:
Fixes issue(s) #1850, with the following caveat:

**BEA hasn't released 2015 figures yet, so the latest we can get here is 2014.**

@coreycaitlin @meiqimichelle: Along with exports, this is the only other data set for which we don't have 2015 data. I've updated the "now" year to 2015 and made self-employment and exports sections use a "local now" of 2014 so we can still show the latest data figures rather than nulls.

### :sunglasses: Previews
* [Explore](https://federalist.18f.gov/preview/18F/doi-extractives-data/update-self-employment-2015/explore/#economic-impact)
* [Montana](https://federalist.18f.gov/preview/18F/doi-extractives-data/update-self-employment-2015/explore/MT/#economic-impact)
* [Utah](https://federalist.18f.gov/preview/18F/doi-extractives-data/update-self-employment-2015/explore/UT/#economic-impact)
* [Delaware](https://federalist.18f.gov/preview/18F/doi-extractives-data/update-self-employment-2015/explore/DE/#economic-impact) (has some null years)

- Move the BEA data script to `data/jobs` and adds a `Makefile` to update `self-employment.tsv`
- Update the BEA script to fetch 2006-2015 data
- Update the self-employment data file to include 2014 (and exclude 2004-2005)
- Make all years in `site.data.years` numbers, and stringify them in the templates that use them
- Implement "local now" year values for self-employment (jobs) and exports sections on both state and national pages

/cc @gemfarmer


gemfarmer:
@shawnbot do we want to keep 2005 data so that we can show a 10 year window from 2005-2014?

_data/years.yml
```diff
@@ -1,6 +1,8 @@
 # XXX this must be a string to use it as an index into other data!
-now: '2013'
+now: 2015
 
 default: [2006, 2015]
 
 exports: [2011, 2014]
+
+self_employment: [2006, 2014]
```
gemfarmer: do we want `self_employment: [2005, 2014]` instead?

_data/state_self_employment.yml
```diff
@@ -246,13 +219,10 @@ DC:
...
     percent: 0.06
+  '2014':
+    count: 554
+    percent: 0.06
 DE:
-  '2004':
-    count: 142
-    percent: 0.03
-  '2005':
-    count: 0
```
gemfarmer: Are these supposed to be 0? Or no data?

gemfarmer: `self-employment.tsv` shows values of 0, but I'm not sure if that is how it is supposed to be, or a transform issue.

gemfarmer:
WOO! 

@shawnbot Other than the little comments that I left below, this looks great!! No more blank 2014-2015!

_data/years.yml
shawnbot: Not sure. We don't have any other data sets that go back to before 2006 now, so I _think_ it makes more sense to treat that as the common start year. What do y'all think, @meiqimichelle @coreycaitlin?

_data/state_self_employment.yml
shawnbot: Hmmm yeah, let me take a closer look.

shawnbot: If you look at the diff in d0de1fcad13360401e9be4ac0952a4ecc9cacc37 you can see that there have always been 0 values for Delaware. I think that's a legitimate zero, not null.

shawnbot: Ah, crap. I can't link directly to this view in the [BEA Interactive Data explorer](http://www.bea.gov/itable/iTable.cfm?ReqID=70&step=1#reqid=70&step=30&isuri=1&7022=5&7023=0&7024=naics&7033=-1&7025=0&7026=10000&7027=2014,2013,2012,2011,2010,2009,2008,2007,2006&7001=45&7028=200&7031=0&7040=-1&7083=levels&7029=31&7090=70), but it looks like they've got their own way of expressing withheld:

image

gemfarmer: Crap! Well at least we are already handling withheld data, thought the `(L)` category is a bit troublesome.

shawnbot: Yep. See #1870. I think we can merge this for now, since those numbers have always been wrong.

PR #1885, opened 2016-09-09
National revenue fix

### LINKED ISSUE #1873: Bug: national revenue charts, asphalt has 'all's data
Looks like we might be chunked over one for our charts. All commodities should have the data that's coming up for asphalt. Maybe all our data is moved over one slot?
screenshot 2016-09-08 12 35 37

shawnbot:
Fixes #1873. The issue here is that the assignment of the variable used to populate the templates was happening _after_ the chart output, which means that all of the charts were being populated with the previous product's data. 😬 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/data-fixes/explore/#revenue-trends)

- Simplify national revenue data by removing the `revenue` key from yearly entries, and updates the national revenue template accordingly.
- Add a `revenue-trends` id to the explore page for easier linking.

/cc @gemfarmer @meiqimichelle


meiqimichelle:
Hahah, oops! Looks good now! 🐭 💥

Issue #1365, opened 2016-04-07
How do we express state revenue and production rankings?


shawnbot:
While building out state prototypes (#1355) and our new data "infrastructure" (#1353) I noticed one issue with state rankings: they might be a little complicated to describe if they're derived from state _and_ offshore region data. For instance, in 2013 California was the "top" _state_ producer of Fuel Oil at a whopping $68, but the national total in that year was actually negative $73k, with Western Gulf of Mexico contributing the overwhelming majority.

So there are a couple of questions here:

1. **Do we include offshore regions in our rankings?** (This is the important one.)
1. How do we express contributions to a negative revenue total?

For 2. I was thinking that we would derive a percentage by taking the absolute values of the regional and national totals, and dividing those. So `abs(68) / abs(-73725) * 100` yields .09%. This accounts for both the edge cases of states with negative revenue in a positive national total and vice-versa.

RyanSibley:
Link to a notes doc after reviewing the California page. I don't actually get to your question about negative numbers until #6:

https://docs.google.com/document/d/1Yju7CYTc90SFg6KaE0vJIgJ-84qWJIZsXRVnellXRW0/edit

meiqimichelle:
This is still underway as part of the state profile updates.

shawnbot:
My thoughts on this have evolved. From what I've read around the internets, it sounds like the best practice for negative contributions to a positive total (or vice-versa) are _not_ typically expressed as a numeric percentage, because they don't really make sense.

I can dig into my history a bit to find some references, but I think the thing to do here is to update our state rankings and percentages to be `null` (undefined) if the numerator and denominator in the percentage fraction for a given state have different signs (±).

shawnbot:
While investigating #1397, I queried the db for state revenue rankings with a % > 100 and found some other offenders (note: these values are truncated for legibility, so the ones with % = 100 are actually between 100 and 101):

year | state | product | $ revenue | $ total | % | #
---: | :---- | :------ | ------: | ----: | ------: | ---:
2004 | ID | Phosphate Raw Ore | 3451909 | 3451909 | 100 | 1
2004 | WY | Helium | -21423 | -21365 | 100 | 2
2004 | WY | Soda Ash-Granular | -583138 | -449974 | 129 | 3
2005 | CA | Soda Ash-Granular | -726947 | -719067 | 101 | 2
2006 | MN | Copper | 28720 | 18072 | 158 | 1
[...]

PR #1886, opened 2016-09-09
Data fixes

### LINKED ISSUE #1874: Bug: national revenue page seems to be only showing onshore
Our national totals are about $5 billion off. Looks like we're only pulling onshore data. We need pull onshore and offshore :) Or, we're only pulling onshore for 2014 and 2015 (and things are OK for the older years).

shawnbot:
Fixes #1874.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/data-fixes/explore/#revenue-trends)

[Before](https://federalist.18f.gov/preview/18F/doi-extractives-data/dev/explore/#revenue-trends) and [After](https://federalist.18f.gov/preview/18F/doi-extractives-data/data-fixes/explore/#revenue-trends)
image

- Update offshore revenue using FY 2006-2015 data from ONRR
- Update data transforms and SQL rollups to match new data
- Update SQL statements that generate state revenue percentages to use `null` if the revenue total is negative per #1365.

/cc @meiqimichelle @mentastc @gemfarmer


meiqimichelle:
Looking good to me! I'm going to merge, and then will have @mentastc review on Monday when we meet.

Issue #1884, opened 2016-09-09
Map, table, and bar chart polishing


ericronne:
blowing this out in epic fashion 💃 ...

## Other
 - [ ] ~~Year selectors should only display "no data" when opened. Closed selectors currently display a truncated "[year] no d." Instead they should only show the year.~~ moved to #2048
 - [x] The charts and maps should horizontally align along the grid. Currently the sets with multiple bar charts don't quite line up with the bar / map combos (one consequence of this is that the bar charts are too close to the maps)

## Bar charts
 - [x] Some bar charts wiggle up and down a smidge when you scrub the mouse across them (seems to occur on charts where some years have no data)
 - [x] All bars should be the same width. If a category has fewer than ten years of information, we should fill up the rest of the chart with "no data" spaces. I think the only offenders here are exports for the three opt-in states (AK, MT, WY)

## Maps
 - [x] Tool tip formatting ( #1992 )
 - [ ] ~~See about adding back a gray ownership texture layer~~
 - [x] Reformat the wide-map legend, using square swatches
 - [x] The numbers and bar still shift back and forth a little sometimes
 - [x] Fix outline-layering issue with counties
 - [x] Bar chart fills should always be 100% solid (no alpha). Since they change to green on hover, we don't need to have the current alphaed-back effect.

 - [ ] ~~Reposition some table headers (eg ones which are over bars but should be over the numbers). We probably have to go through individually :/~~ (#2063)
 - [x] Numbers in tables for opt-in states (AK, MT, and WY) should run to the left of their bars (currently run below them), and be right aligned. (Headers will need to be re-aligned, too, ack.)

## Tables
 - [x] Should we auto-scroll to the table when someone clicks a county? If the map is tall (eg CA) or the user hasn't scrolled far enough, clicking counties appears to have no effect (other than darkening the outline).
 - [ ] ~~The national disbursements chart (and others?) shouldn't have outlines or dividers~~ (#2064)

thx!

meiqimichelle:
Are these all still in play? If yes, might want to break out into separate issues for easier tracking, and to decide which might be a part of this sprint, and which can wait.

gemfarmer:
Some of these are still in play, and yes, it might be a good idea to break them into separate issues. @ericronne thoughts?

ericronne:
Yes unchecked = not done :)

I'll break em out

ericronne:
Split out remaining work into new issues...

PR #1907, opened 2016-09-12
Style polish

gemfarmer:
Fixes issue(s) #1884 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/style-polish/)

- Add support for `page.is_cropped`. This basically allows us to make it so that a certain state uses its smaller "cropped" viewbox instead of the default viewbox.
- Adjustments to cropped viewboxes	
- adjust viewbox threshold ratio
- slight adjustments to vertical legend
- use cropped viewboxes for the following states:
  - AL
  - CA
  - DC
  - GA
  - ID
  - KS
  - ME
  - MS
  - NH
  - NJ
  - VT

/cc @ericronne


js/components/eiti-data-map.js
```diff
@@ -234,6 +236,7 @@
               .shapeHeight(shapeHeight)
               .labelDelimiter(legendDelimiter)
               .shapePadding(shapePadding)
+              .labelOffset(shapeMargin)
```
gemfarmer: I defined `shapeMargin`, but hound still thinks that I didn't.

Issue #1820, opened 2016-09-01
Data Check


gemfarmer:
@meiqimichelle, @mentastc, Nathan, and I had a meeting to discuss the current state of data on EITI.

# State home

## Federal Production
- `NULL` commodity field in bar charts
- `Explore data` showing up as `state_name`. Should be something else.
- Inlet Scrubber (and other commodities that don't have '14 and '15 values) somehow made it into the yml output file. It isn't in the source data and isn't in the `federal_regional_production` database. Related to #1797

### Pulling in wrong dataset for offshore
[offshore.tsv](https://github.com/18F/doi-extractives-data/blob/dev/data/federal-production/offshore.tsv) should not be used. [onshore.tsv](https://github.com/18F/doi-extractives-data/blob/dev/data/federal-production/onshore.tsv) is the only source of truth for federal production onshore AND offshore

- Add EITI Production CY06-15 Geothermal Update to GDrive

### Incorrect values
- Oil is significantly off.
- Salt is wrong for 2006-2013, correct for 2014-15

### Sentinel values (onshore)
Oil | 2013: 623,103,687.09 barrels
Salt | 2015: 764,801 tons
Salt | 2006: 512,832 tons
Gas | 2013: 4,941,408,110

## Federal Revenue
- The year be set to 2015 for the federal revenue by type chart and default for other fed rev data
- Nathan: NGL could be lumped into Oil & Gas (because NGL will always be done During Production).
- Add Other Revenues and Civil Penalties (not tied to a lease) to revenue by type breakdown. Should get added to All Commodities rollup. Maybe give them a caveat that these are revenues not tied to a  product or commodity. Note: this is not as high a priority as other data issues.

# State pages

## Montana – Federal Production
- Correct commodities: Coal, Gas, Oil
- County and state level data is correct

## Misc
- county dropdown tables list `County` in the rows for Production, but not for revenue.

# Offshore

### Sentinel stuff
- Should only have Oil, Gas and Salt for federal production
- Salt | 2015 | Central Gulf of Mexico | Withheld

# TODO
- make a list of links to the source datasets that we are using to populate the site for @mentastc and Nathan to double check.

gemfarmer:
@meiqimichelle @mentastc As of now, all the sentinel values check out on dev!

gemfarmer:
@mentastc should we be expecting `623,103,687.09` barrels of Oil in 2013 *nationwide*? Or  `623,103,681.09`?

[...]

PR #1911, opened 2016-09-13
Federal production sentinel tests

gemfarmer:
Fixes issue(s) #1820 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/fed-production-sentinels/)

Two data issues have come out of this issue:
- Withheld Salt data in the Central Gulf of Mexico is associated with Louisiana instead of Gulf in the [source data](https://github.com/18F/doi-extractives-data/blob/dev/data/federal-production/federal-production.tsv#L716) (see below)
- Oil data is slightly off in the national rollups. It is hard to tell _where_ it is off without creating tests that test all the values, so a follow up, comprehensive federal production data test suite will be needed. :(

screen shot 2016-09-13 at 10 00 04 am

- adds data tests for federal production that test sentinel values provided by @mentastc . These are specific, cherry picked values that we _know_ to be true and will exist as an indicator that the rest of the data is correct 
- there are two pending tests for the issues mentioned above. They run successfully when the above issues are excluded from the tests.

/cc @shawnbot @mentastc @meiqimichelle


test/data/federal-production.js
```diff
@@ -0,0 +1,211 @@
...
+    var assertSentinelMatch = function (products, product, year, value) {
+      var expected = Math.round(value);
+      try {
+        var actual = products[product].volume[year];
+      } catch (error) {
+        assert.ok(false, "product doesn't exist" + error);
+      }
+
+      assert.equal(
+        expected, actual,
```
gemfarmer: This is no longer an issue

gemfarmer:
@shawnbot @mentastc would love your eyes on this when you get a chance!

PR #1918, opened 2016-09-14
[WIP] Makes top of state pages look scrollable. We hope.

### LINKED ISSUE #1868: Sticky-sidenav pp: Consider renaming "Overview" to "Top" (& bug fix)
Since there is no "Overview" header — and this links back to the top of the page — I think "Top" might be more clear to users wherever we have the sticky sidenav. Agree?

Also, bug: the link doesn't currently take the user all the way back to the top of the page.

@coreycaitlin @meiqimichelle @gemfarmer

meiqimichelle:
For #1612

Also fixes #1868 by double-checking that all nav items say the same thing for the first entry (top) and that they all anchor link somewhere reasonable.

Also also adds legend to explore home.

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/will-they-scroll/explore)

screenshot 2016-09-13 21 26 56

@gemfarmer is there a way to make the padding under the states smaller (now that there's a background color, this is more obvious)

/cc @ericronne

### DIFF
--- a/_includes/location/section-overview.html
+++ b/_includes/location/section-overview.html
@@ -1,6 +1,10 @@
-<section id="overview">
+<section class="state-pages-top">
 
   <section class="container">
+    
+    <div class="state-pages-ownership">
+      {% include location/section-ownership.html %}
+    </div>
 
     <!-- Includes the GDP percentage, then outputs employment percentage if it's over 2%.-->
     <p>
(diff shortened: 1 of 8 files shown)


ericronne:
I think this is definitely more user friendly. We can probably make the state one column(?) narrower / smaller, and do a couple other cleanups, but this feels like the right direction to me. Will deal with that other stuff in an issue or pairing!

Issue #1748, opened 2016-08-16
Chart and map bugs in ffx


meiqimichelle:
when the default year has "no data":
<img width="790" alt="screenshot 2016-08-16 10 49 45" src="https://cloud.githubusercontent.com/assets/4827522/17705806/55abb23e-639f-11e6-8977-e62d94462281.png">

meiqimichelle:
We won't know if the above issue is solved until we fix this other issue, which is that the charts aren't working in FFx right now.
<img width="742" alt="screenshot 2016-09-12 15 56 20" src="https://cloud.githubusercontent.com/assets/4827522/18455870/7e56e8f8-7901-11e6-9cf8-f31cefa92b19.png">

meiqimichelle:
Tooltips are also not showing up. The custom element code isn't even running. So, maybe all these issues are related to something in Ffx not recognizing custom elements.

![screenshot 2016-09-14 09 56 06](https://cloud.githubusercontent.com/assets/4827522/18521620/a3f4a3a0-7a61-11e6-96c8-dc2faf96cf9c.png)

shawnbot:
Oof. Firefox appears to be throwing an error inside [d3-svg-legend](https://github.com/susielu/d3-legend), specifically [here](https://github.com/susielu/d3-legend/blob/c02f50490c28bd4906c97e0c77adfb06ffe858ef/src/color.js#L53). I _think_ what's happening is that we're trying to access a `text` node inside a linked SVG and that's throwing an error because Firefox has more restrictive cross-origin rules than Chrome?

PR #1921, opened 2016-09-14
Map tooltips

gemfarmer:
Fixes issue(s) #1520 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/map-tooltips/explore/)

- adds an `eiti-tooltip` component to be used in conjunction with the `<title>` element

`eiti-tooltip` can be used like so: `<title is="eiti-tooltip">` and supports `desc` and `alt` attributes.

`desc`: text to be displayed
`alt`: applied to the `aria-label` attribute on the tooltip

Enhancement possibility: make tooltips clickable. Add `href` attribute support

/cc @shawnbot  @ericronne

### DIFF
--- a/_includes/maps/offshore-regions.html
+++ b/_includes/maps/offshore-regions.html
@@ -4,7 +4,8 @@
   <g class="state offshore-area feature" fill><!--- FIXME: nix fill attribute -->
     <title is="eiti-tooltip"
            desc="{{ region.title }}"
-           alt="{{ region.title }}"></title>
+           alt="{{ region.title }}"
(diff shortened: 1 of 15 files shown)


meiqimichelle:
The tooltips aren't working in Ffx. Pulled that out into the existing Ffx bug issue because it might be the same thing. #1748 .

gemfarmer:
@shawnbot @ericronne @meiqimichelle now works on 🔥 🐺

gemfarmer:
Also, the hound warnings are old. The grievances it is voicing are not real.

meiqimichelle:
😄 I will review once Federalist builds! 🎉

PR #1934, opened 2016-09-14
Tooltip follow up

gemfarmer:
[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/tooltip-followup/)

- commented out additional tab tiles on homepage
- added a tooltip offset that limits the amount of flashing that people see when they rapidly change states
- fixes bug that was preventing the tooltip from hiding when exiting the svg

/cc @meiqimichelle

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,7 +76,7 @@ selector: location
       </aside>
     </section> -->
 
-    <section class="eiti-tab-panel explore_home-main" id="production" role="tabpanel" aria-hidden="true">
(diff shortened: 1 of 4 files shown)


meiqimichelle:
I'm not sure the 'hiding when exiting svg' is working -- on my end, the tooltip still hangs out even when I mouse-off the map. But! The rest looks great, merging. I won't delete the branch in case you want to continue work on that tho.

Issue #1938, opened 2016-09-14
Visual bug: Sticky nav positioned too low on Explore main page


ericronne:
## Expected Behavior
![image](https://cloud.githubusercontent.com/assets/9259185/18532261/b6167f42-7aa0-11e6-8716-d342c13f687f.png)

## Actual Behavior
![image](https://cloud.githubusercontent.com/assets/9259185/18532256/a11e2630-7aa0-11e6-8756-1db7bdaf3960.png)

## Steps to reproduce the behavior
Visit page, scroll a bit...

@meiqimichelle @gemfarmer

meiqimichelle:
Closed with #1951.

PR #1951, opened 2016-09-15
Map polish follow up

gemfarmer:
Fixes issue(s) #1938, other assorted issues

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/homepage-map-polish-followup/)

- center-aligned map legend on explore and homepage
- Updated the selector to include 'Nationwide data' as an option, default option
- addressed #1938

/cc @ericronne @coreycaitlin

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,7 +28,7 @@ selector: location
         Explore production, revenue, and economic impact data for each state, as well as additional contextual information and data about three states that chose to participate in deeper USEITI reporting: <a href="{{ site.baseurl }}/explore/AK/">Alaska</a>, <a href="{{ site.baseurl }}/explore/MT/">Montana</a>, and <a href="{{ site.baseurl }}/explore/WY/">Wyoming</a>.
       </p>
       <figcaption class="state_pages-select">
-        <label for="location-selector">Choose one to a location to explore USEITI data:</label>
+        <label for="location-selector">Choose a location to explore USEITI data:</label>
         {% include selector.html
            nationwide = true
            nationwide_url = '/explore/'
--- a/index.html
+++ b/index.html
@@ -28,12 +28,12 @@ selector: location
         Explore production, revenue, and economic impact data for each state, as well as additional contextual information and data about three states that chose to participate in deeper USEITI reporting: <a href="{{ site.baseurl }}/explore/AK/">Alaska</a>, <a href="{{ site.baseurl }}/explore/MT/">Montana</a>, and <a href="{{ site.baseurl }}/explore/WY/">Wyoming</a>.
       </p>
       <figcaption class="state_pages-select">
-        <label for="location-selector">Explore data</label>
+        <label for="location-selector">Choose one to a location to explore USEITI data:</label>
         {% include selector.html
            nationwide = true
            nationwide_url = '/explore/'
            nationwide_title = 'Nationwide data'
-           default = 'Choose one (COREY???)'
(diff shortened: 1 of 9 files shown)


index.html
```diff
@@ -23,14 +23,18 @@ <h3 class="ribbon-hero-description-light">
...
       </p>
       <figcaption class="state_pages-select">
-        <label for="location-selector">Choose a state or offshore region</label>
-        {% include selector.html %}
+        <label for="location-selector">Explore data</label>
+        {% include selector.html
+           nationwide = true
+           nationwide_url = '/explore/'
+           nationwide_title = 'Nationwide data'
+           default = 'Choose one (COREY???)'
```
gemfarmer: @coreycaitlin we were struggling to pick a default set of text here. To be consistent with the previous layout, we would want `Choose a state or offshore region`, but the width of the selector doesn't fully accommodate that text. Any ideas for how to succinctly communicate what people can select?

meiqimichelle:
Woot! As soon as we get words from Corey, we're good to go here, methinks!

index.html
coreycaitlin: I did this:

screen shot 2016-09-15 at 10 25 30 am

coreycaitlin:
I'm seeing weird padding on the explore/national page still @gemfarmer

ericronne:
@coreycaitlin let's chat! ;)

gemfarmer:
@coreycaitlin where is the padding weirdness?

gemfarmer:
Also, I like the new wording!

coreycaitlin:
@gemfarmer Here's the weirdness:

screen shot 2016-09-15 at 10 49 01 am

gemfarmer:
@coreycaitlin oi!

gemfarmer:
@coreycaitlin how about now?!

ericronne:
@coreycaitlin thanks 👍 

Digging the top label, but "state or offshore region" feels a little long to me. Also it isn't technically accurate, right? (Since you could also choose national data.) I wanted to simply show "Nationwide data" as the default, but i think that's cumbersome to code (and a little odd from an interaction standpoint). Any more general option we could drop in there?

coreycaitlin:
@ericronne Here are a few of the options I considered:

- `National, state, or offshore` — way too long and somehow more confusing?
- `Choose a state or offshore region` — way too long, but yay verb!
- `Location` or `Choose location` — technically accurate, but it feels pretty cryptic; "location" is another layer of abstraction up, and doesn't tell people what they can expect to see in the list
- `State or offshore region` — not technically _quite_ right, since "nationwide" is also in the list, but it felt least bad because it tells people what to expect; since "nationwide" is first, I'm not too worried about people missing it if they do click

My priorities here are:

- Getting verbs ("choose" and "explore") into either the label or the default
- Telling people what to expect when they click (i.e. a list of states and offshore regions, mostly)
- Brevity

I think either `Choose location` or `State or offshore region` could work; if we go with `Choose location` I'll to re-write the content immediately above to remove "location."

ericronne:
You are a killer content designer. Just sayin.

Loving "Choose location"
Could it be as simple as "Explore data" / "Choose location"??

ericronne:
@coreycaitlin also if you dig in there to make updates, could you enlarge the first label ("Explore data:" or wherever we land) to 1rem? thx!

Issue #1961, opened 2016-09-15
Rewrite disbursements content to obviate tables


ericronne:
@coreycaitlin i believe you said that we could eliminate these tables via clever wordsmithing. True?? Thxoxo!

ps for emphasis, please use $weight-book (rather than STRONG)

![screenshot_2016-09-15_15 14 29_720](https://cloud.githubusercontent.com/assets/9259185/18569425/1a94b854-7b68-11e6-8cd5-cbcb00f75f2d.png)

cc @meiqimichelle

Issue #1935, opened 2016-09-14
Use consistent apostrophes throughout


shawnbot:
@femmebot noticed in #1933 that some of our apostrophes are uneducated (`'` vs. `’`). I did a cursory search and found a bunch of instances:

```
_case-studies/campbell.md:166:[^27]: Wyoming Legislature, [Wyoming's Water Development Program (PDF)](http://legisweb.state.wy.us/progeval/REPORTS/2015/WWDCScoping8-7-2015b.pdf), 2015
_case-studies/greenlee.md:25:> Copper is a major industrial metal used in construction, electronics, transportation, industrial machinery, and consumer products. In 2014, the U.S. was the world’s fourth-largest copper producer, mining 1.127 million tons of copper worth a total value of approximately $9.7 billion.[^1] Of the five major copper-producing states (Arizona, Utah, New Mexico, Nevada, and Montana), Arizona increased production most in 2014. Arizona's copper output totaled 893,000 {{ "metric tons" | term_end:"metric ton" }}, or 66% of the national total.[^2] Copper represented 89% of the total value of mining in Arizona, with Greenlee and Pima counties generating the majority of that production.[^3] If Arizona were a country, it would be the seventh largest copper producer in the world.[^4]
_case-studies/north-slope.md:25:> The U.S. has experienced rapid change in domestic oil production since 2008, when crude oil production reached a low of 3.98 million {{ "bbl" | term:"barrel" }} per day. Just five years later, the U.S. had nearly doubled its daily production output, with Texas and North Dakota driving much of the growth.[^1] [^2] Alaska did not experience the same production boom, with crude oil output steadily declining over the past decade.[^3] In spite of that downward trend, Alaska remained  the fourth largest state producer of crude oil in 2015, and the nation's largest oil-producing county is Alaska's North Slope Borough.[^4]
_case-studies/north-slope.md:33:The North Slope Borough is the country's largest organized local jurisdiction, spanning more than 94,000 miles north of the Arctic Circle (66th parallel). Its 9,703 residents, most of whom are Inupiat Alaskan Natives, are spread across eight separate communities.[^5] The northern coast of Alaska was documented as a potential oil-producing region as early as 1900. However, the borough's government was not formally incorporated until 1972, soon after the discovery of oil at Prudhoe Bay, the largest single oil field in North America.[^6] [^7]
[...]

Issue #1920, opened 2016-09-14
Enhancement Proposal: Simplify land ownership legend labels


ericronne:
The ownership maps currently say "lands" four times (title plus each legend label). @coreycaitlin do you think it would be confusing if we removed the word from federal and tribal, like so? ;) ...

![image](https://cloud.githubusercontent.com/assets/9259185/18515354/b4b8b82c-7a5a-11e6-9b99-d9b823b26f31.png)

cc @meiqimichelle

ericronne:
@gemfarmer ...

As discussed in PR #1965, we'd like to create an exception for the legends two large maps (on home and explore main). The point is to change the wording on all of the state-page ownerships maps, removing "lands" from two spots. Like so ...
![image](https://cloud.githubusercontent.com/assets/9259185/18593221/0268fae0-7c00-11e6-961a-76101457a4a8.png)

The theory is that on those pages, the "Land ownership" title provides enough context for the user to get the point.

The large maps on the home and explore main pages don't have such a title, though. So for those two maps, could we keep the legend as is? I.e....
![image](https://cloud.githubusercontent.com/assets/9259185/18593282/586cf428-7c00-11e6-89fb-1035efd988b2.png)

meiqimichelle:
We should also consider whether adding this will add unnecessary complexity to maintaining the site :-/

ericronne:
Not a fan of complexity, but these two are super high-profile instances of the map. With fifty or so instances of above-the-fold redundancy. Your call

gemfarmer:
@meiqimichelle @ericronne Is this done?

ericronne:
done, but new bug introduced on home page map, where "land" was mistakenly deleted from the legend. made a new issue  #2020 👓

Issue #1560, opened 2016-07-01
State pages: Miscellaneous content and data notes


coreycaitlin:
We're going to need some footnotes on the state pages:

- [ ] ~~Geothermal rate details, with links to further detail~~
- [x] Withheld data
- [ ] ~~Oil and gas groupings~~
- [ ] ~~Possible double-counting in self-employment data~~
- [ ] ~~Mention of county-level coal data?~~

coreycaitlin:
Content collected from other spots:

> In some cases, the specific amount of a resource produced in a state or county is withheld to protect trade secrets.

> In a few cases, usually where only one company is extracting a resource in a county, production data may be withheld to ensure there are no violations of the Trade Secrets Act.

> Companies may not know whether they’ll produce oil or gas until they explore the land, so fees are not differentiated by product until production begins.

> Self-employment data describes people who work in natural resource extraction, but don't receive wages or salaries because they own their own companies. Some people may be double-counted (for instance, if they appear on multiple tax forms).

> For coal, there is also data about mining in each county.

coreycaitlin:
Once we implement footnotes, we should send screenshots/links of the geothermal info to Herb for validation.

coreycaitlin:
ALSO: explain that "All commodities" in the megachart includes civil penalties.

coreycaitlin:
Renamed because we aren't going to implement this as footnotes, but I still need to find places to mention the things collected here.

coreycaitlin:
Closing this out — see new issues for details.

PR #1965, opened 2016-09-16
Words, apostrophes, tables, etc.

coreycaitlin:
Fixes issue(s) #1920, #1935, #1961 and adds a `withheld` definition, per #1560 

[:dancer: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/wordy-bugs/)

- definition for "withheld"
- fix some anchor links and an awkward widowed word on the How it Works landing page
- removes the one-row table in state disbursements sections and replaces it with a very brief list when relevant (see AK, UT, and IA for the three states possible in this section)
- removes "land" from the first 2/3 of the land-ownership legend because it's still clear what's happening there
- fixes some errant incorrect apostrophes in non-markdown content chunks

/cc @ericronne @meiqimichelle


coreycaitlin:
Tacked on these changes:

- makes the state-selector dropdown on the Explore landing page more consistent with the homepage, which also resolves the problem where we couldn't get to Alaska
- addresses concerns from the online advisory group that the wording in the Explore intro was implying a narrower scope (i.e. only federal land)
- changes default bold weight to `$weight-book` (I *think* this wasn't addressed elsewhere; sorry if I duplicated work here! I also checked quite a few pages and it looks like this change doesn't do anything weird/unexpected elsewhere)

coreycaitlin:
2 more things!

- changes "Accounting standards" to "Audits and assurances" on the How it Works landing page, thanks to feedback from the online advisory group
- adds a definition for "mineral acres"

ericronne:
Good stuff! The "land" redundancy in the ownership map legend may actually be helpful on the home page and explore main pages, since those maps aren't titled "Land ownership." Agree? Sorry for the confusion on that.

ericronne:
Oh also the title of the page "Audits and assurances" links to is still "Oversight." Should we change for consistency?

There are other such issues on that page, i can help debug

coreycaitlin:
@ericronne the land ownership map legend on the home/explore pages and the state pages is driven by the same file, so changing it once changes it both places.

ericronne:
Wondered if that was the case. If it's easy to make an exception for these two big maps, how do you feel about that?
cc @gemfarmer

coreycaitlin:
@ericronne No objections from me, but I think it would have to go in @gemfarmer's list of things. Which do you want me to leave it at for now?

ericronne:
Let's stick with verbose for now (clarity ftw), and deal with the whole thing in one swoop when @gemfarmer gets the time. I'll make a new issue for that.

Issue #1947, opened 2016-09-15
Bug: missing values in 'other revenue' for oil and gas


meiqimichelle:
There are a few more lines of data that should be summing in the far right Oil & Gas line in the megachart under 'Other revenue'. This image is from Wyoming, I'm talking about the -$20,886,037 number. We think this is only an issue on our onshore data.
![screenshot 2016-09-15 08 41 46](https://cloud.githubusercontent.com/assets/4827522/18556527/65a4b1f2-7b20-11e6-9788-f80e3fa88543.png)

Right now, it only includes other revenues for "Oil & Gas". There are lines in the data for other revenues "NGL" "Gas" and "Oil". They don't need to be broken out like we do in the 'during production' phase, but we should sum them so that total includes them. This will make the right side of the megachart add up correctly.

Side note: the bottom total (on Wyoming, the -$15,382,656 value) is correct as-is.

meiqimichelle:
ie, here's some of the relevant data:
![screenshot 2016-09-15 10 24 34](https://cloud.githubusercontent.com/assets/4827522/18560050/c6ad4d84-7b2e-11e6-9125-76754f1355d9.png)

We're pulling in Other Revenues that have a commodity type 'Oil and Gas', but not pulling in Other Revenues that have any other commodity type.

gemfarmer:
`Other revenue` should manually sum values for Oil, Gas, and NGL like it is being done in the During Production column IN ADDITION to adding the values for Other  Revenues.

shawnbot:
I'm still seeing the $20m number [in Wyoming](https://federalist.18f.gov/preview/18F/doi-extractives-data/megachart-bug-killing-spree/explore/WY/#federal-revenue), but the math checks out:

![image](https://cloud.githubusercontent.com/assets/113896/18574432/d5072f50-7b81-11e6-9ee4-b6e8877f1765.png)

#1957 is merged, but I'll leave this open until somebody else can confirm that it's fixed.

gemfarmer:
@shawnbot the $20m number is still there, but slightly different

OLD: `-$20,886,037 `
NEW: `-$20,887,881`

meiqimichelle:
This looks good, except for some offshore values which are now tracked in #1950.

Issue #1871, opened 2016-09-08
Revenue by type: Civil Penalties


gemfarmer:
Extra work from #1770 

The revenue megachart should have a Civil Penalties revenue type. Need to determine if we are using the correct data for offshore, because our offshore.tsv file currently does not have a Civil Penalties type.

cc @mentastc @meiqimichelle @shawnbot

meiqimichelle:
For offshore, we were missing royalties from other products. We need to add them back in, and list the product types like we do with 'hardrock minerals' on onshore.

Other Products
---------------------
Sulfer $xxxxxx
Salt $xxxxxx

meiqimichelle:
Ok! It turns out, we're talking about two numbers. We have a 'civil penalties' value, and a 'other revenues' that aren't linked to commodities value for 2015 that should be called out in the national revenues megachart, somewhat near the far bottom right column, ie:

![screenshot 2016-09-08 12 28 58](https://cloud.githubusercontent.com/assets/4827522/18363724/1c944168-75c0-11e6-9ecc-b3c540ca5c91.png)

meiqimichelle:
The values for 2015 are:

Civil penalties: $8,035,048.31 
Other revenues: $347,350.54

meiqimichelle:
This is the proper civil penalties file that should be included:
[EITI CY Civil Penalties and Other Revenues not tied to lease.xlsx](https://github.com/18F/doi-extractives-data/files/473251/EITI.CY.Civil.Penalties.and.Other.Revenues.not.tied.to.lease.xlsx)

gemfarmer:
Inspection Fees is in the sqlite database, but not in the yml outputs. Need to figure out how we are handling Inspection Fee, Other Revenues, and Civil Penalties before we add it to the yml.

gemfarmer:
* copy:  `Includes civil penalties, offshore inspection fees, and other revenues that are not categorized by commodity or location.`

gemfarmer:
Asterisk copy will go in the Other Revenue x All Commodities cell. The number will, as the copy suggests, include 2015 civil penalties, 2015 "other revenues" (both in the spreadsheet above), and the offshore inspection fees that are already in the sqlite db.

gemfarmer:
Can find Inspection Fees at `_data/offshore_revenue_areas/None.yml`

shawnbot:
[...]

Issue #1973, opened 2016-09-16
Bug: offshore 'all commodities' bar chart total is not correct


meiqimichelle:
On the offshore pages, the total under the bar chart is actually royalties instead of the all-revenue-streams total. ie, in this image, it should be the $4... number, not the $3... number. 

![pasted image at 2016_09_16 10_09 am](https://cloud.githubusercontent.com/assets/4827522/18591876/5221f904-7be9-11e6-9ba5-0ad562ea5a32.png)

meiqimichelle:
Looking further -- the All Commodities bar chart total on all our offshore areas aren't correct, thought they seem to be incorrect in different ways (ie, the others don't seem to be pulling royalties, but not sure what they're doing).

gemfarmer:
Another issue: The Civil Penalties figures didn't make it into the national All Commodities bar chart. For more detail, see https://github.com/18F/doi-extractives-data/pull/1977#issuecomment-247721503

gemfarmer:
Notes on this one. 

The issue here is that we have two different values for the national revenue rollups: `$7,507,964,282` and `$7,499,581,883`.

The difference between the two is `$8,382,399`, which is exactly the sum of Civil Penalties (`$8,035,048`)  and Other Revenues (`$347,351`), both of which were recently added. 

When we look at the sqlite queries, it paints a helpful picture. See that None|All line?? 😍 
```
sqlite> select * from regional_revenue_type where commodity = 'None' and year = '2015';
2015|None|All|8382398.85
2015|None|All|52239171.25
2015|None|Civil Penalties|8035048.31
2015|None|Inspection Fees|52239171.25
2015|None|Other Revenues|347350.54
```

Unfortunately, that sweet, sweet number didn't make it into the `regional_revenue` table _or_ the `national_revenue` table:

```
sqlite> select year, sum(revenue) as revenue from national_revenue where commodity != 'All' and year == '2015';
2015|7499581882.93
sqlite> select year, sum(revenue) as revenue from regional_revenue where commodity != 'All' and year == '2015';
sqlite> select * from regional_revenue where commodity = 'None' and year = '2015';
2015|None|None|offshore|52239171.25
```

So there's the mission: Get Civil Penalties and Other Revenues into the `regional_revenue` table (which creates the `national_revenue` table)!

meiqimichelle:
Yaaaay!

meiqimichelle:
Closed with #1998

PR #1977, opened 2016-09-16
Megachart data

gemfarmer:
Fixes issue(s) #1947, #1871 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/megachart-bug-killing-spree)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/megachart-bug-killing-spree/)

- #1871  – Civil Penalties
- #1947 - Follow up oil gas funkiness
- added asterisk, note in other revenues national revenue by type
- added Easement and Right of Way data into the offshore revenue regional rollups

/cc @meiqimichelle


test/data/revenue.js
```diff
@@ -117,62 +124,111 @@ describe('revenues by type', function() {
...
+      };
+
+      loadPivot(pivotSource, function(error, rows) {
+        rows.forEach(testRow);
+        done();
+      });
+
+
+    });
+  })
```
gemfarmer: fixed with a3ca8de

gemfarmer:
With that Easement and Right of the Way madness out of the way I think this baddy is ready for review, @meiqimichelle 🎉

meiqimichelle:
On the national profile page, the megachart includes civil penalties etc, but the All Commodities bar chart doesn't so the totals are off. Methinks we should include these new additions in the bar totals as well. ie, the $7,... on the top left and bottom left aren't the same:

screenshot 2016-09-16 15 13 41

meiqimichelle:
Otherwise, I think we're good since Chris checked the dataaaa! Yeeeeaaaaayyyy! Should we keep that other issue here, or pull it into our other total-mismatch issue, #1973 @gemfarmer ?

gemfarmer:
@meiqimichelle let's pull it into #1973

gemfarmer:
Ok. So all satellite bugs related to this have been accounted for, from what I can tell. I say it is good to go @meiqimichelle. Anything else I haven't noticed?

Issue #1994, opened 2016-09-19
Finish data tests on federal production


meiqimichelle:
Our sentinel data tests work. There are two or three tests out of six that aren't passing because the data isn't 100% correct. Once these are fixed, the data should be clean. As a more thorough test, we might test all of the rows after the sentinels are fixed. This issue is to complete the sentinel tests.

We're pretty sure the only problems left are issues with the offshore data, something to do with withheld data not parsing 100% correctly (ie, maybe we're leaving out a few rows of offshore).

gemfarmer:
Comprehensive update: 

The current federal production test script tests key, "sentinel", values for national, state, county, and offshore. Of these tests, here is what is passing ( ✅ ) , and pending ( ⚠️ ):

national: 
- ⚠️ pending until discrepancy in Oil values is resolved 

state: 
- ⚠️ pending until discrepancy in Oil values is resolved 

county: 
- ⚠️ pending until discrepancy in Oil values is resolved 

offshore: 
- ✅ it('only has Oil, Gas, Salt')
- ✅ it('checks offshore areas')


So it looks like the Oil values are the culprit. More soon.

gemfarmer:
This can also be closed!

PR #2009, opened 2016-09-20
Federal Production Data Sentinels

gemfarmer:
Fixes issue(s) #1994 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/data-fed-production)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/data-fed-production/)

- addresses `pending` data checks in federal-production.js

NOTE: it looks like the issue here was less the data tests, and more the sentinel values we received.

Specifically, we need to talk to Chris about
- [this value difference here](https://github.com/18F/doi-extractives-data/compare/dev...data-fed-production#diff-37dc4303b7491a4d4d9955117ebde1e7L162): this is the value for the Alaska state level rollup up for Oil, not a county-level check
- The values that we are checking for the national Oil rollups.
  - sentinel value Chris provided for Oil | 2013: 623103687.09
  - value in our data                                           : 623103681  ( difference of 6)
  - sentinel value Chris provided for Oil | 2015: 755158058.14
  - value in our data                                           : 755158065  ( difference of 7)
  - Because all of our production data in [data/federal-production/federal-production.tsv](https://github.com/18F/doi-extractives-data/blob/dev/data/federal-production/federal-production.tsv) is rounded, I think we are summing different values, and that this difference is very likely rounding error. We could be fine with it as is, or get a more up to date tsv file.

/cc @meiqimichelle

### DIFF
--- a/test/data/federal-production.js
+++ b/test/data/federal-production.js
@@ -35,7 +35,6 @@ describe('federal production (ONRR) sentinels', function() {
       );
     };
 
-    // [pending] until discrepency in Oil values is solved
     it('check sentinel values', function(done) {
 
       var sentinels = [
--- a/test/data/federal-production.js
+++ b/test/data/federal-production.js
@@ -94,8 +94,7 @@ describe('federal production (ONRR) sentinels', function() {
       );
     };
 
-    // [pending] until discrepency in Oil values is solved
-    xit('check sentinel values', function(done) {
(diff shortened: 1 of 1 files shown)


gemfarmer:
@meiqimichelle again, hound errors have been fixed.

Issue #1966, opened 2016-09-16
Glossary links in markdown front matter


coreycaitlin:
## Expected Behavior

It would be lovely to include some glossary links in the opt-in content.

## Actual Behavior

The syntax we use elsewhere — for example, `{{ "word" | term:"words" }}` — seems to break if we put it in the front-matter markdown sections, like so:

![screen shot 2016-09-15 at 7 56 19 pm](https://cloud.githubusercontent.com/assets/1450354/18574100/73551d06-7b7e-11e6-8694-7fd853a08bcc.png)

## Steps to reproduce the behavior

I found this while attempting to link the phrase "material acres" to its glossary item in the MT.md and WY.md files.

gemfarmer:
I agree that this would be pretty powerful. I think I _may_ have found our solution, but will need to test it. 

It looks like a person has created a `liquify` filter that will parse liquid! Check it out:
https://fettblog.eu/snippets/jekyll/liquid-in-frontmatter/

PR #2010, opened 2016-09-20
Allow glossary links and other valid liquid in frontmatter

gemfarmer:
Fixes issue(s) #1966 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/liquid-in-frontmatter/explore/WY/#state-production)

I'm pretty excited about this one! Basically, this PR allows us to use any valid liquid, including custom filters that we have made inside jekyll front matter. 

To make sure that the frontmatter is being read as liquid and not as markdown or as a plain value, we can do the following [(like I did here for the Wyoming state_land attribute)](https://github.com/18F/doi-extractives-data/compare/liquid-in-frontmatter?expand=1#diff-6ec0cd03783b6e398866fa7e25cdef33R116):

```
state_land: |
    The state government of Wyoming administers 3.4 million 
    surface acres of land (about 5.5% of the state) and 
    3.9 million {{ "mineral acres" | term_end }}. For detailed 
    information about land ownership in each county, see 
    the [Wyoming Statewide Parcel Viewer](http://gis.wyo.gov/parcels/).

...


{{ page.state_land | liquify | markdownify }}
```

NOTE: make sure to `liquify` and _then_ `markdownify`. If it is done in reverse order, markdownify will void the values within the `{{}}` as invalid markdown!

screen shot 2016-09-20 at 1 44 10 pm

/cc @coreycaitlin @shawnbot @meiqimichelle

### DIFF
--- a/_layouts/state-page.html
+++ b/_layouts/state-page.html
@@ -113,7 +113,7 @@ nav_items:
 
         {% if page.state_land %}
           <h3 id="state-production">Production on state land in {{ state_name }}</h3>
-          {{ page.state_land | markdownify }}
+          {{ page.state_land | liquify | markdownify }}
         {% endif %}
         {% if page.state_land_production %}
           {{ page.state_land_production | markdownify }}
--- a/_plugins/liquify.rb
+++ b/_plugins/liquify.rb
@@ -1,0 +1,7 @@
+module LiquidFilter
+  def liquify(input)
+    Liquid::Template.parse(input).render(@context)
+  end
+end
+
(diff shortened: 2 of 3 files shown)


meiqimichelle:
🎉 🎉 🎉 🎉 🎉

meiqimichelle:
I merged to get the code in, and I'm also pinging Corey so she sees the caveat

gemfarmer:
Sweet. This is small, but pretty powerful addition. Good idea, @coreycaitlin !

Issue #1988, opened 2016-09-19
Bug: State profile glossary links don't work


coreycaitlin:
## Expected Behavior

On the state pages, we're automagically creating glossary links for each unit, like so:

<img width="524" alt="screen shot 2016-09-19 at 6 53 54 am" src="https://cloud.githubusercontent.com/assets/1450354/18634440/b57122ca-7e35-11e6-85b5-db53c1487032.png">

## Actual Behavior

Even when we've defined the terms, the glossary links aren't working because of minor mismatches between the on-page term and the glossary term. In some cases, the terms may not need definitions. From my review so far, here are the ones not working, with the relevant defined terms in square brackets when we have them:

- barrels [barrel]
- megawatt hours [megawatt hour]
- kilowatt hours [kilowatt hour]
- tons [ton]
- million gallons
- hundred gallons
- million btus
- thousand pounds
- oz
- tons - equivalent
- tickets/pounds

## Steps to reproduce the behavior

Click on a glossary link in the production section on any state page.

coreycaitlin:
One option (ignore if this isn't the right way to solve this): add relevant glossary terms to `production_units.yml`, and only create the glossary link if a term exists for that unit (much like suffixes).

If that sounds like an okay way to handle this, I may be able to implement.

gemfarmer:
@coreycaitlin we could resolve those slight incongruences by doing this with the terms: `{{ "text" | term:"different text" }}` and add a `term` attribute to `production_units.yml`, like you are suggesting.

So we end up with something like this in `production_units.yml`:

```
barrels:
  long: barrels
  short: bbl
  term: barrel
```

and something like this in the template:

```
{% assign long_units = units_map[units].long | default: units %}
{% assign term_units = units_map[units].term | default: units %}

{{ long_units | term: term_units }}
```

meiqimichelle:
Done with #2014

PR #2014, opened 2016-09-21
Make production units work with glossary

gemfarmer:
Fixes issue(s) #1988 

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/glossary-mismatch/explore/)

- added a `term` attribute to `production_units.yml` where the term would be different than the displayed text. 
- Added logic in federal and all production for the state, national, and offshore levels

/cc @coreycaitlin

### DIFF
--- a/_data/production_units.yml
+++ b/_data/production_units.yml
@@ -5,47 +5,58 @@ mcf:
 tons:
   long: tons
   short: tons
+  term: ton
 short tons:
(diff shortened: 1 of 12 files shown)


meiqimichelle:
Checked with @coreycaitlin , she said 👍 💥 🐭

PR #2033, opened 2016-09-23
Glossary items, links, and general liquification

coreycaitlin:
Follow-up on issue #2010 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/content-fixes)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/content-fixes/)

- Implements the awesome fix from #2010 for all spots where we pull in YAML front matter as markdown (we won't use it most places, but this way no one has to memorize when it's okay to use glossary links)
- Replaces all relative URLs in front matter with `{{ site.baseurl }}`
- Adds at least one glossary link

/cc @gemfarmer 🎉 🎈 💯

### DIFF
--- a/_includes/location/opt-in/state-disbursements.html
+++ b/_includes/location/opt-in/state-disbursements.html
@@ -7,7 +7,7 @@
   {% assign revenue_total = funds.Total.All[revenue_year] %}
 
 {% if page.state_disbursements %}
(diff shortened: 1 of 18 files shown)


gemfarmer:
🎉 🎈 💯 ‼️

PR #2040, opened 2016-09-28
Increased vertical spacing for home-page map unit

ericronne:


ericronne:
It'd be great to reduce the bottom padding so that it matches the top on desktop, but this is a job for another day. (And for greater HTML brains)

Issue #2060, opened 2016-10-05
Handle error pages


gemfarmer:
Add CloudFront config to allow us to handle error pages:

[slack conversation, for reference](https://18f.slack.com/archives/18f-site/p1475685441000004)

PR #2062, opened 2016-10-05
Error pages 

gemfarmer:
Fixes issue(s) #2060 

[:sunglasses: PREVIEW 404](https://federalist.18f.gov/preview/18F/doi-extractives-data/error-pages/404/index.html)

[:sunglasses: PREVIEW 500](https://federalist.18f.gov/preview/18F/doi-extractives-data/error-pages/500/index.html)

- move error pages into pages directory. 
- This will not work as an error page on the federalist preview, but should work on the live site if CloudFront is properly set up

NOTE: If I am thinking about this right, we won't be able to detect if this works until it is in production :(

/cc @meiqimichelle @wslack


wslack:
Confirmed; though we know its working just like this at 18f.gsa.gov. As long as the URL works in preview, you should be good.

gemfarmer:
Thanks again, @wslack !

wslack:
Let me know when you verify it works!

Issue #2079, opened 2016-10-11
Green bars on megacharts bars have inconsistent left alignment


ericronne:
On some state megacharts, the green bar is indented from the left edge. This causes a thin strip of the light gray background to appear (only one pixel wide on some states, wider on others)

PR #2080, opened 2016-10-11
Remove padding on bar charts

gemfarmer:
Fixes issue(s) #2079

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/bar-padding/)

- removes offset property that was creating the margin-left and showing the gray pixels to the left of the bars
- @shawnbot do we need that offset anymore?

/cc @ericronne


ericronne:
Checked a bunch of states; 🐛 appears to be 👟 ed!

PR #2091, opened 2016-10-18
Deletes one-pixel rule from "Official" banner

ericronne:
Deleted the rule in order to make the style more consistent with the
overall site design


gemfarmer:
🎉

PR #2104, opened 2016-10-24
JS compiling hotfix

gemfarmer:
This commit, 009f8804ce5413a0d2a604e9938dba635c6c6222, built, but did not have all the minified and compiled javascript, so certain components are failing silently. This hotfix will bring them back to normal!


gemfarmer:
🍀

Issue #2023, opened 2016-09-22
Wrap up withheld product listing styles


shawnbot:
The withheld product listing should look like this, per @ericronne's design:

![](https://cloud.githubusercontent.com/assets/9259185/18717936/559c7f08-7fe7-11e6-86e8-4d0949881b8f.png)

The withheld icon work is in #2008.

ericronne:
bump ;)

shawnbot:
I have an update! I went through all of the states and found that none has more than three withheld products. This means that we can nix the "more" button from this design altogether. 🎉

ericronne:
💥

ericronne:
@shawnbot won't we still need this feature for the explore main / nationwide data page, where the withhelds are aggregated?

coreycaitlin:
After discussing, we think this design is still worth implementing and will help clarify withheld data.

jennmalcolm:
Next step is to test this with users to see if this benefit will be worth the effort required to build.

jennmalcolm:
Testing with one use but determined that we need to have it interactive, so we should just build it in the live site to try out.

mcharg:
For research findings see: https://github.com/ONRR/doi-extractives-data/blob/research/research/researchfindingsbytopic.md#withheld-data

mcharg:
@jennmalcolm We can close this, if withheld data is going away, correct?

mcharg:
Probably won't be relevant with change in withheld strategy and changes to Explore Data page.

Issue #2008, opened 2016-09-20
Add icon to bar charts to indicate withheld & format wh product listing


ericronne:
On bar charts, grayed-out years indicate a lack of data. However, you have to hover over or select the year in order to know whether or not that lack is because the number is withheld or not. So it might be helpful to drop a marker into the empty bar space for those years, like so...

![image](https://cloud.githubusercontent.com/assets/9259185/18682884/959a2696-7f33-11e6-9c6d-ae9531d75476.png)

ericronne:
![image](https://cloud.githubusercontent.com/assets/9259185/18688168/c212605c-7f48-11e6-80fb-2cae2cd813fe.png)

[**Default icon**] (https://drive.google.com/open?id=0BwWYNZcEDfwaWEdMYWJySkRFNmM)

[**Selected icon**] (https://drive.google.com/open?id=0BwWYNZcEDfwaTjhGNnh0bUF6SkE)

ericronne:
For consistency with other bar hovers, let's make the hover/selected color #587F4C (if we do implement this)

ericronne:
## Design...

![image](https://cloud.githubusercontent.com/assets/9259185/18759674/2297ea2a-80c3-11e6-985a-763e5b2ef66b.png)

coreycaitlin:
Closing in favor of #2023

PR #2126, opened 2016-11-01
Re-style withheld list

shawnbot:
Fixes issue(s) #2023.

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/withheld-list)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/withheld-list/)

Before:

image

After:

image

- Re-styles the withheld product list under federal production, e.g. in [Utah](https://federalist.18f.gov/preview/18F/doi-extractives-data/withheld-list/explore/UT/)
  - [x] Turn the section into a chart-like block with similar title
  - [x] Remove bullets
  - [ ] Normalize text size
  - [x] ~~Style the button with the plus-in-circle icon, no background color~~ (no state has more than three withheld products)

@gemfarmer @ericronne: I'm not sure what the text size for this should be, but 0.875rem appears to be the next size down from 1. I'm also not sure what the "right" way is to reduce either the line-height or the amount of whitespace between the list items. 😬


ericronne:
@shawnbot i'm having trouble finding a page with an example of this; could you link me?
thx!

gemfarmer:
@shawnbot is this still a WIP?

ericronne:
@shawnbot here's the formatting i had. I figured it would only occupy one column, and slot in as the last chart of the group.

image

I had also dropped in the W icon parenthetically, which we should only do if there's time to apply that to charts that have withheld years (issue #2008).

image
image

thx!

ericronne:
@shawnbot is it possible to reduce the width down to match the bar charts, as opposed to running the full width of the page? It's fine if there's neg space to the right.

gemfarmer:
@shawnbot I like the list styling way more! Are we supposed to be seeing the 'W' in the bars? https://cloud.githubusercontent.com/assets/9259185/20358942/6e178d88-abf3-11e6-88b5-d50a8b786cd1.png

gemfarmer:
@shawnbot looks like these three lines are redundant!

gemfarmer:
Well withheld icons or not, looks great! I vote that we merge once the Jekyll error goes away.

ericronne:
What about making it a single column?
(We can deal with #2008 in a sep PR, time permitting.)

js/components/open-list-nav.js
```diff
@@ -126,11 +126,13 @@
 
             if (!document.getElementById(item.dataset.navItem) && item) {
               item.setAttribute('aria-hidden', true);
+              item.outerHTML = '';
+              delete item;
+            } else {
+              item.addEventListener('click', function () {
```
dmsnell: Seriously @shawnbot - why would you do that! Never…ever…make functions within a loop.

:wink: 🙃

Issue #2127, opened 2016-11-02
Dividing rules and hover fill don't go full width on state-pg tables (Chrome)


ericronne:
## Expected Behavior
One-pixel dividers and hover fill should stretch the full width of tables
![image](https://cloud.githubusercontent.com/assets/9259185/19929710/ba82deec-a0d2-11e6-942a-c77753311f56.png)

## Actual Behavior
![image](https://cloud.githubusercontent.com/assets/9259185/19929704/a9027ccc-a0d2-11e6-8bb1-1d91b17bf1ef.png)

## Steps to reproduce the behavior
Noticed issue in Chrome; looks fine in Safari

PR #2142, opened 2016-11-16
Fix table bug

gemfarmer:
Fixes issue(s) #2127 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/tables/tables-bug)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/tables-bug/)

This PR addresses puzzling block of white behind the bars in the scrollable table menu (Chrome-specific). It appears that this kind of quirkiness was being caused by the hoops that were jumped through to make the table scrollable. This was resulting in some weird border business and off positioning.

/cc @ericronne

### DIFF
--- a/_includes/location/display-federal-production-county.html
+++ b/_includes/location/display-federal-production-county.html
@@ -6,6 +6,7 @@
   {% assign units = include.long_units %}
 {% endif %}
 
+
 <table is='bar-chart-table'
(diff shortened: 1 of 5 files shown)


gemfarmer:
@ericronne lololololol biggest headache ever over a one line change.

PR #2215, opened 2017-03-01
[WIP] Stylefunzies

ericronne:
Makes skinny (h2) home header fat (h1)
Emboldens the footer subheads (Questions et al)
Emboldens the font in the base button style

**More to come!**

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/stylefunzies/)

### DIFF
--- a/_sass/_fonts.scss
+++ b/_sass/_fonts.scss
@@ -75,6 +75,10 @@
   padding-right: $base-padding-lite;
 }
 
+.icon-padded_after:before {
+  padding-left: $base-padding-lite;
+}
+
 $icon-gold: '\0041';
 $icon-close-x: '\0042';
 $icon-renewables: '\0043';
--- a/_sass/components/_buttons.scss
+++ b/_sass/components/_buttons.scss
@@ -23,8 +23,10 @@ input[type='submit'] {
 
   background-color: $primary-button-bg-color;
   color: $white;
-  padding-left: 0.7em;
-  padding-right: 0.7em;
+  padding-top: 0.7em;
+  padding-bottom: 0.7em;
+  padding-left: 2em;
+  padding-right: 2em;
   -webkit-font-smoothing: antialiased;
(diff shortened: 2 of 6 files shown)


ericronne:
@gemfarmer or @meiqimichelle i'd love to pair for a few minutes to suss out very minor footer formatting. Thx!

_sass/components/_buttons.scss
```diff
@@ -23,9 +23,11 @@ input[type='submit'] {
 
   background-color: $primary-button-bg-color;
   color: $white;
-  padding-left: 0.7em;
-  padding-right: 0.7em;
   -webkit-font-smoothing: antialiased;
+  padding-top: 0.7em;
```
ericronne: fixed

_sass/_fonts.scss
```diff
@@ -75,6 +75,10 @@
   padding-right: $base-padding-lite;
 }
 
+.icon-padded_after:before {
```
meiqimichelle: This file comes mostly from an external source. #wontfix

meiqimichelle:
@ericronne Take a look when you have a chance. Perhaps good to go?

_sass/_fonts.scss
gemfarmer: I think Hound is trying to make a distinction between `pseudo elements` and `pseudo classes`. I have the distinction, but until now, I never took the time to investigate. This is what I discovered(!):

From https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements

> Sometimes you will see double colons (::) instead of just one (:). This is part of CSS3 and an attempt to distinguish between pseudo-classes and pseudo-elements. Most browsers support both values.

`pseudo elements` like before and after are more/less elements
 `pseudo classes` are essentially metadata applied to an element `:first-of-type`

Issue #2218, opened 2017-03-03
Clean the tables  (ノಠ益ಠ)ノ彡┻━┻


ericronne:
Do a bit of restyling and consistency-izing.
Likely via @meiqimichelle and i 🍐-ing.  

## Tweaked megachart reference …
![image](https://cloud.githubusercontent.com/assets/9259185/23569113/52dca240-0023-11e7-92c8-045b653b0a2d.png)

meiqimichelle:
Done with #2222

PR #2222, opened 2017-03-07
Megachart table cleanup

meiqimichelle:
For #2218 .

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/table-cleanup)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/table-cleanup/explore/#revenue)

- Right-aligns values
- Inserts break after split-bar names so they don't run into each other anymore (!)
-

/cc @ericronne

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 _site
 *.db
 node_modules
+tmp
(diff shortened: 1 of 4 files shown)


ericronne:
This is probably the only remaining funkhood …

image

If it's hard to fix, i'm fine merging (so long as everything else seems ok)!

PR #2227, opened 2017-03-14
New card images woo, also changed a link and emboldened small headers

ericronne:
[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/card-agains/)


gemfarmer:
Holy smokes @ericronne!

ericronne:
Trivia: replaced PNGs with SVGs, which saves 345k and looks way crispy.

gemfarmer:
Jiminy crispets those SVGs look good. My only input would be that we should probably also delete the old card images.

ericronne:
I believe that i *did* delete them!

PR #2231, opened 2017-03-15
Official Website banner: simplified / Global header changes

ericronne:
[Click and behold.](https://federalist.18f.gov/preview/18F/doi-extractives-data/map-color-nothingclever/)

NOTE: branch name is incorrect (in addition to being unclever). No map changes were made.

### DIFF
--- a/_sass/blocks/_header.scss
+++ b/_sass/blocks/_header.scss
@@ -113,10 +113,10 @@ header,
 .header-nav_item,
 .header-nav_item_top {
   display: inline-block;
+  font-weight: $weight-light;
   padding-bottom: $base-padding-lite;
   padding-right: $base-padding;
   padding-top: 0;
-  font-weight: $weight-light;
 }
 
 .header-nav_item_top {
--- a/_sass/blocks/_header.scss
+++ b/_sass/blocks/_header.scss
@@ -47,7 +47,8 @@ header,
 }
 
 .header-left {
-  @include span-columns(6);
+   @include span-columns(6);
+   margin-bottom: $base-padding;
 
    @include respond-to(medium-up) {
       @include span-columns(3);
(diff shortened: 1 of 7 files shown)


ericronne:
Changes were all to the banner:
 - Removed the feedback and what's new links from the banner. They appeared to be suffering from "banner blindness," and thus received very few clicks. "What's new" is a high-profile link atop the About main page, and the feedback link is still available via a large button in the footer.
- Centered the Official notice and made it slightly larger, bold
- Slab color change

_sass/blocks/_header.scss
```diff
@@ -53,6 +52,10 @@ header,
   @include respond-to(medium-up) {
     @include span-columns(3);
   }
+
+  @include respond-to(medium-down) {
+    margin-bottom: $base-padding;
```
gemfarmer: @ericronne you could probably do something like this instead:

```
.header-left {
   @include span-columns(6);
   margin-bottom: $base-padding;

   @include respond-to(medium-up) {
      @include span-columns(3);
      margin-bottom: 0;
   }
}

File: _sass/blocks/_header.scss [hunk 2]
```diff
@@ -106,16 +109,17 @@ header,
   margin-bottom: 0;
 
   @include respond-to(huge-plus-up) {
-    margin-bottom: 1em;
+    margin-bottom: 2em;
```
gemfarmer: Would it be possible to use `$base-padding` instead of `2em`? That is `1.25em`

gemfarmer: Trying to stick to variables as much as possible

gemfarmer:
@ericronne It appears that the banner disappears in mobile ☹️ 
screen shot 2017-03-29 at 4 53 16 pm

ericronne:
The gov bit vanishes on the live site, too …
image

I'll see if i can fix; it should be there.

PR #2234, opened 2017-03-27
Increased contrast of map greens

ericronne:
Did a quick-and-dirty opacity adjustment on the black (darkening) layer. At some point, it would be good to rebuild this so colors could be spec’d individually, with hex values. This would allow for greater flexibility, and enable us to create maps and other visualizations with more than three colors.

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -154,8 +154,8 @@ $form-border-radius: $base-border-radius;
 $form-box-shadow-focus: 0 0 0 2px $form-border-color-hover;
 
 // Map legend
-$opacity-federal: 0.35;
-$opacity-state: 0.2;
+$opacity-federal: 0.45;
+$opacity-state: 0.25;
 $legend-row-height: 1em;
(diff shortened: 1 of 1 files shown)


ericronne:
fjss5nqbbu

gemfarmer:
I think the contrast looks better. When reviewing this I noticed that the colors that we are using for the bars and the selectors on the explore pages don't match any of the map colors. I thought we had it matching the middle color, but the preview at `dev` doesn't even do that. Do you remember when we strayed @ericronne ?

ericronne:
I don't remember that being a rule, though maybe it was! The colors work, though. Was there a technical reason to match them?

gemfarmer:
No technical reason. I just thought that maybe having fewer colors would be better. Your domain though! I like that green, so I'm not bummed.

ericronne:
Yep i hear you. Max contrast was the goal of this change, which doesn't necessarily yield the prettiest colors.

Issue #2228, opened 2017-03-14
Consider rewriting SEVERAL home-page-card headlines, for better wrapping and clarity


ericronne:
### The current header text (A below) for the disbursements card on the home page breaks awkwardly. Perhaps we should rewrite?

**Here are some options, with a Hemingway pic for inspiration:

![image](https://cloud.githubusercontent.com/assets/9259185/23918884/b4b1d2e0-08c2-11e7-9ebd-fd968f471ae0.png)

gemfarmer:
We can also also change the mininum height on those titles and then the content would start at the same spot.

ericronne:
True, true, but the uniform heights make for nice gridding!

In fact, I just got inspired to make **everything** line up nicer, by ensuring that each of these headlines fit on exactly two lines on desktop. That meant editing the third and fifth card, too.

## Current text   >   Suggested edit
Where does the money go?   >   **How is the money from extractions spent?**
How do extractive industries affect communities?   >   **How do these industries affect communities?**
How does resource extraction result in federal revenue?   >   **How does extraction result in federal revenue?**

Here's how that lays out …
![image](https://cloud.githubusercontent.com/assets/9259185/24254189/42070716-0fb0-11e7-8cf8-360f68c1cd36.png)

cc @meiqimichelle

ericronne:
Here's a rewrite of card 2's headline, which eliminates one of the How's:
"Where does the money from extraction go?"

@meiqimichelle either this or my previously posted one do the trick for you?

![image](https://cloud.githubusercontent.com/assets/9259185/24315194/aecb487c-10b3-11e7-817c-8088b8e9968c.png)

PR #2235, opened 2017-03-27
Edited three card titles so all six now fit comfortably on 2 lines on desktop

ericronne:
For #2228

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/card-header-text-changes/)
(scroll to see cards, natch)

### DIFF
--- a/index.html
+++ b/index.html
@@ -86,7 +86,7 @@ selector: location
 
           <p class="card-content-subhead">Explore data</p>
 
-          <a href="{{ site.baseurl }}/explore/#federal-disbursements"><h3 class="card-title">Where does the money go?</h3></a>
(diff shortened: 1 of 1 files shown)


ericronne:
IMO i didn't compromise as you claim, @meiqimichelle!

**Where does the money from extraction go?** is more explicit than the current version, for example. I've basically swapped specificity between that card and the next. When scanned together, the titles seem clear to me, but let's chat. xo

meiqimichelle:
@ericronne if I recall, we have a practice of not using the word 
"extraction" alone, unmodified. ie, it should always be used together with "resource extraction" or "extractive industries". Likewise, "these industries" doesn't necessarily stand alone.

If we'd like more specificity in that particular sentence, it should be, then, *Where does the money from resource extraction go?*.

ericronne:
stet for now

Issue #1638, opened 2016-07-21
Update styleguide


meiqimichelle:
- [ ] @gemfarmer should update KSS in the our Sass files so that our styleguide accurate reflects the current styles in use on the site
- [ ] @gemfarmer to pair with @ericronne to make sure the colors and typography etc are up to date and correct
- [ ] @gemfarmer and @shawnbot to add component suite

gemfarmer:
Use [Fractal](http://fractal.build/guide) for component library??

shawnbot:
@gemfarmer possibly! Let's try incorporating the components into the KSS output first, though.

ericronne:
I'm also updating the overall palette right now, with an eye to simplifying and moving the look in a more subtle direction, color-wise.

ericronne:
New grays! Let's try swapping these in, using my chart below as a guide. I've reduced the overall count of grays we're using from 11 down to 8. I've also added two new grays, just for snicks. I don't think we need them right now.

 - [ ] I need to update a couple of graphics as a result of these changes

 The ribbons should all be colored as follows...
![image](https://cloud.githubusercontent.com/assets/9259185/18091575/be1f1ffa-6e8e-11e6-81b8-abce0c1d771d.png)

Ping with q's and comments, thx!

gemfarmer:
This issue should/could also update the print stylesheet, which has fallen to the wayside

ericronne:
New grays!...

![image](https://cloud.githubusercontent.com/assets/9259185/18211441/29ab72de-7103-11e6-944b-f3b40e8e043d.png)

ericronne:
Blues! (mostly for map legends, and we really need to see these in action to judge)...

![image](https://cloud.githubusercontent.com/assets/9259185/18211807/32555524-7105-11e6-9daa-5ad1fd4584d1.png)

coreycaitlin:
We should still update the styleguide at some point, but the comments here probably don't represent the latest typography/colors.

ericronne:
The blues haven't yet been fully implemented, but yep, probably better to hit the reset button on this issue and circle back later.

ericronne:
(second thought, leaving this open as a reminder)

ericronne:
OBE (covering in #2360)

PR #2236, opened 2017-03-27
Brings link/button blue into 508 compliance over $pale-gray

ericronne:
Barely noticeable to the naked eye, but here's [what it looks like](https://federalist.18f.gov/preview/18F/doi-extractives-data/springtime-palette-tweak--508/).


ericronne:
image

ericronne:
Noted in #1638. ...
image

Issue #1900, opened 2016-09-12
Megachart: design for negative numbers


ericronne:
I'm hesitant to run negative numbers in a new color. Red (eg) would communicate well, but it would add another color to the palette, and also draw inordinate attention to itself. A simpler solution is to italicize negative numbers, and not visually represent the "amount" at all. Instead, we could simply outline a blank rectangle with one pixel of $gray-light (#d3dfe6), like so …

![image](https://cloud.githubusercontent.com/assets/9259185/18451398/5561c1c0-78fc-11e6-8c0b-ad7732b500a4.png)

thoughts @gemfarmer @meiqimichelle @coreycaitlin?

meiqimichelle:
_like_

ericronne:
Changed this task to "implement"!

gemfarmer:
Closed with #2238

PR #2238, opened 2017-03-30
Megachart negatives numbers

gemfarmer:
Fixes issue(s) #1900 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/negatives)

[:sunglasses: PREVIEW](https://federalist.18f.gov/preview/18F/doi-extractives-data/negatives/explore/#federal-revenue)

- Added `intcomma_dollar` filter to make it so that negative numbers would read `-$XXX` instead of `$-XXX` 
- Changes style in accordance with @ericronne's mocks. I did not add italics because we don't have a Lato light italic font and because of that it looks _weird_:

screen shot 2017-03-29 at 5 21 02 pm

/cc @meiqimichelle

### DIFF
--- a/_how-it-works/laws-and-regulations/tribal-laws-and-regulations.md
+++ b/_how-it-works/laws-and-regulations/tribal-laws-and-regulations.md
@@ -177,7 +177,7 @@ In the tables below, revenue may be grouped differently depending on the stage o
       {% for revenue_type in page.revenue_types %}
         {% assign _revenue_type = revenue_type[0] %}
         {% assign _revenue = commodity[1][_revenue_type] %}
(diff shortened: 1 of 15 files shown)


ericronne:
Feels a little subtle to me. We could use parens, or change the text to red. Either of those is likely to stand out well. Red is the more blatant choice (except maybe for those with red-green color blindness).

image

Comes down to how much emphasis should be placed on these numbers.

gemfarmer:
@ericronne I like the idea of `($41,000)` more than a red - prefixed number. That seems pretty standard, now that I think about it.

ericronne:
Indeedy. Parens it is! 👍

PR #2251, opened 2017-04-07
Darkened county fills on case-study — and state laws — maps

ericronne:
Brings case-study ***and state laws*** maps color into line with palette. We ~should~ DID change the light gray outlines as well, but this one is the more blatant offender.

### [Case study example](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/svg-map-green-de-uglify/case-studies/boone-logan-and-mingo/)


ericronne:
@gemfarmer checkit! 🤞

PR #2285, opened 2017-05-02
Links "How does resource extraction result in federal revenue?" home-pg card to #process hash on how page

ericronne:
[**Click here and scroll to the truck card, then click. Vroom vroom!**](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/home-card-link-targeting)

This card currently links to the top of the How landing page. I’ve changed it to link to the section which i think answers the posed question more directly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -134,7 +134,7 @@ selector: location
 
           <p class="card-content-subhead">How it works</p>
 
-          <a href="{{ site.baseurl }}/how-it-works"><h3 class="card-title">How does resource extraction result in federal revenue?</h3></a>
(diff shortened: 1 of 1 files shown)


gemfarmer:
[**Should this go to `#revenues` instead?**](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/home-card-link-targeting/how-it-works#revenues)

ericronne:
I think process is the better option. Revenue is too thick with information, whereas #process explains how resources eventually end up as revenue.

gemfarmer:
up to you @coreycaitlin !

PR #2294, opened 2017-05-09
[WIP] Roboto? Plays much nicer than Lato in Safari.

ericronne:
Don't merge!

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/font-weight-n-see/)

Safari hates Lato, rendering the light weight extremely thinly. There seems to be no straightforward way to make it work, so i'm playing with other type options. One constraint is tabular figures, which narrows the field considerably. Roboto has tabular figs, and reads well at small sizes. Will let it sit in WIP form for a bit, to see how we like it....


ericronne:
indeed! thx

ericronne:
@shawnbot do we need to download a css file, since we're using [google fonts](https://fonts.google.com/selection?selection.family=Roboto)?

ericronne:
fancy. thx onnit!

_sass/fonts/_all.scss
```diff
@@ -1,5 +1,5 @@
 // include the CSS files with relative @font-face declarations
-@import url(fonts/lato/lato.css);
+@import 'Roboto';
```
shawnbot: This needs to be:

```scss
@import 'roboto';
```

I.e., lowercase `r`. I think this works on your computer because OS X's filesystem is case-insensitive, whereas the FS on Circle's Linux build containers is not.

coreycaitlin:
@ericronne Let's check in about this one later this week!

ericronne:
houdza bee-atch

ericronne:
Due partly to its tabular numerals, I'm recommending that we stick with Lato (#2830), and eliminate the light weight.

PR #2302, opened 2017-05-17
Case study map embed sizing

ericronne:
[ 😎  **Preview**](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/case-study-map-embed-sizing/case-studies/boone-logan-and-mingo/)

Created "wide" and "tall" map classes for these, to improve relative sizing.


ericronne:
Went out on a cray css limb, yipes!
@gemfarmer we can review together in pairing if you like, thx!

coreycaitlin:
👍  I think this looks good, but it looks like we're getting a conflict now.

Issue #2345, opened 2017-08-15
Update Federal Production data


coreycaitlin:
We have federal production data for FY and CY 2007-2016 (limited access: [spreadsheets here](https://drive.google.com/drive/folders/0B4b900J0no9nRkQzMkZIQmJBbHM)).

This issues is done when:

- [x] The latest dataset is pulled into the site
- [x] New data is tested
- [x] New data is verified by @mentastc

winzenburg:
Did our friend "brine barrels" get eliminated?

coreycaitlin:
@winzenburg Not totally sure what you mean! Could you add a bit more detail?

coreycaitlin:
@winzenburg I figured it out! Yes, it looks like 2006 was the only year for which we had _non-withheld_ `brine barrel` production data. With #2386, this section is now showing data for 2007-2016, so there is no graph for brine barrel production on federal land.

If you expand the "Data withheld" section below, it notes that production data for brine barrel production on federal land was withheld for 2007, 2008, and 2010.

coreycaitlin:
Closing this out, because we've merged this data into dev, and it's been reviewed by @mentastc. Will reopen if we spot any data issues as we push to production.

PR #1990, opened 2016-09-19
Year default consistency

gemfarmer:
Fixes issue(s) #1925 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/year-consistency)

[:sunglasses: PREVIEW AZ 🌵](https://federalist.18f.gov/preview/18F/doi-extractives-data/year-consistency/AZ/#federal-production)

[:sunglasses: PREVIEW CO 🗻](https://federalist.18f.gov/preview/18F/doi-extractives-data/year-consistency/AZ/#federal-production)

Changes proposed in this pull request:
- makes it so that years in the county (and offshore area) maps and charts default to the same year as the corresponding bar chart

/cc @meiqimichelle

PR #2349, opened 2017-08-24
Update federal production data

gemfarmer:
Fixes issue(s) #2345 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/update-federal-production/)

[:sunglasses: PREVIEWS](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/)

- [National federal production](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/explore/CO/#federal-production)
- [CO fed production](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/explore/CO/#federal-production)
- [AK fed production](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/explore/AK/#federal-production)
- [Data download page](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/downloads/federal-production/)
- [Downloads summary page](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update-federal-production/downloads/#federal-lands-and-waters)

- [x] Adds new federal production data
- [x] Updates new sentinel data for data tests
- [x] Update years.yml config so that the bar charts show the correct years
- [x] Update Downloads with new Excel files
- [x] Update wiki to reflect these changes

/cc @coreycaitlin @winzenburg @toolness


coreycaitlin:
@mentastc This could use your review of the data! (See preview links above)

mentastc:
Noticed an issue, the bar chart highlights 2016 but the map data is 2015. Could we get them to both default to 2016?
image

gemfarmer:
@mentastc good catch! And definitely. 

@toolness want to debug this with me at some point this week?

mentastc:
I'm on vacation after today, but I have some time today

coreycaitlin:
@toolness Any chance you can take a look at this today?

coreycaitlin:
@toolness I think this is our highest priority right now — as soon as we get this fixed, we can push this new dataset live. Happy to pair on it if that helps!

toolness:
Looking at this now, sorry for the delay!

toolness:
I believe the bug may be occurring due to the way that #1990 is implemented.  In particular, that PR added a line in [`eiti-data-map.js`](https://github.com/18F/doi-extractives-data/pull/1990/files#diff-5efbdcea04d96182d0804ce4483fe460):

```js
          if (init) {
            this.setYear('2015');
          }
```

This seems to be the line that is causing the "2016" in the map to change to "2015" on page load (yes, the original HTML markup delivered to the browser, which can be seen by disabling a browser's JS, actually mentions 2016, not 2015).  That is, when I change this line to "2016", the map is synced with the bar chart.

However, my fear in actually committing this change is that it may be appropriate for some maps to actually show "2015": before he departed, Brian told me that some data sets in EITI go through 2015, while others go through 2016.

It seems like a more robust long-term fix would be for the year in the map to flow from some higher-level component that passes the current year to both the map and the bar chart, but I'm too unfamiliar with the code to implement that kind of solution easily. Hmmmmm.

toolness:
Ok, I *think* I've fixed this problem with the least possibility of accidentally creating additional regressions by attempting to get the year from the existing HTML that was written into the page by Jekyll, and falling back to the hard-coded `'2015'` if that fails.

@mentastc & @coreycaitlin, can you check the federalist build to see if this fixed things, while also hopefully not breaking other things? 😅

js/components/eiti-data-map.js
```diff
@@ -308,7 +313,9 @@
 
 
           if (init) {
-            this.setYear('2015');
+            // If a year has already been included in the HTML delivered
+            // the browser, use that; otherwise, enforce a default year.
+            this.setYear(this.getYear() || '2015');
```
gemfarmer: Nice! It probably should have been like this to begin with

gemfarmer:
@toolness looks fixed to me!

toolness:
Oh cool, thanks for taking a look @gemfarmer !

PR #2351, opened 2017-08-28
Add optional docker support.

toolness:
This attempts to add optional support for developing with Docker, which potentially reduces the barrier to entry, as people only need to install Docker rather than the proper version of node, ruby, and other dependencies.

Note that unlike the other config files, the Docker setup uses Node 6, in anticipation of #2357.

## How to try out this PR

Follow the instructions I've added to `CONTRIBUTING.md`! If you're confused by anything, others will be too, so we should fix those instructions as needed.

## To do

- [x] Ensure that Jekyll doesn't copy docker-related files into `_site`.
- [x] Add documentation.
- [x] ~~Add support for developing the style guide.~~ Spun this out into #2355.
- [x] Add support for working on the site data.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -641,6 +641,7 @@ $(db): \
 	tables/federal_production \
 	tables/all_production \
 	tables/company_revenue \
+	tables/tribal_revenue \
 	tables/jobs \
 	tables/gdp \
 	tables/exports \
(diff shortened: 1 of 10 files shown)


toolness:
Ok @gemfarmer, I think this is ready for review!

gemfarmer:
Thanks @toolness! I'll take a look 👀

Makefile
```diff
@@ -686,7 +693,7 @@ tables/reconciliation: data/reconciliation/output
 	tmp=$^/all.ndjson; \
 	for revenue_filename in $^/????.tsv; do \
 		$(tito) -r tsv --map ./data/reconciliation/transform.js \
-			$$revenue_filename >> $$tmp; \
+			$$revenue_filename $(append_to_tmp); \
```
gemfarmer: Nice!

Issue #2353, opened 2017-08-29
tribal_revenue table isn't built when 'make db' is run


toolness:
I noticed when running `make db`, the `tribal_revenue` table isn't built, because the `tables/tribal_revenue` rule in the Makefile doesn't seem to be a dependency of the `db` rule. Is this intentional? It could be added to the `db` rule pretty easily, I'm happy to submit a PR.

PR #2358, opened 2017-08-31
Add makefile test script

toolness:
**Note:** This is a PR against #2351, not `dev`, since that PR fixes things that make this test actually pass.

This adds a script called `test-makefile.sh` that just makes a copy of `_data`, then re-runs `make db` and `make site-data`, and ensures there are no differences between the new `_data` and its original/copy. It's the kind of thing that would have caught the regression in #2353, ensures that `data` is always in sync with `_data`, and could also be useful whenever we decide to refactor the Makefile or any of the scripts/tools it uses.

@gemfarmer does this seem like it'd be useful?

## To do

- [ ] Document this in the "tests" section of `CONTRIBUTING.md`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,3 +22,4 @@ test:
     # - npm run test-html
     # - npm run test-ruby
     - npm run test
+    - bash test-makefile.sh
--- a/test-makefile.sh
+++ b/test-makefile.sh
@@ -1,0 +1,34 @@
+#! /bin/bash
+
+set -e
+
+export DIR=_data
+export ORIG_DIR=.original-_data
+
+echo "Copying ${DIR} to ${ORIG_DIR}..."
+
+rm -rf ${ORIG_DIR}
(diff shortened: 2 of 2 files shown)


gemfarmer:
I haven't run it, but it seems like an improvement over running `git diff` after running `make db && make site-data`?? It would be cool if it was non-destructive so that you could run it in the middle of your workflow and essentially see a diff from whatever the current state of `_data` is.

gemfarmer:
So after looking at this a bit more it looks like it is doing exactly what I was suggesting :man_facepalming: . So this is great! 

One thing that could be interesting in the future to allow the test to only run a certain set of commands? I'm imagining being able to update the federal production table and running something like `bash test-makefile.sh federal_production` and having it build the federal production tables and data without wasting time building other parts of the data file. This seems like it could be tricky and what you have here already works well.

PR #2374, opened 2017-09-13
Moved year selector up to sticky area, ripple-effect tweaks

ericronne:
[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/stick-shift/explore/)

I introduced at least one bug here (see below), and also created a bit of “cruft,” as it were, but hopefully not too bad :)

---

### Before
image

---

### After
image

---

### Glitch which seems to be new (in at least this texas "no data" selector)
image

### DIFF
--- a/_includes/location/national-all-production.html
+++ b/_includes/location/national-all-production.html
@@ -17,15 +17,16 @@
 
   <div class="chart-selector-wrapper">
 
-    {% include year-selector.html year_range=year_range %}
-
-    <p class="chart-description">
-      The Energy Information Administration collects data about energy-related natural resources produced on federal, state, and privately owned lands and waters. This data does not include information about nonenergy minerals.
-      <br>
-      <a href="{{site.baseurl}}/downloads/#all-lands-and-waters">
-        <icon class="fa fa-file-text-o u-padding-right"></icon>Data and documentation
-      </a>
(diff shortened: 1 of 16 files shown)


toolness:
Nice, I like this change!

I'm trying to figure out why the glitch is happening, and I'm noticing that on the existing production site, even though that glitch isn't appearing, the HTML being delivered to the browser actually *does* have the markup that the glitch shows in your PR:

```html
<select class="chart-selector flex-row-icon">
  <!-- empty years: ["2016"] -->
    <option disabled value="2016" selected>2016 no data</option>
```

Um, also, I just noticed that the glitch *does* actually appear in the production site, but only when the "Wage and salary jobs" header is sticky at the top of the page:

> 2016-no-data

So I don't think your PR so much *creates* a glitch as much as makes an existing one a lot easier to notice...

ericronne:
Soooooo .... merge 'n' see? :)

toolness:
Heh, I filed it separately as #2378 and I think it can be tackled separately from this PR... So feel free to merge this if you think that the higher visibility of the glitch isn't that big a deal!

ericronne:
Since we aren't introducing a bug with this branch … merging!

PR #2380, opened 2017-09-22
Font weight comparison: retains Lato but thickens "light" to "book" weight (Plus other typographic adjs!)

ericronne:
[ WIP don’t merge ]

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/font-weight-door-number-three/)

Compare to [this branch](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/font-weight-n-see/) and [this branch](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/font-weight-comparison/)

Resolves #875.

Some ripple-effect consequences are created by this, but they should be easy to resolve.

PR #2406, opened 2017-10-26
Lato thickened!

### LINKED ISSUE #875: Visual Bug: Lato light appears too light on Safari
I think @msecret may have dealt with this issue on Peace Corps.

Also may be light on other browsers, but i only tested on safari. (Fine on chrome)

ericronne:
Bottom line: safari hates Lato 200, but has no problem rendering Lato
300.

Resolves #875

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -4,9 +4,9 @@
 $eiti-logo-width: 195px;
 
 // Typography
-$weight-light: 200; // use
-$weight-book: 400; // use
-$weight-bold: 700; // use
+$weight-light: 300; // (Lato Light) use
+$weight-book: 400; // (Lato Regular) use
+$weight-bold: 600; // (Lato Semibold) use
 $line-height-caps: 1.2;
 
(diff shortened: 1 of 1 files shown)


ericronne:
I've split this change out from the other misc adjustments i've been playing with over in PR #2380. Perhaps we can pair on those changes later @msecret :)

coreycaitlin:
[PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/lato-weight-adjustment-here-goes/)

I'm not seeing a difference visually, but that might be more about the limitations of my eyes than anything else. Approving because I think this is definitely safe to merge — all looks normal to me in Chrome, Safari, and Firefox, at least.

ericronne:
looks as it should! :)
solves the safari problem.
keeping the other pr as a sandbox, for next week...

msecret:
I'm ok with merging this for now, but we should remove any weights we're not using (like 200 right?) in the future

PR #2407, opened 2017-10-26
Tweak user groups

coreycaitlin:
This research plan looks great — this PR suggests some revisions to the participants section, just to bring the list of user groups slightly more in line with past persona conversations. (I also reduced the target number to 5-8 interviews, because of time.)

/cc @alexpandel @winzenburg


coreycaitlin:
Don't worry about Circle — this branch doesn't need to build!

PR #2423, opened 2017-11-01
Name change footer, formatted

ericronne:
[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/name-change-footer/)


msecret:
Why don't we merge this into `name-change` branch?

PR #2451, opened 2017-11-15
Provides easy way to update EIP data from docker

msecret:
While updating the EIP data, I found I had to run the update commands from outside of docker as there was no registered command to run it. In this PR, I'm adding a shortcut to run the command in the Makefile for two reasons:

- Allow the command to easily be run from docker, as easy as making site-data.
- Provide more transparency about the EIP data update and even that this separate update script exists, as I had no idea where to find it previously.

So now the command to update the EIP data would be to ensure you have the API key exported into the current env and then run: `docker-compose run app make eip-data`.

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/ms-easy_eip_data_command)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/ms-easy_eip_data_command/)


jseppi:
I did a rebuild that worked after a few attempts that didn't: https://federalist-staging.18f.gov/sites/42/builds/3247/logs

As mentioned in Slack, the initial reason federalist-staging builds were sometimes failing was because we had the memory on the build container set too low (512M instead of 1G). That's been corrected.

On one [failed attempt](https://federalist-staging.18f.gov/sites/42/builds/3246/logs), I noticed a security warning about the version of ruby specified in `.ruby-version`:

```
Please be aware that you just installed a ruby that requires 2 patches just to be compiled on an up to date linux system.
This may have known and unaccounted for security vulnerabilities.
Please consider upgrading to ruby-2.4.1 which will have all of the latest security patches.
```

I'm unclear why it failed that time but then worked after a restart. Maybe a blip in rvm? It might be worth using a newer version of ruby anyway as it should speed up build times a little bit since it won't have to apply those security patches.

coreycaitlin:
Merging — and let's make sure to do a careful data review before launching to prod.

PR #2507, opened 2017-12-04
Style updates / cleanup

ericronne:
This feels like a good time to merge this chunk of changes, since we're creating the style guide now.

Updates include:
- Simplified and reorganized color palette
- Made subtle typography changes here and there
- Created a "pre-footer" nav module and applied to case studies and "diving graphics" pages.

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/typography-normalizing-1/)


coreycaitlin:
It looks like the build failed, and at a glance this seems to include some non-style-related data changes... @msecret could you take a look and make sure this won't re-break anything?

ericronne:
Odd. I didn't change any data, so i would think git would ignore those files.

ericronne:
mergypoo??

_sass/blocks/_footer.scss
```diff
@@ -12,24 +12,24 @@ footer,
...
 
 footer,
 .footer {
-  background-color: $light-black;
+  background-color: $black-light;
   margin-bottom: 0;
   padding-bottom: ($base-padding-extra * 1.5);
   padding-top: ($base-padding-extra * 1.5);
 
   // .fa-circle {
```
msecret: I saw we just ret rid of these if they're commented out

_sass/blocks/how-it-works/_footer.scss
```diff
@@ -1,29 +1,36 @@
-.revenues_page-footer {
-  padding-bottom: $base-padding-large;
-  padding-top: $base-padding-large;
+.pre-footer {
```
msecret: What is `pre-footer`, as in, what is the purpose of the component? I see it used a lot in this PR

_sass/blocks/how-it-works/_steps.scss
```diff
@@ -8,47 +8,27 @@
 
   @include respond-to(medium-down) {
 
-    @include heading(h4);
+    @include heading(h3);
 
   }
 
-  @include respond-to(tiny-down) {
+  @include heading(h2);
```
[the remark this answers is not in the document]
msecret: This should actually be fixed. So `@include heading(h2);` should be moved to line 10 right after `.revenues_steps-heading {`

ericronne:
Stop the presses: the new "pre-footer" is appearing in downloads. Must de-bug.
image

ericronne:
@msecret you able to tackle this? I think in content.html we need a line to prevent the pre-footer from appearing in downloads.

image

coreycaitlin:
After talking with @msecret and @brentryanjohnson, I think it makes sense to avoid adding complexity to the `content.html` template, especially given the additional work we'd need to do to make sure it only showed up on the appropriate pages. It's definitely better to have the design fixed on the diving graphic pages, but we just aren't sure that the UX payoff of adding a selector at the bottom of case studies makes it worth the maintenance burden here. (Especially since someone would have to scroll past very long content + all footnotes to reach this element.)

ericronne:
can we chat before this merges

msecret:
Discussed with Eric, we're ready to merge now

Issue #2504, opened 2017-12-01
Broken GDP data value (percent)


brentryanjohnson:
The percent value in the [GDP chart](https://revenuedata.doi.gov/explore/#gdp) isn't pulling in the data. We're also missing the 2016 data.

![broken_gdp](https://user-images.githubusercontent.com/32855580/33498555-a357ce22-d686-11e7-9261-0f06bd9bd26a.PNG)

coreycaitlin:
🎈 This is now fixed!

Issue #2506, opened 2017-12-02
Data: Update GDP data to 2016


coreycaitlin:
As @brentryanjohnson noted in #2504, GDP data is out-of-date. We should pull in the latest data (likely 2016), update it, and document that process fully.

coreycaitlin:
This is done with #2526

PR #2523, opened 2017-12-07
Update gdp data

msecret:
Fixes issues #2504, #2506 

[CircleCI](https://circleci.com/gh/18F/doi-extractives-data/tree/update_gdp_data)

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/update_gdp_data/)

- Downloaded the new data from BEA
- Updated the data on site with the data update scripts

/cc relevant people


coreycaitlin:
Huh, I'm still seeing this missing on both [national](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_gdp_data/explore/#gdp) and [state](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_gdp_data/explore/IL/#gdp) pages.

Looking at the updated data files, it looks like past years' data changed slightly (which often happens), but I'm not seeing new data for 2016.

msecret:
OK national gdp seems to have 2016 now, as do states. The states.tsv file was deleted somehow?

There were still some failed requests when I ran the script, I might look into that further, but now would be a good time to check the data on site. There may be additional changes on site to make to get data to show up. I'm waiting for federalist to build

coreycaitlin:
It looks like it works! The $ number looks good, but looking at the national % number (in conjunction with #2504) makes me think something's wonky. (That is, the national % issue might be a longer-standing bug, not a new thing.) Let's make sure to validate the $ figures **and** spot-check the % enough to figure out the scope of that bug.

brentryanjohnson:
I'll help spot-check the data this morning. 

We should probably expedite a fix for production release because of the way the GDP percent value affects the state descriptions.

image

mentastc:
two questions, the numbers are down a lot, are we pulling in all the data? second, the percent on the national page is 1.4% and it never changes for any of the years, is this correct? @coreycaitlin @msecret

brentryanjohnson:
@coreycaitlin and @msecret 
I wonder if [this csv file](https://github.com/18F/doi-extractives-data/blob/dev/data/gdp/bea/gdp-by-industry.csv) has something to do with our GDP percent value issues. It's titled _Value Added by Industry as a Percentage of Gross Domestic Product_. 

It hasn't been updated since 2014, but it also seems to be locked in to the 2004 value for "mining."

image

coreycaitlin:
@msecret In the process of fixing this, we realized that we have some apparently-unused files in the `data/gdp` folder:

- `data/gdp/bea/gdp-by-industry.csv`
- `data/gdp/bea/README.md`

(and maybe the rest of the files in the `bea` folder)

What do you think? can we remove those?

msecret:
The only reference I see to those files is for documentation purposes in the `get-bea-data.js` script. For example, https://github.com/18F/doi-extractives-data/blob/dev/data/gdp/get-bea-data.js#L19

coreycaitlin:
@msecret Let's do clean-up on unused files later, if possible — otherwise, does this look good to go?

PR #2568, opened 2018-01-02
Right-aligns numbers in disbursement charts

ericronne:
Numbers in tables should be right aligned, for easier visual scanning.

 - [x] Nat’l chart
 - [x] Opt-in state charts (#2569, need @msecret’s css help here, thx!)


coreycaitlin:
I've added the relevant class (`numeric`) to the opt-in state revenue and disbursements charts — I think this should do the trick. I'll preview and merge as soon as tests pass, etc.

PR #2570, opened 2018-01-02
Chart updates

ericronne:
Replaces outdated SVG graphs with new versions. I also updated colors, for better contrast.

View diffs like so! ...
jyhqs0r0bc


brentryanjohnson:
At some point, we should pull actuals (if available) for the  '16-'17 "projected" figures in the [Alaska revenue sustainability chart](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/er-chart-updates/img/AK_revenue_sustainability.svg).

brentryanjohnson:
@ericronne I haven't heard back from Deloitte regarding the data discrepancies, so in the meantime, perhaps we should pull the questionable charts out of this PR into an issue marked as blocked. That way we can move forward with the charts we know are sound and track the others independently. Let me know your thoughts.

ericronne:
sure, can do …

ericronne:
@jeremiak i wonder if this conflict is due to the new `public/` directory?

Issue #2535, opened 2017-12-14
Research: Synthesize what we learned


coreycaitlin:
Let's make sense of what we learned from #2476 and identify how we'll use this info to move the product forward!

This issue is done when the following have occurred:
- [x] group synthesis session scheduled (for Tues 12/19)
- [x] all researchers (Chris, RJ, RW, Kim & Alex) meet for group synthesis
- [x] a brief research summary is added to [this folder](https://github.com/18F/doi-extractives-data/tree/research/research/15_cheeseycheetah) in the research branch
- [x] a new issue is created for next steps identified during synthesis

alexpandel:
FYI, I'm being a goober and starting my draft in [a Google doc here](https://docs.google.com/document/d/17j1lcfViQQGvS4iXBOHzK0RPSLAVrstvf8HQFMa0AYY/edit#) so I can make mental notes to myself more easily while I'm in process, but I plan to transfer this into a PR on Github for review once I have a first draft ready for some other :eyes:!

alexpandel:
I didn't get as far as I would have liked on this, so I'm actually gonna leave it in [the google doc](https://docs.google.com/document/d/17j1lcfViQQGvS4iXBOHzK0RPSLAVrstvf8HQFMa0AYY/edit#) vs transferring it to Github since it's still rough enough that it might be easier for y'all to collaborate on in the doc itself. Got a very rough narrative flow mapped out, with some placeholders for specific examples from the Mural/notes if needed.

alexpandel:
This is now done, and posted [here](https://github.com/18F/doi-extractives-data/blob/research/research/15_cheeseycheetah/research_results.md)! 🎉

PR #2577, opened 2018-01-04
Added first draft of research_results.md

alexpandel:
@coreycaitlin & @brentryanjohnson, would love your eyes on this - feel free to make any and all changes you see fit! (addresses issue #2535)

### DIFF
--- a/research/15_cheeseycheetah/research_results.md
+++ b/research/15_cheeseycheetah/research_results.md
@@ -11,12 +11,12 @@ We sourced participants for this research sprint from a list of folks who had re
 
 ## For some internal DOI users, we discovered user needs that the NRRD site will not be able to serve, given its current scope and constraints.
 
-Employees in other bureaus (e.g. BSEE) often need lease-level or highly detailed data to manipulate in order to answer specific questions. They may also need detailed data in order to slice the data by different geographies, such as basin, region, or tribal area.
+Employees in other bureaus (e.g. BSEE) often need highly detailed data to manipulate in order to answer specific questions. They may also need detailed data in order to slice the data by different geographies, such as basin, region, or tribal area.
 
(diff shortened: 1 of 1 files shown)


coreycaitlin:
This looks great @alexpandel! I made a few [minor edits](https://github.com/18F/doi-extractives-data/pull/2577/commits/c89c5bc0623eaf0af93b566337281b2de77c360a), but I think this structure works well.

The only other thing I heard mentioned in synthesis that _might_ be worth noting here is the question about longer time-series...what do you think about adding that to the "other" section? (If it wasn't mentioned by enough people to be worth including, I completely defer to you on that!)

alexpandel:
👌👌👌 edits, @coreycaitlin! Hmm yeah the longer time-series one was tricky. It's probably worth mentioning (because yeah, we def heard it), but I'm not sure if there's a conclusion or recommendation we can definitely draw from it yet, since the notes on that issue don't get very deep into the "why"... @brentryanjohnson, you were in both of the interviews where that was mentioned - any thoughts or insight on this one?

brentryanjohnson:
@alexpandel I definitely consider it worth mentioning, although I agree it is difficult to track to a deliverable. What we heard was a desire for "historic" data in order to observe trends that are only discernible by looking further back. Depending on the analysis, it may necessitate data that goes back 5 more years, or it could require 30 years back. I don't think we know enough about the trend analyses users are conducting to know for sure "how much" data would be useful.

As a side note, we do know that data is available back to 1982 on the [statistics site](https://statistics.onrr.gov/Information.aspx).

alexpandel:
Roger @brentryanjohnson 👍 Just added a section about long-term trends - this puppy should be ready for a final review and merge if all looks good to you two!

PR #2588, opened 2018-01-09
Some fractal work

jeremiak:
[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/er-fractal_css_hacking/)

Ok, so we got a bunch of Fractal work done in this branch and upgraded the node version (#2586). We want to merge this into `develop` so that the branch isn't so long lived and we can have fewer merge conflicts. This work is not YET done but we just want to be working off the same commit.


ericronne:
💥

Issue #2563, opened 2017-12-22
Proposal: Use webpack instead of Jekyll for many build tasks


jeremiak:
Currently, we leverage Jekyll as one of the main build tools. For example, it watches and compiles the SCSS directory for us. This is out of the box behavior for Jekyll but one of the problems is that we do things differently for JS and some other files and `jekyll` isn't available at the time we need it on Federalist. We use webpack to compile our Javascript, but perhaps we should use it to compile the CSS files and move other files like `images` or `fonts` to a distinct compilation folder for static assets (`/public`).

Our current set up presents a challenge for using Fractal because:
1. Federalist doesn't have `jekyll`/`bundle` available at the time it runs `npm run federalist`
2. To make Fractal changes requires waiting for Jekyll to rebuild the site

The goal would be to have Jekyll only do template/HTML compilation and have webpack do CSS/JS compilation and img/font asset moving. This would enable Fractal to work without Jekyll (ensuring that it works on Federalist and decreasing the amount of time it takes to do work on the styleguide itself).

@msecret and I came up with this list of tasks to get this done:
- [x] Set up webpack for building CSS
- [x] Set up webpack to dump compiled JS into public and not /js/lib (which is confusing to have both source and compiled files in the same place)
- [ ] Set up webpack for moving images and fonts around
- [x] Configure Jekyll to ignore all of those files (hopefully this has a nice secondary benefit of speeding up the build time)
- [x] Rename static paths to use `/public` if we can't figure out how to do this dynamically with Jekyll defaults
- [x] Configure Fractal to use `/public` and not rely on Jekyll at all

jeremiak:
I have been doing much of this work on the `move-some-css-around` branch and have webpack building the CSS so that Jekyll need not worry about it and the Fractal app can get access to it without `ruby` or `jekyll` being present (which is what happens in the Federalist builder).

However, I have run into an issue where the site cannot build on Federalist because (the current theory goes) the dependencies and app take up too much memory ([slack link for convo in #federalist-support](https://gsa-tts.slack.com/archives/C1NUUGTT5/p1514997464000239)). I might try updating the node version since npm v5 changes the way dependencies are installed so as to have a lower memory footprint.

ericronne:
Thanks for the update, and for keeping up the fight! 💪

coreycaitlin:
Closing this out, since this work has been merged in. Remaining troubleshooting will get new issues!

PR #2601, opened 2018-01-16
Chokidar for watching

jeremiak:
This is related to #2563 but it mostly is aimed at getting the site to build, and build on changes, in a reasonable amount of time on @brentryanjohnson's machine. The major feature of this pull request is to use node (and a library called [`chokidar`](https://npm.im/chokidar)) to watch for file changes and then tell Jekyll to build the site instead of relying on Jekyll's watch feature. In our testing, it seemed that Jekyll, especially when run inside of Docker, would frequently miss file changes and provide no indication of if it had noticed changes and was processing them.

With our new configuration, we wrote more code to do the watching, but it is more reliable and enables us to output information to the console so that site developers are made aware of the system recognizing changes.

Because we are already using `docker-compose` I hope that this change is mostly invisible to developers and should work with `docker-compose up`.


coreycaitlin:
@jeremiak This didn't break anything for me, but I'm not sure it fixed it, either! I'm still able to run `docker-compose up` successfully, and that does build the site. But I'm not seeing any new indication that it's seeing and regenerating the site based on changes to markdown files. Happy to pair on this if it'd be helpful for you to see what I'm seeing!

jeremiak:
@coreycaitlin Let's pair tomorrow on this! Maybe before you do your docker onboarding session with @alexpandel so that we can make sure you have at least one working implementation?

docker-compose.yml
```diff
@@ -1,20 +1,35 @@
 version: '2'
 services:
-  app:
+  chokidar:
```
msecret: Is there a reason we should call it this rather then just `app`?

coreycaitlin:
Okay, I've confirmed that this _does_ fix the "build twice to make styles show up" problem 🎉

docker-compose.yml
jeremiak: I'm open to just calling this `app`, I don't have a really strong opinion. My thinking was that it is a bit nicer in the output to actually know what the service was. Maybe we should call this `jekyll`?

coreycaitlin:
We've confirmed that watching is _working_ 👍 but chokidar is not watching enough files (currently only watching files in the `_layout` directory. Once we address that, let's get this merged in!

scripts/jekyll-watch.js
```diff
@@ -0,0 +1,48 @@
...
+        resolve();
+      } else {
+        reject(`${cmdline} exited with code ${code}!`);
+      }
+    });
+  });
+}
+
+function runJekyll() {
+  return run('/doi', 'jekyll', ['build', '--incremental'])
```
msecret: Is any of the stuff happening in `serve-fast` still relevant here (besides `--incremental`)?

jeremiak: I don't believe so. I'd like to remove the many different `*serve*` files and only have on Jekyll config. The host binding and such happens with docker and I don't know that we want to run the profiler/tracer every time we run the build. What do you think?

msecret: Agreed

jeremiak: Done in https://github.com/18F/doi-extractives-data/pull/2613

coreycaitlin:
We've confirmed that this PR successfully resolves build issues so that:

1. The site builds all files successfully, in the right order, on the first try
2. It watches for file changes and rebuilds the site when file changes are saved
3. Icon fonts are appearing correctly in local builds

The Federalist issue is *not* related to any changes made here, but is an unrelated (albeit annoying) Federalist/cloud.gov issue.

Let's merge this puppy 🐶 🚀

jeremiak:
Right now this is failing to build on Federalist but that is actually an issue with the platform and not this project. The cloud.gov team recently updated their platform to ensure that containers can only use 2GB of disk space, and between this projects, its dependencies, and Node, Ruby, and Python (for Federalist) the project weighed in at 2.4GB. Thus, Federalist isn't able to finish our builds.

The Federalist team is working on expanding the container size to 4GB, but @coreycaitlin and @msecret have tested this on their machines so we're going to merge now.

Issue #2360, opened 2017-08-31
Build fractal style guide


winzenburg:
I'd like a style guide for the USEITI website, so we (ONRR) can maintain and extend the design vocabulary as the site evolves.

gemfarmer:
We might consider using [Fractal](http://fractal.build/) as the backbone of a new style guide. There are a few 18F projects, including USWDS that have used it successfully without too much overhead.

winzenburg:
Looking for a font that works better across browser

ericronne:
#875 is the root issue. We may stick with Lato, if we don't find a killer alt soon.

ericronne:
Before he left the premises, Brian recommended that we scrap the current style guide and start over from scratch. The current one is very out of date, and likely not worth keeping.

ericronne:
@msecret

msecret:
The best option these days is to use something like [fractal](https://github.com/frctl/fractal).

ericronne:
Check out (and off) these fun steps!
(more to come) …
- [ ] Stub out basic Fractal shell #2443
- [ ] Conduct a style inventory #2444
- [ ] Simplify type styles @2445
- [ ] Adjust grids and spacing #2446

ericronne:
For better tracking, should we create two new documentation issues to add to the checklist: technical (@msecret) and "content" (@coreycaitlin)? 

Then we could close out #2482 (after narrowing it to "mvp documentation" eg) and check off that box.

coreycaitlin:
Let's keep this open until we've cleared up a few remaining issues with the build process.

coreycaitlin:
We also need to test this locally on ONRR machines, using Docker.

coreycaitlin:
Next step: @jeremiak and @mentastc to test during pairing today, @brentryanjohnson to test on his machine.

coreycaitlin:
We've opened #2620 to reflect the work on running the style guide locally. Closing this issue out now, since all work is reflected elsewhere.

PR #2602, opened 2018-01-16
Use consistent way of including fonts

msecret:
Previously, the eiti font was being included in the code differently
then the rest of the fonts, causing confusion when changing build
processes. Now it's included in the same way as the other fonts,
after being generated by fontsquirrel.

Fixes issue #2360 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/ms-fix_fonts/)


coreycaitlin:
This looks fixed in the preview link, but I'm still seeing the broken icon fonts when I pull it down to run locally...
screen shot 2018-01-16 at 5 40 23 pm

coreycaitlin:
@msecret Did we expect this to solve the icon-font problem? If not, what else should I be looking for?

msecret:
I definitely expected to solve it (it was working locally). Good catch though, I'll have to look into it further.

msecret:
Move this to chokidar branch

Issue #2533, opened 2017-12-14
Add contact information for data services


brentryanjohnson:
With the integration of statistics.onrr.gov, we need a way for users to contact our data services team to request data that may not be available on the site. We'll also be redirecting users from statistics to NRRD. I'm hoping @coreycaitlin and @ericronne can help me think through design ideas to support contact info.

coreycaitlin:
Options discussed with @brentryanjohnson:

- add to footer
- add to "About"
- add to "Downloads: Data and Documentation" page — this feels like the right first answer

On that page, we could create some kind of callout inviting people to contact Data Services if they don't find what they need. Ideally, this could be an include so that we can have this callout on all data/documentation pages.

brentryanjohnson:
Contact information for data services:
- onrrdatarequests@onrr.gov
- 303-231-3162

coreycaitlin:
Next steps:

- Discuss with @ericronne what this callout might look like
- Discuss with @msecret how to implement as an include (rather than as content we have to copy/paste onto multiple pages)

ericronne:
I think About, Downloads, and footer all make good sense. Hopefully a lot of interested users will want to interact with us, about this and all sorts of other things. It seems a little odd that the only contact info we now have is a mailing address 🐢 

Let's chat!

coreycaitlin:
For content to start with, let's try:

> If you need ONRR data that is not available here, contact Data Services with your request:
> 
> - onrrdatarequests@onrr.gov
> - 303-231-3162

ericronne:
Here's a little module that can go almost anywhere necessary. I capriciously modded the text, to fit onto one line. Feel free to un-mod :)

![image](https://user-images.githubusercontent.com/9259185/34269750-1ffe9248-e64b-11e7-9ad3-645b0f92bda7.png)

brentryanjohnson:
I think this is a good start, @ericronne. Why don't we start here and reassess based on user research as we go along. 

One minor edit...the email address should be: onrrdatarequests@onrr.gov

Any additional thoughts @coreycaitlin or @mentastc?

ericronne:
Do we need "onrr" in the pre-@ part of the address, since it's the domain?

brentryanjohnson:
I don't think the email address will resolve without it...we could look into simplifying the address with the ONRR IT folks.

coreycaitlin:
Since it sounds like this is an existing email address that the Data Services team uses, let's include as-is.

[...]

PR #2612, opened 2018-01-24
Adds prototyped contact information for data retrieval team

brentryanjohnson:
Fixes issue #2533

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/)

Valid concerns (documented in #2533) led us to explore alternatives for placement of content information for the data retrieval team. An [earlier prototype](https://github.com/18F/doi-extractives-data/issues/2533#issuecomment-358377867) used the right column (in desktop view) to provide 'sticky' contact information. This would have departed from the convention of using that area for navigation. So...

- I've prototyped adding a subheading with contact information at the bottom of the page _along with_ adding navigation items to the download pages. Currently, only the [main downloads page](https://revenuedata.doi.gov/downloads/) has navigation items. The others (e.g. [Federal revenue by location](https://revenuedata.doi.gov/downloads/federal-revenue-by-location/)) do not. **Now the user can see there is contact information available on the page no matter where they land on the page.**

- This introduces a new file in the `_data` directory called `contact.yml`. This file contains variable assignments for contact information, so we can change them in one place and the associated values change on each page they are referenced (this will likely come in handy, since the name of our data team recently changed from "Data Services" to "Date Retrieval").

- Finally, the second-level headings on the [main downloads page](https://revenuedata.doi.gov/downloads/) were all `h1`s, so I've changed each of them to the semantically appropriate `h2`.

For now, I've added the contact info and nav items on only two pages, so everyone can review and compare them to the existing layout.
- [main data downloads](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/)
- [federal revenue by location](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/federal-revenue-by-location/)

### DIFF
--- a/_data/contact.yml
+++ b/_data/contact.yml
@@ -1,0 +1,10 @@
+data_retrieval:
+    name: Data Retrieval
+    email: onrrdatarequests@onrr.gov
+    phone: (303) 231-3162
+
+information_data_management:
+    name: Information and Data Management
+    street: 1849 C Street NW MS 5134
+    city: Washington, D.C.
+    zip: 20240
--- a/_downloads/disbursements.md
+++ b/_downloads/disbursements.md
@@ -89,4 +89,4 @@ _Other._ Certain offshore funds are directed back to the federal agencies that a
 
 ## Contact us
 
-Do you need {{ "ONRR" | term }} data that isn't here? Contact our {{ site.data.contact.data_retrieval.name }} team at [{{ site.data.contact.data_retrieval.email }}](mailto:{{ site.data.contact.data_retrieval.email }}) or {{ site.data.contact.data_retrieval.phone }}.
+{% include contact.html %}
--- a/_downloads/disbursements.md
+++ b/_downloads/disbursements.md
@@ -3,6 +3,23 @@ title: Disbursements | Documentation
 title_display: Disbursements
 layout: content
 permalink: /downloads/disbursements/
+nav_items:
+  - name: introduction
+    title: Top
+  - name: scope
+    title: Scope
+  - name: data-publication
+    title: Data publication
+  - name: data-dictionary
+    title: Data dictionary
+    subnav_items:
+      - name: onshore
+        title: Onshore
+      - name: offshore
+        title: Offshore
+  - name: contact-us
+    title: Contact us
+selector: list
 breadcrumb:
   - title: Downloads
     permalink: /downloads/
@@ -34,11 +51,11 @@ We also have [notes on this data](https://github.com/18F/doi-extractives-data/wi
 
 This dataset includes natural resource disbursements for U.S. federal lands and offshore areas, and Indian lands. It does not include privately-owned lands, or U.S. state lands. The dataset is tracked and managed by the Department of the Interior’s Office of Natural Resources Revenue. It contains disbursement information to funds for fiscal years 2012-2015. Disbursements of revenue due to extractive activities on U.S. federal lands occur monthly; this dataset is a sum of those disbursements by fiscal year.
 
-## Data Publication
+## Data publication
 
 The disbursement dataset is updated in December after the end of the federal government fiscal year.
 
-## Data Dictionary
+## Data dictionary
 
 Laws treat revenues from offshore natural resources and onshore natural resources differently. There are set percentages and amounts from each that go certain places every year.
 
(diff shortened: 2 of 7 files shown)


jennmalcolm:
This looks like a good approach to balance convention with the need to make the contact info easily visible and accessible. We can test this later to make sure users are on the same page, but for now, this is enough for the Stats site redirect to happen. Thanks @brentryanjohnson

brentryanjohnson:
Excellent. Thank you for the feedback @ericronne and @jennmalcolm. I'm going to add the contact and nav items to the other download pages, and then we can get this merged in!

brentryanjohnson:
I have added the contact information and nav items to the following `/downloads/` pages:

- [Downloads](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/)
- [Federal production by location](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/federal-production/)
- [Federal revenue by location](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/federal-revenue-by-location/)
- [Federal revenue by company](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/federal-revenue-by-company/)
- [Disbursements](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/data-contact-info/downloads/disbursements/)

I **did not** add the contact information or nav items to [Reconciliation](https://revenuedata.doi.gov/downloads/reconciliation/). It is the only other page in the `/downloads/` directory, but I didn't use the same convention because of these assumptions:
1. The data and content were heavily dependent on participation in EITI
2. The data retrieval team doesn't traffic in this particular data, and thus would not be the appropriate contact for this data.

If this approach is acceptable for now, please review and merge.

coreycaitlin:
@brentryanjohnson This looks fantastic. I agree with @jennmalcolm that this is a great first step toward including this contact information. This _also_ brings those downloads pages into our newer sidenav pattern AND makes the pages easier to skim and navigate, to boot. That seems like a huge win.

Two minor questions/notes (that don't need to block merging):

- Are we using the physical contact information now set in the `contact.yml` file in the footer (or anywhere), or did you include it for future use?
- At some point we might make it even easier to pull in the contact blurb, by putting that markup/structure in `_includes`...but that seems minor, and can happen later.

In general, this is 💯  and I agree with Jennifer that we should merge!

brentryanjohnson:
@coreycaitlin 

- Yes, I did add the mail address (currently in the footer) to the `contact.yml` file, but I didn't replace the footer content with the variables. I assigned it the following way so we could break the address onto new lines, as opposed to it being one long string that can't be formatted how we want it to be.
```data_retrieval:
    name: Data Retrieval
    email: onrrdatarequests@onrr.gov
    phone: (303) 231-3162

information_data_management:
    name: Information and Data Management
    street: 1849 C Street NW MS 5134
    city: Washington, D.C.
    zip: 20240
```

- I agree, an `include` makes sense. As it is now, we wouldn't have to change each instance if the contact information changed, but we would have to change each instance if we want to reword the content _around_ those variables. An include is a great next step!

_data/contact.yml
```diff
@@ -0,0 +1,10 @@
+data_retrieval:
```
msecret: Does data retrieval not have a street address? And does information management not have an email? Asking just for consistency sake.

_downloads/disbursements.md
```diff
@@ -70,3 +87,6 @@ _States._ States receive federal Outer Continental Shelf revenue in two ways:
 
 _Other._ Certain offshore funds are directed back to the federal agencies that administer these lands (e.g., BOEM and BSEE) to help cover the agencies’ operational costs.
 
+## Contact us
+
+Do you need {{ "ONRR" | term }} data that isn't here? Contact our {{ site.data.contact.data_retrieval.name }} team at [{{ site.data.contact.data_retrieval.email }}](mailto:{{ site.data.contact.data_retrieval.email }}) or {{ site.data.contact.data_retrieval.phone }}.
```
msecret: What happened to this being an include rather the repeating the information across the different files?

_data/contact.yml
brentryanjohnson: Data retrieval probably does have a street address, but there is no use case for including it right now. We are developing a new email address for Information and Data Management, but it doesn't exist yet.

_downloads/disbursements.md
brentryanjohnson: We thought we would start here and abstract it into an include later. If it's trivial to create the include now, I don't see an immediate reason not to. We just don't want to hold up the process of publishing the contact info.

msecret: I'd argue to the include should happen now before we commit it to dev, with the idea that any code that goes into dev is using the best practices. I'm definitely open on this though.

brentryanjohnson: Okay...I created the include and replaced the content with the include reference on the downloads pages.

PR #2613, opened 2018-01-24
Remove *serve* files from project root

jeremiak:
We don't use these files anymore as we leaned heavily into using docker as our way of serving the site during development.

- Remove `serve`, `serve-fast`, `re-serve`, `re-serve-fast`, and `_config-fast.yml` from the project


coreycaitlin:
I pulled this down, and everything still works to run things locally with Docker! I also updated the `CONTRIBUTING.md` file to remove the instructions about running `./serve`.

Ran into an odd merge conflict when I pushed that change up, but I _think_ all's well to merge? If I created a problem or accidentally undid something, please feel free to trash my commit — it would only take 2 minutes to re-do.

coreycaitlin:
@jeremiak Could you take a quick look to make sure I didn't mess something up before merging?

jeremiak:
@coreycaitlin Just took a look and I rebased so the branch will merge cleanly. I can merge this once it turns green

coreycaitlin:
Now I'm seeing weird merge-conflict artifacts here: https://github.com/18F/doi-extractives-data/pull/2613/files 😕

jeremiak:
Woah, I messed up that merge. One second, let me take a look

jeremiak:
Good catch @coreycaitlin, I actually made the changes but didn't save the file. Saved, and pushed.

Issue #2557, opened 2017-12-20
Mothball the old styleguide, and deal with associated cruft


ericronne:
Our current [style guide](https://revenuedata.doi.gov/styleguide/index.html) is incomplete, and will be obviated by the fractal one currently under way in #2360. Once that's complete, let's delete the old one, and clean up sass files which reference it.

coreycaitlin:
@msecret to investigate whether this is done...stay tuned.

jeremiak:
I'm calling this done since 2614 was merged

PR #2614, opened 2018-01-25
Remove references to old styleguide

msecret:
Includes:
- Documentation updates
- Removing old files for kss styleguide
- Removing references to old files and folders for odl styleguide.

Fixes issue #2557 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/remove_old_styleguide/)


CONTRIBUTING.md
```diff
@@ -103,11 +103,7 @@ If deploying the site to a production environment, make sure to minify the JS fi
 ### Styleguide
 
 ```sh
-npm install --dev
-npm run init-styleguide
-cd styleguide-template && npm install
-cd ..
-npm run watch-styleguide
+docker-compose run webpack npm run build-styleguide
```
jeremiak: Do you want to also include some information about how to run the styleguide in watch mode instead? It seems like you'd want to be doing watch mode during dev and build would really only happen on Federalist?

_maps.yml
```diff
@@ -4,7 +4,6 @@ exclude:
   - package.json
   - show
   - test
-  - styleguide-template
```
jeremiak: This is a larger question about this `_maps.yml` file: what is it used for?

msecret: I don't know, we should put it in the question doc.

msecret:
In terms of a docker service, I think our docker build is a little slow already, so am a little worried about adding more to it. That being said, they run in parallel, so not sure perf will be changed that much by doing it.

coreycaitlin:
@jeremiak Now that we've tackled the style guide in #2625, is this ready for final review + merge?

PR #2615, opened 2018-01-25
Use different _config.yml for dev and federalist

jeremiak:
During development, we use a system of docker, express, and jekyll to make the environment run on different machines. However, in doing so we told jekyll to ignore the /public directory so that it could be managed by webpack & express. On Federalist, there is no Express server so jekyll needs to include /public when it builds the site.

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/jekyll-public/)

- Add `_config-dev.yml` for use with docker

### DIFF
--- a/_config-dev.yml
+++ b/_config-dev.yml
@@ -1,0 +1,64 @@
+exclude:
+  # top-level log
+  - '*.log'
+  # markdown files
+  - CONTRIBUTING.md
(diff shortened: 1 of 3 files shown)


coreycaitlin:
Alright, this is building on Federalist 🥇 and I'm able to run it locally as expected 🎉 . I'm going to merge!

PR #2616, opened 2018-01-26
Refactor some redirects

coreycaitlin:
We currently handle redirects in two ways: through standalone files [like this one](https://github.com/18F/doi-extractives-data/blob/dev/_explore/economic-impact/exports.html), and through the `redirect_from` gem, which uses the frontmatter of the destination page, [like so](https://github.com/18F/doi-extractives-data/blob/dev/_how-it-works/laws-and-regulations/state-laws-and-regulations.md).

The `/_explore/` folder is an artifact of an old site architecture (circa 2015), and predates when we started using the `redirect_from` gem.

This PR tackles the low-hanging fruit in that folder by replacing relatively simple redirect files with `redirect_from` entries in the frontmatter of the appropriate destination files. There were a couple I couldn't figure out quickly, so I left them as-is for now.

The catch is that `redirect_from` doesn't seem to work in Federalist previews. As far as I can tell. So I'm not sure quite how to test that this is working as intended. Thoughts @jeremiak or @msecret?

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/refactor-redirects/)

### DIFF
--- a/_explore/economic-impact/exports.html
+++ b/_explore/economic-impact/exports.html
@@ -1,6 +1,0 @@
----
-title: Exports | Explore Data
-layout: redirect
-permalink: /explore/exports/
-redirect_url: /explore/
----
--- a/_explore/economic-impact/gdp.html
+++ b/_explore/economic-impact/gdp.html
@@ -1,6 +1,0 @@
----
-title: Explore | Gross Domestic Product
-layout: redirect
-permalink: /explore/gdp/
-redirect_url: /explore/#gdp
----
--- a/_explore/economic-impact/jobs.html
+++ b/_explore/economic-impact/jobs.html
@@ -1,6 +1,0 @@
----
-title: Jobs | Explore Data
-layout: redirect
-permalink: /explore/jobs/
-redirect_url: /explore/#employment
----
--- a/_explore/production/production-all-lands.html
+++ b/_explore/production/production-all-lands.html
@@ -1,6 +1,0 @@
----
-title: All Lands Production | Explore Data
-layout: redirect
-permalink: /explore/all-lands-production/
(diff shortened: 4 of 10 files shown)


coreycaitlin:
Once we get this to wherever it's testable (which may be production), here are the urls to check:

- /explore/exports/ _should go to_ /explore/
- /explore/gdp/  ➡️  /explore/
- /explore/jobs/  ➡️  /explore/
- /explore/all-lands-production/  ➡️  /explore/
- /explore/federal-production/  ➡️  /explore/
- /explore/disbursements/  ➡️  /explore/
- /explore/federal-revenue-by-location/  ➡️  /explore/
- /explore/federal-revenue-by-company/  ➡️  /how-it-works/federal-revenue-by-company/2016/
- /explore/corporate-income-tax/  ➡️  /how-it-works/corporate-income-tax/

coreycaitlin:
I should note: this PR also changes the behavior of the redirects slightly. The old behavior sent users to _specific anchor links_ on the `/explore/` page. My changes remove the specificity, and simply send users to the top of the `/explore/` page. I think this is okay for a couple of reasons:

- The old links are pretty old at this point (particularly for the age of the site), and we have no reason to think that many — or any! — users still rely on those links.
- The new redirect behavior still allows users to get to the data (and avoids 404-ing them).
- I think the benefits of moving toward simpler, more maintainable, and more consistent redirect handling outweigh the small — and possibly only hypothetical — cost to users.

coreycaitlin:
@jeremiak Now I see! that totally makes sense, and I'm glad we have at least some indication that it's working. Since you've both approved, I'm merging!

Issue #2486, opened 2017-11-29
Data: Update state revenue and disbursements data for AK, MT, WY


coreycaitlin:
We have updated data for our three prior opt-in states. [Data here (limited access)](https://drive.google.com/drive/folders/1NMOZPvOR8tpgYHnEsfrkP_RuglvsH2Zb)

coreycaitlin:
Next step: break down next steps and smaller tasks to complete this

coreycaitlin:
Currently blocked by a data discrepancy in WY revenue/disbursements — it looks similar to something we [ran into last year](https://github.com/18F/doi-extractives-data/pull/1813#issuecomment-243847159). @brentryanjohnson is working with Deloitte to resolve.

brentryanjohnson:
Deloitte responded with the following:

>So, given the disparity within the Excel, I am inclined to go with the highlighted number on the ‘Distributions by Revenue Stream’ tab (2,126,209,645). I say this because we can trace it back to the tally of the revenues streams which have corresponding sources. Without the sources for the other figure within the excel, I am hesitant to list that number

Since we updated the WY opt-in data with the figure above, I am unblocking this one.

coreycaitlin:
Moving the full set of checklists here:

## Wyoming

- [ ] State revenue chart [🏇 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_optin_state_data/explore/WY/#state-local-revenue) - 2016 data, but needs validation
- [ ] State disbursements  [🏇 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_optin_state_data/explore/WY/#state-disbursements) - 2016 data, but needs validation
- [x] Content updates ([source; limited access](https://drive.google.com/drive/folders/13ZJPUWeZbUs7kFQzNEGVvUSyalwPiPLi))
- [ ] Static charts — update or remove

## Montana

- [ ] State revenue chart [🐻 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_optin_state_data/explore/MT/#state-local-revenue) - still showing 2014 data
- [ ] State disbursements  [🐻 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/update_optin_state_data/explore/MT/#state-disbursements) - still showing 2014 data
[...]

PR #2623, opened 2018-01-29
Change the revenue years for both optin states

msecret:
Fixes issue #2486 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/ms-opt_in_data/)

### DIFF
--- a/_data/opt_in_state_revenues/AK.yml
+++ b/_data/opt_in_state_revenues/AK.yml
@@ -2,39 +2,55 @@ streams:
   All:
     Total:
       '2015': 2568248186
+      '2016': 1761666223
     General Fund:
       '2015': 1743439860
+      '2016': 1132972459
     Alaska Permanent Fund:
       '2015': 510400000
+      '2016': 390500000
     Constitutional Budget Reserve Fund:
       '2015': 297308326
+      '2016': 225293764
     Other Restricted:
       '2015': 9200000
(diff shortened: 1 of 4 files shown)


coreycaitlin:
This seems like a good indicator of why AK, at least, isn't working: https://github.com/18F/doi-extractives-data/tree/dev/_data/opt_in_state_revenues

coreycaitlin:
Huzzah! Changing the year in the state markdown file successfully fixed MT (because the data was there, it just wasn't looking for the right year). It didn't fix AK because the 2016 data isn't there yet for AK. This is great because it means this is working pretty predictably.

coreycaitlin:
🎉 👍 🥇 Alaska is working!!!

brentryanjohnson:
Upon reviewing this PR, I noticed the county choropleths are no longer mapping their respective values properly.

Alaska map comparison between this branch and production site

This doesn't appear to be related to this PR, as I noticed it's a problem in dev, too. @coreycaitlin and @msecret should we open a new issue for this one, or try to take care of it in this PR?

coreycaitlin:
@brentryanjohnson If the maps are broken in `dev` too, let's treat that as a separate bug and PR. One of the ways we can keep things more predictable is by keeping PRs smaller and avoiding long-running unmerged branches.

Issue #2611, opened 2018-01-19
Case study charts: New alt-text for 2017 updates


coreycaitlin:
PR #2570 updated most of the static charts in the case studies and opt-in state pages, but we need to follow up and ensure we've updated the alt text as well, so any users using a screen reader are able to understand the data presented.

PR #2629, opened 2018-02-01
Updates alt text for static charts

brentryanjohnson:
Fixes issue #2611

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/static-chart-alt/)

- Updates alt-text for static charts in case studies and opt-in states.


jeremiak:
We should update the issue template so that it formats the text such that Waffle can link the issues and the PRs together.

Doing that manually for this issue with: Fixes #2611

coreycaitlin:
@jeremiak good idea. I thought it did at some point, but we may have broken it. Let's fix (but separately from this PR).

Issue #2496, opened 2017-11-30
Case study charts: update


brentryanjohnson:
It's time to update our case study charts again. 

- [x] **Boone, Logan, and Mingo Counties**
[Chart 1: Coal production](https://revenuedata.doi.gov/img/counties/wv-production.svg)
![boone1](https://user-images.githubusercontent.com/32855580/33439805-2c45ce30-d5a3-11e7-8437-4d3def2e2264.PNG)

2015 values:

- Total: 21,955
- Boone: 8481
- Logan: 9562
- Mingo: 3912

### Draft for proofing
![01-line chart - boone coal prod](https://user-images.githubusercontent.com/9259185/34310359-ce762a6c-e71c-11e7-8c7a-ef7f67faba70.png)

---

- [ ] **Boone, Logan, Mingo wage and salary chart** 
[Chart 2: Wage and salary employment...](https://revenuedata.doi.gov/img/counties/wv-wage.svg)
![boone2](https://user-images.githubusercontent.com/32855580/33439820-33e4cc7c-d5a3-11e7-9bd1-02762bfa8c56.PNG)

2015 values:

- All other industries: 16,816
- Mining industry: 3743

### Problem with this (and most employment charts)
![02-bar chart - boone wage](https://user-images.githubusercontent.com/9259185/34310367-e05f4a88-e71c-11e7-96dc-d5cab71337c9.png)

---

- [x] **Campbell County**
[Chart 1: Coal production](https://revenuedata.doi.gov/img/counties/wy-production.svg)
![campbell1](https://user-images.githubusercontent.com/32855580/33440034-c54f4cc8-d5a3-11e7-9e6d-0290899439d6.PNG)

2015 values:
- Total coal production: 328,221,901

### Draft for proofreading
![03-line chart - campbell coal prod](https://user-images.githubusercontent.com/9259185/34310415-1f3d7900-e71d-11e7-8693-9bd80b34438b.png)

---
- [ ] **Campbell County wage and salary employment chart**
[Chart 2: Wage and salary employment](https://revenuedata.doi.gov/img/counties/wy-wage.svg)
![campbell2](https://user-images.githubusercontent.com/32855580/33440041-c8893296-d5a3-11e7-9e54-1e3a6db6f12c.PNG)

2015 values:
- All other industries: 16,085
- Mining industry: 6802

### Problem
[...]

PR #2631, opened 2018-02-01
Remaining case-study and opt-in chart updates

ericronne:
Here's a [preview link](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/er-charts-jan/), but the rich view of the diffs is a handy way to review here.


ericronne:
For updates outlined in issues #2486 and #2496

PR #2643, opened 2018-02-09
Removes incorrect phone number from data retrieval contact info

brentryanjohnson:
I discovered that the phone number we pulled over from the statistics site is no longer correct. This update removes the phone number and leaves just the email address for the data retrieval team.


jeremiak:
LGTM!

Issue #2284, opened 2017-05-01
Switch ALL icons to svgs


ericronne:
Currently use font awesome; probably more efficient to use svgs.

This should also serve to fix the horizontal jump seen when clicking glossary terms.

ericronne:
Here's what i found in a semi-manual search …

chevron-right
circle
envelope
facebook
file-text-o
flickr
info
instagram
pinterest
rss
tumblr
twitter
youtube-play

gemfarmer:
From the style guide: https://useiti.doi.gov/styleguide/section-webfont.html

ericronne:
Given the many ways we've implemented icons (and my noob-iness!), i think it's wise for me to switch gears and pair on this later.

shawnbot:
FYI, I made [a little repo](https://github.com/shawnbot/jekyll-svg) to demo some ideas around templatizing SVGs that might be worth taking a look at. LMK if you want to bounce some ideas around!

gemfarmer:
@ericronne totally can do a pairing later. And highly recommend Shawn's repo!

ericronne:
Bump! I should focus on this soon. Eg to avoid bugs like this …
![image](https://user-images.githubusercontent.com/9259185/30500611-c0133f04-9a24-11e7-996c-d3a48d9ea226.png)

coreycaitlin:
Keeping this open — this would help address issues with icon fonts, add text alternatives, and other good stuff.

master12:
related to 2650, fixed the issue that corey reported

PR #2650, opened 2018-02-09
Replace font awesome with svgs

### LINKED ISSUE #2639: Missing icons
## Expected Behavior

- Data and documentation links should load in a "document" icon
Data and documentation link with icon

## Actual Behavior

- "Document" icon isn't appearing

Data and documentation link with missing icon in Firefox

Data and documentation link with missing icon in Chrome

## Steps to reproduce the behavior

- Clear cache and load any data page that contains "Data and documentation" link.

msecret:
Fixes #2639 #2284 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/ms-replace_fa_with_svg/)

- Add font awesome icons as svg files to the img section of the project
- Replace all uses of the `fa` classes with our own styling with svg
- Remove all instances of font-awesome
- ~Combine repeated social media code in the how it works section to make the above easier.~


coreycaitlin:
Unfortunately, the different social-media sections at the bottoms of those pages aren't the same content — some link to BOEM accounts, others to BLM accounts, etc.

msecret:
I bet I can get things working with some variables set on the separate pages.

coreycaitlin:
Let's avoid focusing too much energy on these "get involved" sections, especially given how many broken links we might have here. @brentryanjohnson What do you think about simplifying these sections instead? For instance, we could remove the social icons on these pages or edit down the number of links — either of which would make this easier to deal with.

brentryanjohnson:
Agreed, looking at the [coal page](https://revenuedata.doi.gov/how-it-works/coal/) (for example), none of the descriptions for why users would engage with those particular links matches up with our product vision or user scenarios. I would almost say we can eliminate the social media section entirely, but I'm  willing to be persuaded otherwise if they met a defined user need.

coreycaitlin:
@msecret I think you'd said you made progress on this — any update?

coreycaitlin:
Looks like we've got svgs on the Downloads page 🎉 ...but they're *enormous*:

screen shot 2018-02-13 at 8 55 54 am

After those get sized right, we also have a bunch of instances of these icons on the Explore Data pages (both state & national).

And don't forget to roll back the change that replaced the social sections of each diving graphic with `social.html`.

coreycaitlin:
This is looking good! 🎉  Once you're ready for review, could you assign @ericronne to double-check spacing/alignment of the svgs?

coreycaitlin:
I'm going to pull this down and see if I can sort the alignment out.

coreycaitlin:
Update: nope, this is not for me to do.

coreycaitlin:
Here's the minor alignment issue I'm seeing (this branch is on the left, current live site is on the right):

screen shot 2018-02-21 at 12 40 38 pm

I think the bottom of the image should be aligned with the _underline_ under the link, rather than the bottom of the text itself. So we want it to look like the current live site. @ericronne correct me if I'm wrong!

master12:
@coreycaitlin ok just added a simple align to text bottom. this is the closest i could get to the current state

coreycaitlin:
Nice — that looks much better @master12!

@ericronne I think this is solved 🎉

PR #2652, opened 2018-02-09
Upgrade Jekyll to 3.7.2 and ruby to 2.5.0.

### LINKED ISSUE #2648: Upgrade ruby from 2.2.3 to at least 2.5.0
Currently, the project uses version `2.2.3` of ruby ([`.ruby-version`](https://github.com/18F/doi-extractives-data/blob/dev/.ruby-version)) but the [current stable release Ruby is `2.5.0`](https://www.ruby-lang.org/en/downloads/). In fact, the `2.2.x` branch is only in security patch mode until the end of March 2018. We should upgrade to the latest ruby so that we continue to get security patches.

From the [Ruby 2.2.9 release notes](https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-2-9-released/):
> Ruby 2.2 is now under the state of the security maintenance phase, until the end of the March of 2018. After the date, maintenance of Ruby 2.2 will be ended. We recommend you start planning migration to newer versions of Ruby, such as 2.4 or 2.3.

### LINKED ISSUE #2649: Upgrade Jekyll from 3.0.5 to at least 3.7.2
We are quite behind on updating Jekyll and are currently using [version `3.0.5`](https://github.com/18F/doi-extractives-data/blob/dev/Gemfile.lock#L28), which was released in [April 2016](https://jekyllrb.com/news/2016/04/26/jekyll-3-0-5-released/). Let's upgrade jekyll to the latest version.

line47:
Fixes #2648
Fixes #2649 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/sw-jekyll-ruby-update/)

- Upgrade ruby to 2.5.0 
- Upgrade Jekyll to 3.7.2
- Updated `gems` in config to `plugins` per the new Jekyll specification


coreycaitlin:
It looks like this has some odd effects for some of our maps and state pages:

- The national maps (on the homepage and explore page) have lost their offshore areas
- The national maps are no longer clickable, and the hover tooltext isn't showing up
- The state pages have a number of issues — for an example, see [Alaska](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/sw-jekyll-ruby-update/explore/AK/#federal-revenue)...
    - map shows WA
    - all data is missing

coreycaitlin:
_Adding WIP to remind us not to merge this until we track down some of the effects_

yozlet:
@jeremiak I don't know how far you've got with this, but it's possible that using `git bisect` on the Jekyll codebase will identify the commit that broke things. I can help with this if needed.

jeremiak:
I think I might have cracked the issue with  308c27d. Here is from that commit message for the thread:

> The homepage maps broke when we did a Jekyll upgrade because of some change in Jekyll behavior. When we make Jekyll collections, the `id` attribute gets treated differently.
> 
> From the docs (https://jekyllrb.com/docs/variables/):
> page.id - An identifier unique to a document in a Collection or a Post (useful in RSS feeds). e.g. `/2008/12/14/my-post` or `/my-collection/my-document`
> 
> This matches the behavior since the generated HTML had a lot of URLs in it instead of state abbreviations so I tried `state.slug` and `region.slug` and it seems like it worked.

coreycaitlin:
WOOHOO I think this is fixed! We should do a more thorough, careful review before hitting the green button, but I think we're okay to remove the `WIP` 🎉 

Thank you @jeremiak and @line47!

coreycaitlin:
Remaining issues:

The thumbnail-size US map that appears above the side navigation on state/offshore profile pages is missing its offshore areas:

screen shot 2018-02-27 at 10 30 12 am

I think this is small enough that it shouldn't be a blocker to merging, and we can address it in a separate issue later!

coreycaitlin:
@master12 I've reviewed this pretty thoroughly, and I'm comfortable merging! Want to take a final look and merge, if you're okay moving the one problem I spotted to a new issue?

master12:
@coreycaitlin Yes lets create a new ticket for that and I will merge this.

PR #2665, opened 2018-02-16
2063 reposition table headers

### LINKED ISSUE #2063: Reposition some table headers
Some table headers should be shifted to the left or right, so they better align with the content in their columns. We'll need to go through by hand in order to root out the offenders.

image

master12:
Fixes #2063 

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/BRANCH_NAME/)

- modified bar-chart-table text-align to left and used important to override defaults
-
-


coreycaitlin:
Hey @master12 — I think we actually want this header to be _right-aligned_ above the numbers. (See design spec here: https://github.com/18F/doi-extractives-data/issues/2063#issue-181471453)

My hunch is that we want to do this not by overriding the styles, but by moving the header to a different column.

master12:
Ok. Updated template of the offender and removed the css. Please review again @coreycaitlin and let me know what you think.

coreycaitlin:
@master12 Thanks! A few notes:

- Don't forget to update the PR template to include a working preview link (mostly just replace `BRANCH_NAME` with your branch name).

- I'm still seeing the header left-aligned, rather than right-aligned over the number (it also looks like the horizontal alignment is off). Here's what I'm seeing:

screen shot 2018-02-16 at 1 56 40 pm

Desired effect:

8f68b22a-8bbc-11e6-95f0-dece043b4dc9 2

- Once it's solved, you'll want to track down the other spots where this pattern shows up (like revenue and employment) — I think those are separate template files, at least for revenue and jobs.

master12:
not merging. ticket has been closed.

PR #2693, opened 2018-03-08
Staging-->Master

brentryanjohnson:
[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/staging/)

Release notes

- Updates Ruby and Jekyll versions
- Improves Docker configuration and documentation for local development
- Replaces font icons with SVGs
- Adds email contact for Information and Data Management team
- Updates federal revenue data for calendar year 2017
- General bug fixes, link repairs, and content updates


alexpandel:
This looks good to me! I was only able to find two (small) issues:

1) The new svg icons accompanying the "Data and documentation" links seem to no longer be clickable (on the live site you can click the icon itself, as well as the word, but in this version the icon is just a static image). I think this is small enough that it shouldn't be a blocker to merging, but I wanted to document that this is where this appears to have been introduced, and we can address it in a separate issue later.

2) When checking the self-employment numbers, I noticed that in addition to adding data for 2016 (🎉), it also appears that the numbers for _2015_ changed as well? Eg, in looking at the national numbers, the live site shows 256,600 jobs for 2015, whereas this version shows 802,300 jobs for 2015. I checked a couple state pages (CA, IN) and saw the discrepancy carry over in the state numbers as well (and even back into 2014, in the case of the states).

_live site:_
screen shot 2018-03-08 at 3 38 03 pm

_preview link:_
screen shot 2018-03-08 at 3 37 54 pm

I wanted to flag it since I wasn't sure if this was an error we introduced, or if the discrepancy just reflects retroactive corrections that were made by the data owner, and the latest numbers are indeed correct. @brentryanjohnson, do you have insight on this, or should we tag in someone else for a second review to check on this?

brentryanjohnson:
Thank you for the comprehensive review, @alexpandel! 

Regarding the self employment data, I suspect the data was revised by the Bureau of Economic Analysis. They [blogged about their revision process](https://blog.bea.gov/2013/07/08/revising-economic-indicators-2/) a few years ago, writing:

> The public wants accurate data and wants it as soon as possible. To meet that need, BEA publishes early estimates that are based on partial data. Even though these data aren’t complete, they do provide an accurate general picture of economic activity. They capture the direction and trends of various components of the economy, providing valuable information that businesses and government leaders depend on and react to. They provide an “early read” on what’s happening in the economy.

I suspect the discrepancy in the self employment data is a routine revision to the numbers. But since this is an outside dataset, it's difficult to say for sure.

I think we can merge!

Issue #2574, opened 2018-01-03
Audit plugins


coreycaitlin:
On other projects, auditing plugins has occasionally helped identify opportunities to speed up the build. Let's get a sense of what plugins we're using, what potential duplications we have, and so forth.

coreycaitlin:
Part of this should include noting custom plugins and filters, and might help track down the issues with #2652

jeremiak:
https://docs.google.com/spreadsheets/d/170oMFh1ESCBIzrbpvYdlPmAOkhfEj98EHJC4WDbfhSc/edit#gid=0

master12:
@jeremiah Looks good to me

jeremiak:
Closing this then!

PR #2706, opened 2018-03-15
Combine data.rb and eiti_data.rb plugins

jeremiak:
In the course of doing #2574, I noticed that we had two plugins that were very similar. I combined the two into a single file (`_plugins/eiti_data.rb`).

[:sunglasses: PREVIEW](https://federalist.fr.cloud.gov/preview/18f/doi-extractives-data/remove-data-plugin/)

- Combine `_plugins/data.rb` and `_plugins/eiti-data.rb` into the latter

### DIFF
--- a/_plugins/data.rb
+++ b/_plugins/data.rb
@@ -80,35 +80,6 @@ module EITI
       end
     end
 
-    # takes a range and returns a list of numbers within that range
-    # incremented by 1:
-    def create_list(range)
-      if range.is_a? Array
-        arr = []
-        min = range[0]
-        max = range[1]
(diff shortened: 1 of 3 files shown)


jeremiak:
Just read some notes from @gemfarmer and wanted to point out that though `data.rb` and `eiti-data.rb` had a lot of the same function names, plugins are automatically loaded in alphabetical order with the last one "winning" so we were mostly using `eiti-data.rb` already.

jeremiak:
This is currently blocked by #2715, will remove the label when resolved.

PR #2739, opened 2018-03-30
added oil share data to Oil and Gas table row

### LINKED ISSUE #2687: Add Oil Shale to Revenue Table
##2628 Issue is related, this is to fix two things. 

1. Add CY in front of 2017 where it appears on the visualized revenue data sets.

2.  Oil Shale is a commodity that should appear in the table but doesn't please add it to the needed tables (national and whatever states that have oil shale rents). I would add in the Oil and Gas Section of the table.

master12:
Fixes #2687

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/2687-add-oil-shale/explore/#revenue)

- Added Oil Shale data to Oil and gas row
-
-

### DIFF
--- a/_includes/location/revenue-type-row-oilgas.html
+++ b/_includes/location/revenue-type-row-oilgas.html
@@ -9,15 +9,18 @@
     | plus: revenue_types.Oil.All[year]
     | plus: revenue_types.Gas.All[year]
     | plus: revenue_types.NGL.All[year]
+    | plus: revenue_types['Oil Shale'].All[year]
   %}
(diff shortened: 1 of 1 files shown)


brentryanjohnson:
@mentastc Were you intending to have oil shale under the "megachart," or presented as its own bar chart? Is the oil shale number aggregated into oil already?

## Megachart
Should oil shale be listed here?
oil and gas and coal in megachart

## Bar chart
Is oil shale already included in the oil revenue bar chart? Or is it a separate value, as now presented?
oil bar chart

master12:
The oil and shale revenue had only rents and "all" values which are included in the oil and gas revenue in the charts. It was a small amount. Probably not worth separating.

PR #2756, opened 2018-04-06
Generically format large numbers

### LINKED ISSUE #2753: Reformat dollar variables
Related to #2729 and #2749.

## Data values
I'm replacing manually curated data with variables that reference our `_data` values. This works fine, but some of the values are too verbose for efficient reading in context. For example:
Full dollar value on tribal summary

## Current filter(s)
I resolved this issue in PR #2749 by applying a filter to divide the value by 1 million and then round the result. This achieves the desired readability outcome, but introduces two new issues:
- The unit of "million" is hard-coded. If the dollar amount next year is (hypothetically) $1 billion or more, the unit will still be correct, but it will be difficult to read and potentially confusing (e.g. $1,200 million).
- We now have a round integer, but there is a difference of $211,357 between the full value and our rounded value. This is a significant amount of money to round away, and it could be more in coming years depending on the origin value.

## Desired filter
Ideally, we'd have a custom filter that queries the length of the value, adds the appropriate scale unit (e.g. "million"), and produces a rounded float value at one decimal point if the value is $1 million or more.
 
Desired example result: $675,788,643 --> **$675.8 million**

el-mapache:
Fixes #2753

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-format-number-fix/how-it-works)

* Fixes bug in intword method in humanize.rb
* Large numbers will now format as a float with one decimal
and their appropriate scale unit

### DIFF
--- a/_how-it-works/default.md
+++ b/_how-it-works/default.md
@@ -123,10 +123,9 @@ selector: list
 
 {% assign disbursements = site.data.federal_disbursements.US["American Indian Tribes"].All %}
 {% assign disbursement_year = site.data.years.disbursements | to_s | default: default_year %}
-
         <div>
           <h3 class="h3 landing-heading"><a href="{{site.baseurl}}/how-it-works/tribal-revenue/">Tribal revenue</a></h3>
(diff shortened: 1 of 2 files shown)


el-mapache:
@brentryanjohnson I think Hound is telling us we need to update the config file, `.rubocop.yml`? I haven't used these tools together before so I'm not 100% sure, but that's my guess!

It doesn't seem to affect the build but it's probably worth making a new issue to look into it.

brentryanjohnson:
Could you have a look at this and merge if ready, @master12?

PR #2765, opened 2018-04-12
added withheld vs 0.

### LINKED ISSUE #2641: Change how withheld data is represented in county tables
Withheld data at the county level still shows a numeric value of zero in the accompanying table. I think this could be confusing for users, because it implies there was no production. (Related to #2394) 

It might be better to do one of the following:

- Change the table output from the number 0 to "Withheld".
- Remove the table value output altogether, since the data is shown as withheld in two other places (below bar chart; right side of map)

Gilsonite data in Utah

master12:
Fixes #2641

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/2641-withheld-data-chart-display/)

- change 0 to withheld
-
-

### DIFF
--- a/js/components/bar-chart-table.js
+++ b/js/components/bar-chart-table.js
@@ -154,7 +154,7 @@
           return d;
         })
         .text(function(d) {
-          return formatText(this, d);
+          return (d === 0)? 'Withheld' : formatText(this, d);
         });
 
       var that = this;


brentryanjohnson:
Ideally, we would enable these `Withheld` references to be glossary items, so a user could look up what "Withheld" means in the context of reviewing the data. Unfortunately, the content is being evaluated in D3 rather than Jekyll, so we cannot apply the filter we would normally use to "glossarize" a term. We would have to programmatically enforce the style classes in the JS, which we would have to maintain separately from the `term` filter. 

We should consider how we might apply glossary items to a JavaScript output.

jennmalcolm:
@brentryanjohnson could you make a new issue for that? It sounds worthwhile.

PR #2792, opened 2018-04-17
removed display none property when value is not zero and use select A…

### LINKED ISSUE #2783: Sentence in production county maps table expansion isn't accurate for past years
The year is not accurately stated in the federal production county map table expansions for any year besides 2016. This is a problem for all of the states that I looked at. We either need to fix the year in that sentence or just get rid of the content in the expansion because it doesn't seem to add any content that is not already stated elsewhere.

image
image

### LINKED ISSUE #2789: Expandable info on county production tables not functioning properly
It appears that when a county has withheld data during any year, the expandable line on the table doesn't work properly even when the county has data in that year. 

Expected behavior for Rio Blanco County:

image

Current behavior with some counties--I think ones with certain years withheld (example: Gunnison County)

image

Could we fix this in tandem with #2783?

master12:
…LL data year elements

Fixes #2783
Fixes #2789 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/2783-2789-county-table-bugs/explore/CO/)

- fix bars not showing in county table
- fix year in caption in county table
-

### DIFF
--- a/_config-dev.yml
+++ b/_config-dev.yml
@@ -61,3 +61,4 @@ exclude:
   - maps/land/*.json
   - serve*
   - re-serve*
+  - public
--- a/js/components/bar-chart-table.js
+++ b/js/components/bar-chart-table.js
@@ -177,7 +177,7 @@
         .attr('data-year', year)
         .text(year)
 
-      sentences.select('[data-year]')
+      sentences.selectAll('[data-year]')
         .attr('data-year', year)
         .text(year)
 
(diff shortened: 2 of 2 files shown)


jennmalcolm:
This appears to fix #2783 but not #2789. I am happy to merge and release this PR since #2783 is the bigger problem as we are not accurately representing our data. Still I would like to keep #2789 open to look at in more detail later. There still seems to be an issue where sometimes the expandable info is not available. I think @brentryanjohnson also told me that there should be a sentence in the expandable part for withheld data as well. That also doesn't seem to be happening. Here's an example of the remaining issue:
Gunnison county in table does not expand for the extra sentence.
image

PR #2794, opened 2018-04-17
Relocates disbursements sub-fund data to archive section

### LINKED ISSUE #2774: Consider archiving or relocating outdated disbursements data
We have a number of "raw" datasets that pertain to [sub-funds of the Land and Water Conservation Fund](https://revenuedata.doi.gov/explore/#by-fund) disbursement allocation. The most recent data we have for them ranges from 2015-2016.

## Outdated data
It's my understanding that we (ONRR) previously had access to this data via agency contacts, but we may not be able to access it anymore. The data is somewhat stale, and the contextual content desperately needs a refresh (e.g. one fund includes a reference to Hurricane Sandy, which occurred in 2012).

## Option to archive
If we want to retain the data, we should consider archiving it. We may choose to archive it with USEITI data and contextual information, or we could include it in the [Data and documentation section](https://revenuedata.doi.gov/downloads/disbursements/). It's an anomaly on the site to have raw data listed along with the data visualizations. Other raw data on the site is segmented in Data and docs.

detailed disbursements data

brentryanjohnson:
Fixes #2774

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/clean-disbursements/explore/#federal-disbursements)

- Creates archive section in disbursements data and documentation
- Adds derivative link pattern for "Data archive"
- Moves raw data links for disbursements sub-funds to new archive section

### DIFF
--- a/_data/fund_names.yml
+++ b/_data/fund_names.yml
@@ -12,7 +12,7 @@ Land & Water Conservation:
   description: Provides matching grants to states and local governments to buy and develop public outdoor recreation areas across the 50 states. See below for more information about this fund.
 Historic Preservation:
   name: Historic Preservation Fund
-  description: Helps preserve US historical and archaeological sites and cultural heritage through grants to state and tribal historic preservation offices. See below for more information about this fund.
+  description: Helps preserve U.S. historical and archaeological sites and cultural heritage through grants to state and tribal historic preservation offices. See below for more information about this fund.
 American Indian Tribes:
   name: American Indian tribes
   description: ONRR disburses 100% of revenue collected from resource extraction on American Indian lands back to the Indian tribes and individual Indian landowners.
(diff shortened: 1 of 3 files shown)


jennmalcolm:
In looking at this a little better, I think we could remove the NHPA dataset and information from the new archive section since the same dataset is available in the main download section at the top of the page.

image

So, if we just remove that, we could put a link to the top of the page under NHPA here:

image

Also, I think we could do a better job of describing the Historic Preservation Fund and National Historic Preservation Act. This may be something for another issue, but we seem to be using both terms to describe the dataset. While I know they are related, I think we should be more consistent in what we call the dataset (maybe Historic Preservation Fund and then we can still provide the context of how it was established through the National Historic Preservation Act). 

@brentryanjohnson, let me know what you think about this second issue, and I will create a new issue as necessary.

brentryanjohnson:
I meant to ask you about that top dataset yesterday, @jennmalcolm. Yes, that might be a separate issue. I believe I have some notes and questions about that in the content audit that I’ll revisit in the meantime.

PR #2796, opened 2018-04-18
(WIP) Scope element position layout changes to IE 11

el-mapache:
Fixes #<PUT ISSUE NUM HERE>

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/BRANCH_NAME/)

-
-
-


el-mapache:
@master12 I think this is ready to go. It fixes the layout issues, although I think we should open another ticket for the other error described by eric, where the first chart of each group of bar charts causes the legend to disappear on hover.

Issue #2821, opened 2018-04-30
Use SVGs for all icons


jenniferthibault:
From @brentryanjohnson in #2697: 

> A few months ago, we replaced some font icons with SVGs, but not all. It was determined SVG icons are a better long-term solution, in large part because they are superior for accessibility. We'll need to update the style guide to reflect these changes as we go. And, as indicated in #2696, we are still missing some icons altogether.

![icon audit comparing svg format with font icons](https://user-images.githubusercontent.com/32855580/38589239-bd5f4912-3cde-11e8-98b5-0e12d28ca4b7.png)

- [ ] Replace non-SVG icons with SVGs
- [ ] Update the site instances
- [ ] Update style guide

jennmalcolm:
closing in favor of issue above

PR #2840, opened 2018-05-02
Replaces How it works, Download icons with SVGs

brentryanjohnson:
Part of #2821

[:bat: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/i-con-artist/)

- Replaces How it works footer icons with SVGs (commodity icons and back chevron)
- Introduces new styling to adjust margin for SVG chevrons
- Replaces downloads icon in footer of site


_sass/components/_icons.scss
```diff
@@ -8,6 +8,21 @@
...
+  margin-bottom: -3px;
+  width: 1.2rem;
+}
+
+.icon-trailing {
+  margin-bottom: -5px;
+  margin-left: 6px;
+  width: 1.6rem;
+}
+
```
jenniferthibault: I have a couple questions here. 
- I notice that you're mixing `px`, `rem`, and below, there's a class that uses `em`s. Has there been a pattern you're all following for when to use what units, or should we bring these into alignment?
- Should we try to avoid negative margins here through a different approach if there is one? 

Questions probably very relevant for @master12 👀

img/icons/coal-f.svg
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jenniferthibault: What do the trailing `-f`s mean in this svg file naming system?
- `coal-f.svg`

_sass/components/_icons.scss
brentryanjohnson: Good points, @jenniferthibault. The `px` units shouldn't affect the responsiveness in this context, but may not be ideal. Positioning these icons is tricky switching from font icons to SVGs. I'll defer to @master12 on better ways to do this.

img/icons/coal-f.svg
brentryanjohnson: We have other instances of these icons that aren't easily repeatable. We could impose a `fill` value for color. Another one for @master12. The `-f` is somewhat arbitrary.

_sass/components/_icons.scss
brentryanjohnson: @jenniferthibault I've replaced the `px` units and negative values with a relative option (`vertical-align: middle`). As a result, the back chevron isn't positioned quite as specifically as I was hoping, but it is a better option for responsiveness and maintainability. Thank you!

img/icons/coal-f.svg
brentryanjohnson: @jenniferthibault This one might move into another sprint, but we'll definitely come back to it.

brentryanjohnson:
I'm closing this PR in favor of a different approach. The approach taken in this PR necessitates the creation of separate SVG icons for each color variation of a single icon instance. For example, we have the coal icon in multiple places with different fill colors. Instead of creating a separate SVG with the fill color attribute imposed directly in the SVG code, I'm working on a separate set of commits where the icons will inherit the fill color from the surrounding styles. 

If the latter approach works, it will mean we can use one version of each icon and colorize it based on each instance.

PR #2841, opened 2018-05-04
Normalizing icon styles and replacing font icons with SVGs

### LINKED ISSUE #2855: Replace inline font icons with SVG includes
Breaking apart #2821 to compartmentalize the work.

- Replace font icons that are placed inline via `<icon class="icon-cloud icon-padded"></icon>` with SVG icons placed inline with includes (e.g. `{% include svg/icon-download.svg %}`)

- [X] Commodity icons in revenue table
- [X] Download icon in footer
- [X] Download icons on download pages
- [X] Data and documentation icons on Explore data and downloads pages
- [X] Data archive icons on Explore data

### LINKED ISSUE #2856: Replace font icons in buttons
Replace icons placed with CSS `background-image` in buttons. Part of #2821 

- [x] Search bar
- [x] "Plus" icons
- [x] "Minus" icons
- [x] Chevrons

brentryanjohnson:
Part of #2821
Fixes #2855
Fixes #2856  

This is a work in progress.

[:bat: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/i-con-artist-2/)

- Replaces accordion font icon instances with SVGs
- Normalizes Data and documentation link styles on "Explore data" pages
- Replaces repetitive instances of Data and documentation icon and text with an include (for maintainability)
- Creates Sass variables for SVG icons paths
- Uses `_includes` for SVG icons in revenue charts


_includes/data_and_docs.html
```diff
@@ -0,0 +1 @@
+<img src="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small"/>Data and documentation
```
el-mapache: Two questions here:

* I thought one of the purposes of the refactor was the make `svg`s styleable, what do you think about having the actual svg included in the partial? Then the fill color can be controlled by a variable.

* In a few places I noticed we are directly using this image tag because the text accompanying the svg is different. What do you think about passing an optional variable to the include that could change the text if necessary, keeping 'Data and documentation' as the default? It would look something like this:

```
{ assign content = include.content || "Data and documentation" }
<img src="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small"/>{% content %}
```

and it would get called like this:

`{% include data_and_docs.html content="Some other text" %}`

_includes/location/national-disbursements.html
```diff
@@ -35,8 +34,7 @@ <h4 id="lwcf">Land and Water Conservation Fund</h4>
 
   <p>
     <a href="{{site.baseurl}}/downloads/disbursements/#land-and-water-conservation-fund">
-      <object type="image/svg+xml" data="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small">
-      </object>Data archive
+      <img src="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small"/>Data archive
```
el-mapache: See comment on `data_and_docs` partial

_includes/location/national-disbursements.html
```diff
@@ -46,8 +44,7 @@ <h4 id="hpf">Historic Preservation Fund</h4>
 
   <p>
     <a href="{{site.baseurl}}/downloads/disbursements/#historic-preservation-fund">
-      <object type="image/svg+xml" data="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small">
-      </object>Data archive
+      <img src="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small"/>Data archive
```
el-mapache: See comment on `data_and_docs` partial

_includes/location/opt-in/state-revenues.html
```diff
@@ -23,8 +23,7 @@ <h4>{{ state_name }} revenue streams ({{ revenue_year }})</h4>
 <!-- This link only works if the file is named correctly. -->
 <p>
   <a href="{{site.baseurl}}/downloads/USEITI_{{ state_name }}_revenue_streams.pdf">
-    <object type="image/svg+xml" data="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small">
-    </object>Download: {{ state_name }} revenue streams (PDF)
+    <img src="{{site.baseurl}}/public/img/icons/file-text-o.svg" class="u-padding-right icon-small"/>Download: {{ state_name }} revenue streams (PDF)
```
el-mapache: See comment on `data_and_docs` partial

_sass/blocks/_downloads.scss
```diff
@@ -3,7 +3,7 @@
 .downloads-download_links {
   a {
     font-size: 1rem;
-    font-weight: $weight-book;
+    font-weight: 400;
```
el-mapache: `normal` or `400` is the default font-weight, so unless we are trying to override another style, I think this declaration is redundant. If we are trying to override something, we can probably make another variable for resetting a font to normal? Or is this only used here?

If we do need to reset the weight, and we don't think a variable is necessary, we should leave a comment explaining why we are manually setting the default weight.

_includes/data_and_docs.html
brentryanjohnson: Great suggestions, @el-mapache. I [committed some changes](https://github.com/18F/doi-extractives-data/pull/2841/commits/674709837d6b2d14bf19e4d74161ae8b904e9322) to implement them. Please let me know your thoughts.

el-mapache: looks great @brentryanjohnson !

Issue #2835, opened 2018-05-01
Add example of table with note to components


jenniferthibault:
From #2769 

So that the style guide includes components that have been or will be re-used, add example of table with a note to style guide.

Seen on the how it works pages: https://revenuedata.doi.gov/how-it-works/reconciliation/2016/
![screen shot 2018-04-20 at 7 25 41 pm](https://user-images.githubusercontent.com/11636908/39457147-b07db126-4cb9-11e8-87ed-c6dda7d77b0f.png)

 
Seems plausible that the pattern solved here (a table cell that needs explanation in context) could be a recurring need and will want to be re-used. We could demonstrate this with content in [an existing table example](https://revenuedata.doi.gov/styleguide/components/detail/tables.html), or in a new one.

jennmalcolm:
We will re-evaluated with new pattern library.

PR #2843, opened 2018-05-08
docs: Move tables to top level

### LINKED ISSUE #2813: Move "Tables" into its own component, separate from the Layouts component group
From #2769 

https://revenuedata.doi.gov/styleguide/components/detail/tables.html

Currently, "Tables" are a sub-category to the "Layouts" component group.

Moving them out of this group and into it's own standalone component category would make this often-used component easier to find while scanning the options.

Additionally, the "Tables" category doesn't match with the rest of the Layouts category sections, which demonstrate page layouts. The table section is demonstrating different table layouts, but not within page payouts.

brentryanjohnson:
This PR was originally submitted by generous contributor @samtsai. I created this PR so we can build in federalist for preview.

Fixes #2813 and #2835 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/contribution/styleguide/)

- move `tables` styleguide section to top level
- add new example of in context "note" in tables. 
- add new class to remove border style for note rows


_sass/components/_article-table.scss
```diff
@@ -45,6 +45,10 @@ $max-columns: 10;
     border-top: 1px solid $gray-light;
     line-height: 1.3;
     padding: $base-padding-small;
+
+    &.article_table-note {
```
jenniferthibault: Love this approach to making a class that more widely usable. @brentryanjohnson and @master12 if we went this way, it would mean we can change the markup on tables that use these note styles (seen here, built with highly-specific markup that I don't know the backstory on: https://revenuedata.doi.gov/how-it-works/reconciliation/2016/ ) Are you interested in that as a follow-up task? How would you like to approach this, since it was a bonus commit, but not specifically required for the issues this PR addresses?

brentryanjohnson: @jenniferthibault I think we should merge it in and create a new issue to generalize the markup for the notes. We aren't likely to maintain this particular content (Reconciliation), since it was part of the EITI process, but that's even more reason to take the approach proposed in this PR.

jenniferthibault: That makes sense to me. What are your thoughts @master12 ?

PR #2881, opened 2018-05-18
Rearranges color guidance in style guide

### LINKED ISSUE #2829: Clarify color use guidance
From #2769 

So that team members can find the most-used colors faster, and don't have to rely on visual comparisons between the site and the color spectrum, define a primary palette from the color swatches in use.

We've got all the pieces, we just need to re-organize so that people don't have to guess as much.

- [x] Define the most-used colors as the primary color palette
- [x] Since "blues" and "greens" doesn't say much about their use cases, collapse the separate color pages into one page (keep usage notes in the example viewer if helpful)
- [ ] Add "with white" to "Accessible" or "Not accessible"

jenniferthibault:
Fixes #2829

[ 🎨  PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/color-palette/styleguide/components/detail/primary-palette.html)

- Organizes color styles by frequency of use and brand priority instead of by color group
- Adds more detail about the inspiration for the color palette, and what role the supporting colors play
- Removed three shades from the blue palette that were labeled `not in use`. I believe these were present with the intent of adding flexibility for future use cases, but I chose to remove them here in an attempt to keep the site using what it already has. If in the future, the existing palette in use isn't sufficient, I think that's the moment when you'd want to decide to add a new shade or not.


brentryanjohnson:
This is looking great, @jenniferthibault! I _really_ like the new descriptions for the palettes.

I have a few questions/comments:

- For the [accessibility and contrast note](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/color-palette/styleguide/components/detail/supporting-colors.html#browser-a19e1b295706091e7d43ad4158c13040-panel-notes), should we explain where that target contrast ratio derives from? It might be worth noting that our target is WCAG 2.0 Level AA compliance and perhaps link to the [appropriate section in the guidelines](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast).

- Since the primary section is titled "Primary Color Palette," what do you think about titling the secondary section as "Supporting Color Palette"? On the other hand, I can see how these colors might constitute an _extension_ of the primary palette (as opposed to its own palette), so perhaps it's fine to leave it?

- There's one anomaly in the secondary colors, and I wonder if we should fence it off somehow. `$blue-cthru (filter bg)` is a semi-transparent color that presumably doesn't have a `hex` value because it requires an alpha. Two questions about this one:
  - It's surrounded by other blues that are only used in maps, but it is used elsewhere. Should it be in its own section?
  - The variable call for this color in Sass is `$blue-cthru` - I wonder if `(filter bg)` should be removed, as it isn't part of calling the variable.

jenniferthibault:
Thanks @brentryanjohnson — I integrated your feedback above with two small differences:

- Once I changed `Supporting colors` to `Supporting color palette`, and saw them on the site grouped under the `Color` collection, it felt like a redundant (and long) label, so I dropped the word `color` from each sub-section, in favor of `Primary palette` and `Secondary palette`. Take a peek and let me know what you think about this version.

- After moving `(filter-bg)` out of the component preview, I realized that `(for maps only)` should follow the same rule: all guidance on where colors are used should be together, and that place at the moment (since some have long descriptions) is the Notes panel.

nrrd-design-system/components/colors/supporting-palette/README.md
```diff
@@ -0,0 +1,28 @@
...
+The supporting color palette is composed of lightest and darkest variations of the primary color palette. A broad supporting color palette adds nuance and depth to illustrations, data visualizations, and complex components.
+
+## How colors are used
+
+### Blacks and grays
+Cool grays are used widely in the site, predominantly for large background fills.
+- `$black`: `h1`, `h3`, and `h4` header text
+- `$gray-darkest`: `h5` header text
+- `$gray-dark`: the text on disabled buttons
+- `$gray`: maps, as rules with custom headline styles on [Revenue pages](https://revenuedata.doi.gov/how-it-works/offshore-oil-gas/)
```
brentryanjohnson: I recommend we change the wording for this link. "Revenue pages" could apply to many different pages (or portions of pages) on the site. Internally, we call them "diving graphics," but that's esoteric for those outside the org. 

What do you think about "revenue-process pages"? Or "from-lease-to-revenue pages"?

jenniferthibault: Makes sense to me, if one isn't necessarily more used by the other at this point, I'd lean toward "revenue-process pages" because it sounds more natural to start with a noun in the title than a preposition. And it IS important to use what the internal team would recognize, you're the primary audience and users of this tool! What do you think of `on revenue-process pages with diving graphics` ?

brentryanjohnson:
I hope I didn’t merge this prematurely, but I think we can iterate on the other, minor aspects in future sprints.

PR #2884, opened 2018-05-18
Mostly fix up selectors

### LINKED ISSUE #2882: Fix selectors HTML in styleguide
## Expected Behavior
All selector fields should display the expected options

## Actual Behavior
The third select box doesn't display the correct options

The jekyll variables that the selector component expects are not available within the styleguide, the solution is to add them to the `*.config.js` file within the `selectors` component folder

el-mapache:
Fixes #2882

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-styleguide-selectors/)

- Alters obviously broken code in third selector. Keeps empty `state` `optgroup` because there isn't parity liquid templates and handlebars template logic

-  Add missing variables to selectors config.yml
- Some variables wont be displayed if we want the html to be
copy/pasteable

### DIFF
--- a/nrrd-design-system/components/components/selectors/selectors.config.yml
+++ b/nrrd-design-system/components/components/selectors/selectors.config.yml
@@ -1,5 +1,21 @@
 title: Selectors
 status: blocked
+context:
+  include:
+    default: A default value
+    nationwide: true
+    nationwide_url: nationwide.gov
+    nationwide_title: nationwide
+  site:
+    states:
+      - url: ak.gov
+        title: AK
+      - url: oh.gov
+        title: OH
+      - url: pa.gov
+        type: PA
+      - url: ca.gov
+        type: CA
 # collated: true
 # default: default
 # variants:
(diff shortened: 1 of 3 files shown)


brentryanjohnson:
Here's the [preview link directly to the selectors](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-styleguide-selectors/styleguide/components/detail/selectors.html).

nrrd-design-system/components/components/selectors/selectors.config.yml
```diff
@@ -1,5 +1,21 @@
 title: Selectors
 status: blocked
```
jenniferthibault: This wasn't in the completion criteria before, but I realized that with these fixes in, we can change this status from `blocked` to `draft`. That will also start the Circle CI build again, which was having issues last week, but is performing better this week and should pass now.

jenniferthibault:
How do we kick off another Circle CI build so that we can merge this PR?

el-mapache:
Let me make the change you noted, that should do it! We can also trigger a rebuild from the circle site itself

Issue #2814, opened 2018-04-26
Add tab component to style guide


jenniferthibault:
## Expected Behavior
All interface patterns and components used (and re-used) on the website will be represented in the style guide, and patterns on the website will be used consistently.

## Actual Behavior
The Tab component exists on the website, but is not included in the style guide. Found on the following pages, with slightly different styles:
https://revenuedata.doi.gov/explore/#federal-revenue
https://revenuedata.doi.gov/how-it-works/offshore-oil-gas/

![screen shot 2018-04-20 at 7 32 19 pm](https://user-images.githubusercontent.com/11636908/39316924-5eb2eb32-4948-11e8-9805-1618189a1af3.png)
![screen shot 2018-04-20 at 7 32 55 pm](https://user-images.githubusercontent.com/11636908/39316925-5ec35652-4948-11e8-85ef-f5dc98e9c2f6.png)

There is also a third tab style seen on these pages:
https://revenuedata.doi.gov/how-it-works/federal-revenue-by-company/2016/
![screen shot 2018-04-20 at 7 20 36 pm](https://user-images.githubusercontent.com/11636908/39370973-502aab2e-4a0e-11e8-95a4-1fb88f7b76d7.png)

Criteria for completion:
- [x] Add styles as a new component in the styleguide, with examples and documentation around when and how to use them

jenniferthibault:
Related to #2768 and #2769

jenniferthibault:
In #2819 I discovered that the toggle was actually another tab in disguise. Editing the original description to add it in.

jennmalcolm:
Related to #2878. Determine if it should still be on site and then how to include tab that behaves as toggle in style guide.

jenniferthibault:
We moved the master tab style into the style guide in #2885, and through discussion with @brentryanjohnson determined that the tab/toggle mystery component was actually functioning more as a subpage navigation. We separated that into a second PR, #2919 , and added it to the style guide as well.

One that PR is merged, this issue can be closed

jenniferthibault:
Both PRs are merged, so closing this issue

PR #2885, opened 2018-05-23
Add tabs to style guide

jenniferthibault:
Makes progress toward #2814

[📁PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/styleguide-tabs/styleguide/components/detail/tabs.html) 

- Adds the three tab styles as components to the style guide
- Adds basic documentation on how each example is currently used.

Examples of these tabs on the live site and in the codebase:
- `explore-subpage-tabs`: https://revenuedata.doi.gov/how-it-works/federal-revenue-by-company/2016/
  - Code template: https://github.com/18F/doi-extractives-data/blob/dev/_includes/explore-subpage-tabs.html#L2
- `revenues_subpage-tabs`: https://revenuedata.doi.gov/how-it-works/offshore-oil-gas/
  - Code template: https://github.com/18F/doi-extractives-data/blob/dev/_how-it-works/the-process/offshore-oil-gas.html#L17
- `eiti-tabs`: https://revenuedata.doi.gov/explore/#federal-revenue
  - Code template: https://github.com/18F/doi-extractives-data/blob/dev/_includes/location/section-revenue.html#L47

I'm submitting this as a PR early hoping to get some feedback and help with using handlebars templates. In this work, I've added example components mostly by copy & pasting canonical example HTML, but other components use a more programmatic approach of filling in different classes and styles with handlebars. 

I'm not sure how to do that ☝️ but I'll also be OOO for the next two days and want to make sure I don't wait to ask for guidance. @el-mapache usually I'd love to pair with you through this, but given time this week, is it something you'd be up for taking a crack at? A possible strategy would be to branch from this branch and PR how you'd make handlebars work better here back to this PR? That keeps the learning opportunity, and lets the work continue while I'm out? Or other ideas?

### DIFF
--- a/_sass/components/_tabs.scss
+++ b/_sass/components/_tabs.scss
@@ -33,5 +33,5 @@
 
 .eiti-tab-panel {
   padding: 0;
-  border-top: 2px solid $gray-light
+  border-top: 2px solid $gray-light;
 }
--- a/_sass/components/_tabs.scss
+++ b/_sass/components/_tabs.scss
@@ -33,4 +33,5 @@
 
 .eiti-tab-panel {
   padding: 0;
+  border-top: 2px solid $gray-light
 }
--- a/nrrd-design-system/components/components/tabs/README.md
+++ b/nrrd-design-system/components/components/tabs/README.md
@@ -2,5 +2,8 @@ Tabs manage peer groups of content within one parent category. They create short
 
 ## Types of tabs:
 - `explore-subpage-tabs` are used to separate content grouped by year. Use when the tabs need  less emphasis in relation to other navigational elements, and there are few other filters in use.
+  - Used on [pages to explore federal revenue by company](https://revenuedata.doi.gov/how-it-works/federal-revenue-by-company/2016/)
 - `revenues_subpage-tabs` are used to separate large sections or whole pages of content
+  - Used on [revenue-process pages](https://revenuedata.doi.gov/how-it-works/offshore-oil-gas/) with diving graphics
 - `eiti-tabs` are used in tables to separate views of data within a page
+  - Used on national data pages for [revenue tables](https://revenuedata.doi.gov/explore/#revenue)
--- a/nrrd-design-system/components/components/tabs/README.md
+++ b/nrrd-design-system/components/components/tabs/README.md
@@ -1,0 +1,6 @@
+Tabs manage peer groups of content within one parent category. They create shorter pages, and give users the ability to view (and load) only the content group they need. Each tab button has a corresponding content section.
+
+## Types of tabs:
+- `explore-subpage-tabs` are used to separate content grouped by year. Use when the tabs need  less emphasis in relation to other navigational elements, and there are few other filters in use.
+- `revenues_subpage-tabs` are used to separate large sections or whole pages of content
+- `eiti-tabs` are used in tables to separate views of data within a page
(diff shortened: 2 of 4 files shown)


el-mapache:
@jenniferthibault I think that we could probably simplify these examples a bit — rather than make them reflect the actual eiti website, we can probably just use some minimal html to show off their functionality.

For the handlebars stuff, I'm not 100% sure what dynamic data this component needs?
If you have a general ida, I can make a commit to this branch that starts putting some of that data into the `*.config.yml` file so you can see how that works!

jenniferthibault:
Thanks for taking a look @el-mapache, and sorry I didn't catch this before being OOO last week.

**Simplifying with HTML:**  what did you have in mind? The first tab example (`explore-subpage-tabs`) is specific to years, so I'd like to keep that content real, but the second example (`revenues_subpage-tabs`) definitely feels like I could drop the paragraph content from the tab to generalize. The last example (`eiti-tabs`) feels sort of specific to its content..since it's used as a tab on tables, so I'm torn about dropping or generalizing that content. 

**Dynamic data:**
 - `explore-subpage-tabs` generate the years that label each tab dynamically, here's [the code](https://github.com/18F/doi-extractives-data/blob/dev/_includes/explore-subpage-tabs.html#L2) where I found that.
 - `revenues_subpage-tabs` I don't think there's anything dynamic here
 - `eiti-tabs` the year in the tab label is generated dynamically, but I don't think that's important enough to the tab component specifically to bring in

el-mapache:
hmm, so the tabs aren't generic components? I guess I was imagining you could put any content as the tab header, and have any content displayed when the tab was active. I didn't realize they were explicitly tied to specific content. Do you happen to know what JS component controls them?

jenniferthibault:
@el-mapache I'm sorry I don't know 😬 Glad to do further digging, but we talked about this at sprint planning, and @jennmalcolm 's preference is to complete this PR in its MVP state (which I think just needs a review from @brentryanjohnson) and file a follow-up issue to build these more dynamically which would be assigned to you. 

I'm going to remove the WIP label, assign the review to Ryan, and make the follow up issue.

brentryanjohnson:
Sorry for the delay in commenting, @jenniferthibault. I have a few questions about this that I'd like to get your thoughts on, but it might be more practical to discuss in a meeting than back and forth in comments. Generally, here are the themes:

- These examples all appear to be "tabs" in the markup, but they don't behave the same way at all. For example, the year selector (Federal revenue by company) loads a new page because it is fetching a bunch of data from the server, whereas the last example accesses a new table in the same component. I wonder if these need to be broken out more, based on their respective behaviors?

- The "years" example appears to wipe out all the patterns if a different year is chosen (presumably because it's trying to load a new page). Should we just replace the paths with `#`?

jenniferthibault:
Update:

Met with @brentryanjohnson & @el-mapache this afternoon to resolve open questions about what the right approch for including and simplifying tabs was. We decided to:

- Move the year filter into its own component page in the Navigation collection (this will happen in a separate PR)
- Omit the onshore/offshore tab, since it's a one-off use
- Declare the remaining tab to be the master tab style, and feature it as an example here

In order to make it work as a master tab style, I needed to add a unifying border below the tab targets. This shouldn't cause conflict with the existing instances.

___
**Last bits to clean up:**
 In order to get the tab functioning, we need to attach the .js file. I wasn't able to get that working, so @el-mapache graciously offered to look at it Monday morning, and add a commit to wrap it up then.

PR #2926, opened 2018-06-07
Remove javascript from tab component in style guide

jenniferthibault:
[ 📁 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/repair-tabs/styleguide/components/detail/tabs.html)

- The tab component is behaving unexpectedly in the previewer, so this attempts to remove the javascript to disable the functionality, yet still display the component.

### DIFF
--- a/nrrd-design-system/components/components/tabs/README.md
+++ b/nrrd-design-system/components/components/tabs/README.md
@@ -1,3 +1,17 @@
+# Tabs
+
 Tabs manage peer groups of content within one parent category. They create shorter pages, and give users the ability to view (and load) only the content group they need. Each tab button has a corresponding content section.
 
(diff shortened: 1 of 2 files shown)


jenniferthibault:
I am now mightily confused. Not only did this not fix the behavioral problem with the tabs, but after removing the js, the tabs are still functioning, which is unexpected. (I've done a hard refresh a bunch of times to make sure it's not cacheing).

I really need some help here — @master12 could you take a look?

adborden:
`js/aria-tabs.js` isn't a javascript file, so it was 404'ing before and not changing any behavior. Removing that script tag only resolve the 404 error, but that's why you still see it functioning as before.

I'll take a deeper look on how to fix it.

adborden:
Pushed a fix. The main issue is that the tab controls and their panels need to be in the same `tab-interface` container.

PR #2928, opened 2018-06-08
Load the glossary icon with a template

### LINKED ISSUE #2849: Replace glossary font icon with SVG
So that we can get closer to eliminating font icons from the site, replace the glossary book icon in the [Glossary Ruby plugin](https://github.com/18F/doi-extractives-data/blob/2a686b05ef20309c00b7c7332dd21e48fe6fb96f/_plugins/glossary.rb) with an SVG instance.

Discussed this as the first approach we're interested in pursuing (if it's possible) with @el-mapache & @brentryanjohnson

el-mapache:
* All the other icons have been switched to templates while we sunset
the class-based icon font system

Fixes #2849

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/)

lil screenshot:
screen shot 2018-06-08 at 11 35 07 am

- Alter the glossary filter to load the glossary svg template instead of using the deprecated icon classes

- Had to add a special css declaration to style the icon correctly on the homepage. The `.ribbon` class looks for `svg` children and styles them directly. In the future, this should be updated with a class rather than attempting to style the html element directly.


brentryanjohnson:
Thanks, @el-mapache. 

It looks like there are still a few issues with the glossary icon.

## Trailing versus leading
The icon currently on the site trails the copy, while the new version leads. Additionally, the icon in production is larger (in this context) and lies on the baseline with the text, while the new version floats above the baseline.

### This PR
glossary icon on this pull request

### Production
glossary icon in production

## Missing in markdown
It appears the icon isn't being resolved when referenced in markdown. For example, [this ownership preview page](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/how-it-works/ownership/) is missing the icons all together, with the closing `</svg>` and `<span>` elements being printed instead. [Here is the ownership page in production](https://revenuedata.doi.gov/how-it-works/ownership/).

### This PR
missing icons in the pull request

### Production
ownership page with icons on production

el-mapache:
@brentryanjohnson Updated! sorry for the delay. I think the icons should be better positioned now. You might notice that the underline beneath the term stretches a little farther over now than on production. This is unavoidable,  I think, because we are using an actual element and not a pseudo element.

brentryanjohnson:
Thanks @el-mapache.

It looks like we still have a few issues.

## Sizing
- I wonder if we can introduce a class to accommodate varying sizes for these. It looks good in the normal body text, but the larger, intro text looks awkward with the smaller icon.

### Production
image

image

### Branch
image

image
image

## Internet Explorer 11

We're still aiming to support IE 11 on account of relatively significant use according to our analytics. Unfortunately, the icon is breaking the layout in IE 11.

We were having some icon spacing issues in IE recently; [I discovered that they require both `width` and `height` values](https://github.com/18F/doi-extractives-data/pull/2953) to avoid layout breaks.

image

el-mapache:
@brentryanjohnson oh cool, thanks for checking in IE. I will add the height too. Weirdly in chrome the sizing seems to match vs my branch and production, are you seeing the discrepency in IE? Or is it happening in chrome too? If so, it might be tough to reconcile it. Maybe we should just stick with the original 20px you set? I assume that is the size you want to see on IE

brentryanjohnson:
@el-mapache I think the challenge is related to how many contexts these book icons show up. The nice aspect of the font icons is they resize to the text size. 

## Contexts

The icons appear in four main contexts:

1. [Intro descriptions and body copy on Explore data pages](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/explore/), which are sized at `1.125 rem`.

2. [`Blockquote`-styled download page intro text](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/downloads/federal-revenue-by-location/), sized at `1.375 rem`.

3. [Caption `span`s for bar charts on Explore data pages](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/explore/#all-production), sized at `0.875rem`.

4. ["Dashboard" text on How it works pages](https://federalist-proxy.app.cloud.gov/preview/18f/doi-extractives-data/ab-replace-glossary-icon/how-it-works/), sized at `1rem`.

I'm not sure how to accommodate all of these contexts, except for a unique class for each. This is an issue in all browsers.

## IE issues

This is probably related to `height` properties, but sometimes the icon appears huge:

IE huge icon

And other times it appears properly sized, but has the spacing issues (which I suspect are due to the missing `height` attribute we've already discussed):

IE missing text

brentryanjohnson:
We need to revisit this issue with our new group; I'm closing to take up the issue later.

PR #3031, opened 2018-07-04
Use CircleCI 2.0

### LINKED ISSUE #3002: Update CircleCi to 2.0

adborden:
Fixes #3002

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/circleci-2/)

- Update CircleCI config to version 2


master12:
Also dont think we need docker

adborden:
Rebased to resolve the merge conflict.

I spoke to @master12 offline, CircleCI 2 is all-in with docker, so you don't _have_ to use it, but that is the recommended path. Instead of installing tools or specifying specific versions of ruby/node.js, it's all bundled into their supported docker containers. It also makes it easier to add a database just by adding a postgres container.

I added the webpack step, although I was surprised there was no check that would've caught this. html-proofer is one way to do that, but it will probably be its own PR.

adborden:
@master12 can you take another look?

Issue #3084, opened 2018-07-20
Build the button component in new pattern library framework


adborden:
We want to establish a new framework for the pattern library (formerly styleguide) that allows us to better share the technology and code between the site and the pattern library. This way a component only needs to be implemented once. We want to focus on having a usable pattern library that all new implemented components can be added to this new pattern library going forward.

This story establishes the library with a single component end-to-end. We'll start with the existing [button component](https://revenuedata.doi.gov/styleguide/components/detail/buttons.html).

- [x] Component uses the source code from our Gatsby component.
- [x] Component has documentation in markdown, or easily editable plain-text
  format.
- [x] Pattern library is live at `/patterns`
- [x] Pattern library is built as part of CI
- [x] All new components going forward can be implemented in this pattern library.

adborden:
@master12 had found a few frameworks that are better aligned with Gatsby:

- [Gatsby's styleguide example](https://github.com/gatsbyjs/gatsby/tree/master/examples/styleguide) based on react-styleguidist
- [React Styleguidist](https://react-styleguidist.js.org/)
- [React Live](https://react-live.kitten.sh/)

adborden:
@brentryanjohnson also pointed out [Brad Frost's Gatsby Style Guide guide](http://bradfrost.com/blog/post/style-guide-guide-gatsby-edition/) which is based on the Gatsby example style guide site.

adborden:
This was completed by https://github.com/ONRR/doi-extractives-data/pull/3093

PR #3093, opened 2018-07-25
Gatsby-based pattern library

adborden:
Part of #3084 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/pattern-library/patterns/components/)

- Creates a new pattern library based on Gatsby available at `/patterns`
- Pattern library build is included as part of the Federalist build.
- Pulls all components from `/gatsby-site/src/components` that have a README.md (only documented components).
- Pulls in documentation from README.md (markdown) as well as comment based docs in the component JS files.
- Uses `react-live` for live preview and editing.
- I had to tweak lato.css and eiti.css which are font and font-icon styling. I don't think they were being built correctly before and I suspect they could be removed or integrated better into `/_sass` although I'm unsure how to proceed.

Some things to consider:
- Fractal had a system for providing data for different variants. We don't have that here, so you have to copy/paste the variants yourself.
- Currently, the pattern-library build is separate with its own package.json, although the top-level package.json knows to build the pattern-library. I didn't see how the Gatsby site is being built with Federalist, but it might make sense to make the pattern-library part of the the Gatsby site build, sharing the same package.json and node_modules.
- For styles, I'm pulling in the compiled CSS from `/public`. This keeps the build simpler and once Jekyll has been removed, we can figure out how we want the styles to work in a Gatsby-centric world.
- Move the Table of Contents to sidebar so it's not a separate page.
- Add a new Docker container for cross-platform support. I'm not sure what the docker strategy is going forward, but since our CONTRIBUTING docs say to use Docker, I'm assuming Docker is still the way to go.


adborden:
This unblocks https://github.com/ONRR/doi-extractives-data/issues/3027

PR #3106, opened 2018-08-06
Gatsby explore data page

### LINKED ISSUE #2988: Build Disbursements Summary View
https://m13r6x.axshare.com/fed_1b.html

Can we do this with our currently technology for a quick win? 

Note: Design needs a few refinements before being finalized.

master12:
Fixes #2988 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/gatsby-explore-data-page/explore/#federal-disbursements)

- Added new disbursements feature
-
-

* please note that I will make updates to the build script but wanted to start the review process.


master12:
build script is now fixed and auto-magicked

master12:
@mcharg If there is no data for a fund for a year do we still show the fund?

mcharg:
@master12 To be consistent with other graphs, we should display it and indicate there is no data.  Can we swap out the graph with text? 

"There is no data about disbursements to [fund name] in [year]."

master12:
@mcharg We can do anything we want! We are invincible!!!!!

jennmalcolm:
I'm excited. This is looking pretty good. Besides from what Shannon noted about disbursements design, here are a couple of things I noticed that differ from the current version of the explore data page or need work:
1. In revenue, carbon dioxide is abbreviated CO2 rather than spelled out. Since the rest of the products are spelled out, we should spell that out too. 
2. The live version of the explore data page seems to take up more of my screen with smaller margins and the words appear bigger. This is not a big deal to me, but I thought I would note it.
3. The line at the top of the disbursements feature, under recipient and amount is not the same number of pixels across the whole thing. 
image
4. On the disbursements feature, the first bar under each fund is missing a label.
5. When I go to the About page, it flashes and then goes to the Explore data page.

On another note, @brentryanjohnson 
and I decided we could use "funded by Congress" under LWCF and HPF
instead of "appropriated" (which I know I indicated would be most accurate to you @mcharg). Sorry for the switch again. Funded by congress seems to be more understandable, and I have seen it in other government documents being used instead of appropriated, so I think it will work.

mcharg:
@jennmalcolm @brentryanjohnson Will we still have a glossary on "Funded by Congress"?

jennmalcolm:
I think we should still have the glossary for funded by congress, @mcharg to explain a little more what that means. Do you agree, @brentryanjohnson?

brentryanjohnson:
@mcharg Possibly, although we can also link to a relevant section in the contextual narrative. I'm not sure it works for the glossary, because it isn't a "technical term" or acronym in the way most of the other entries are. But it would be nice to describe it in context, so the user isn't offloaded to a separate page. We could look at using tooltips or accordions, too, but probably not if it's going to be a one-off.

As I work on the contextual content, I'll think more about this. Maybe we can co-work on it, too.

mcharg:
It seems weird for it to be a glossary, since it's not an individual term, but I agree that we should have some way to get an explanation.

mcharg:
@master12 Do you need to generate a new Preview link?  I'm not seeing the changes in the original link.

master12:
@mcharg still building...

brentryanjohnson:
It's odd because the CSS appears to be the same, but the margins for the icons are not resolving the same way in the Gatsby version.

## Production
proper icon margins

proper icon margins all commodities

## Gatsby
cramped icon margins

cramped icon margins all commodities


It looks like changing the `.icon` margin from `0 .1rem` to `0 .3rem` should fix this, but I haven't checked how that might affect other icon instances.

master12:
new build ready to go.

mcharg:
These issues appear to be fixed:
- Funds should be sorted by their 2017 total with the highest displayed first and not change order when you switch years.
- The "How this fund works" links should be underlined, since they're in a paragraph.

These are still issues:
- The sources should always be in the same order: Onshore, GOMESA, 8(g), Other Offshore
- All of the sub tables with the source numbers should be the same width, so the totals line up on the right.
- Column headers on the table should be regular font weight instead of light.
- The "Funded by Congress" line for LWCF and HPF is missing.

master12:
@mcharg "Column headers on the table should be regular font weight instead of light."
- I am using the article styling for tables here. So if I change this it will affect all tables. Is that ok?
-> [style guide basic table](https://revenuedata.doi.gov/styleguide/components/detail/basic-table.html)

mcharg:
@master12 I think it should be okay.  We'll just have to look at it after and see if it breaks any layouts.  The one area that I'm thinking might break is the smaller tables that display under the expand plus sign under the county graphics.

master12:
@mcharg actually the font weight is 400  and regular is 300, do you want this bold? 600?

master12:
New Build! Please Review!

mcharg:
@master12 Yes, let's bold the table headers.  Otherwise, the build looks great!

master12:
Ok here is the typography list form the css. The table headers were "regular" will switch to bold.

// Typography
$weight-light: 300; // (Lato Light) use
$weight-book: 400; // (Lato Regular) use
$weight-bold: 600; // (Lato Semibold) use

mcharg:
@master12 Can you also make the font larger?  The wireframe has 15pt.

master12:
@mcharg @brentryanjohnson @jennmalcolm  Ok hopefully this is the final build for this PR. Give it a check.

brentryanjohnson:
@master12 It looks like we might need a media query for this. The bar and numbers get cropped in a mobile viewport.

It might make sense to display the recipient and amount as stacked blocks below a certain viewport width.

## iPhone 8 viewport
phone viewport disbursements

## iPhone 8 viewport overflow
phone viewport disbursements overflow

brentryanjohnson:
@master12 Also, the year transitions are **awesome**. 👍

brentryanjohnson:
It looks like there is an extra space around the comma after the year:

In 2017 , ONRR disbursed a total of $7,111,809,625

` , ONRR disbursed a total of `

brentryanjohnson:
@master12 It looks like the 8(g) glossary entry is empty. I'll open a branch and add it to the terms (assuming this version of the site is reading off the same `terms.yml` file).

8(g) glossary entry empty

mcharg:
@master12 @brentryanjohnson The wireframes show what should happen to the design at different viewport sizes.

You can use the viewport picker icon in the top left to switch view:
image

From the notes on the table:
Columns should be equal widths.  Changes to one column view when the columns get below 300 px each.

The header row is not sticky on scroll because that would end up with 3 sticky headers on mobile, which would take up too much of the viewport and the header isn't very useful once you get to the one column view.

jennmalcolm:
I'm not looking in mobile, but the explore data page and the disbursements feature are looking very good to me.  I also really like the year transitions. I checked all of the disbursements data to the spreadsheet totals, and it looks good.

One issue I'm still seeing is that when I try to go to the about page, it flashes for a second and then just goes back to the explore data page.

master12:
@brentryanjohnson Gatsby uses different data files in the gatsby site. Specifically the terms.yml is a different structure. Look in here:
\doi-extractives-data\gatsby-site\src\data-graphql

@mcharg @brentryanjohnson Not sure what you want mobile too look like. Need mockup. Also think this shoud be a new ticket since the current table uses the current media query. This should be an enhancement and scheduled for future sprint, so we dont delay this any longer and it sounds like @jennmalcolm is ok with that.

@jennmalcolm I see that about page flash. I know why its doing this. It wont do it in prod but let me see if I can do a quick fix anyway.

mcharg:
I added a new issue #3145 for making tables mobile friendly.

jennmalcolm:
Yes, I think that's reasonable to create a new issue for the mobile design so we can settle on all of the details that @master12 needs to build.

Also, if the about page flash won't be a problem in prod, don't worry much about it now.

master12:
@jennmalcolm fixed the about page issue

PR #3157, opened 2018-08-20
Constrains data range to 10 years and fixes GPD year in dev

### LINKED ISSUE #3129: Constrain 'all production' to 10 years
I didn't catch it in [the review](https://github.com/ONRR/doi-extractives-data/pull/3126), but we constrain data to a 10-year range. The all production update is giving us 11 years.

It looks like we just need to update the [`years.yml`](https://github.com/ONRR/doi-extractives-data/blob/dev/_data/years.yml) file pair to `all_production: [2008, 2017]`.

brentryanjohnson:
Fixes #3129

[:pig: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/years/explore/WY)

- Constrains `years` data file to 2008–2017 for `all-production` and `revenue`
  - We're currently charting 11 years for those two datasets
- Reverts `now` year from 2018 to 2016 to fix GDP data in `dev`
  - We fixed this in production, but we merged directly into `master`, so this PR fixes in `dev`

### DIFF
--- a/_data/years.yml
+++ b/_data/years.yml
@@ -1,17 +1,17 @@
 # XXX this must be a string to use it as an index into other data!
-now: 2018
+now: 2016
 
 default: [2007, 2016]
 
 exports: [2006, 2015]
 
-revenue: [2007, 2017]
+revenue: [2008, 2017]
 
 self_employment: [2007, 2016]
 
 disbursements: 2017
 
-all_production: [2007, 2017]
+all_production: [2008, 2017]
 
 federal_production: [2008, 2017]
 


jennmalcolm:
@brentryanjohnson is it because of  Gatsby that I can't see those changes on the national explore data page?

brentryanjohnson:
I believe so, @jennmalcolm. And the Gatsby edition of the state page template will overwrite these pages, too, because @master12 mentioned that the Gatsby edition doesn't reference the same `years` data file as the current site. Even so, GPD is still broken in `dev` (because we merged directly to `master`), so I still think we should merge this to bring it up to date with production.

brentryanjohnson:
Which reminds me, @jennmalcolm, the new Gatsby `explore` PR with the disbursements summary overwrites the [nationwide production update](https://github.com/ONRR/doi-extractives-data/pull/3126) we previously merged, because they aren't reading off the same data file.

PR #3206, opened 2018-09-04
added mobile table css for disbursements

### LINKED ISSUE #3145: Make tables mobile friendly
Use the design done for the disbursements summary table.

https://qka4h2.axshare.com/#g=1&p=federal_disbursements_summary

master12:
Fixes #3145 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3145-mobile-tables/)

- Updated css to include breakpoint
-
-


mcharg:
The 768 px view is also off.  For the two column view, each column should be 50% of the viewport width (minus gutters).

mcharg:
@master12 

Took a look in Chrome:
768: looks good, except the right nav is missing.
600 & 480: Can the table fit the content instead of being the whole column width?  It looks funny with the labels so separate from the numbers.
320: Looks good.

I'll take a look in other browsers and on my phones after retro.

mcharg:
@master12 Looks good on my phones, aside from the menu button issue.

master12:
the menu issue is also in prod already. I think it has to do with hiding the side menu and showing the mobile view. I will see if there is any easy solution. Basically change 768 by 1 pixel and the menu shows up. @mcharg  So the question is at 768 do we want the side menu or mobile menu?

mcharg:
@master12 We should keep the same breakpoint for the menu that we have on the live site, which switches at 767.

mcharg:
@master12 Also looks good on Firefox.

mcharg:
@master12 None of the charts are showing up in IE 11.

mcharg:
@master12 Just noticed that above 768, the charts are squished to the right.  The column widths should be 50/50.

master12:
ok check it out

mcharg:
- Firefox issue is fixed.
- 768 right nav issue is fixed.
- Graphs are still not showing up in IE 11.  This is for all sections on Explore Data, not just Disbursements.
- The table width still looks funny at 600 and 480.  Can we size the table to the content instead of the column width?

mcharg:
The green is also missing from the map in IE 11.
image

mentastc:
Some of the issues I noticed in IE 11 
-the top map isn't clickable and it takes a long time to load, at first there isn't color in the map
-The glossary book isn't showing up (see image)
image

mcharg:
The map is also not clickable on my computer and the dividing lines between states aren't displaying properly.
image

The glossary icons do show up for me.

master12:
Comments for release:

1. Accordians have been fixed.
2. Performance on IE 11:  Nothing can be done about this currently. It takes 28 secs for react and 30s for "state-pages-min" to run. In Prod there is no react so it takes 30 secs. Not sure why IE 11 takes so long to let the code write to the DOM. The best thing we could do is provide a loading page icon as a quick fix. 
3. The map is not clickable in prod so this can be a seperate bug fix and should not impact this release.

All issues have been resolved other then the performance and the state map bug which exists in prod already.

mcharg:
I added #3237 to fix the map.

mcharg:
The accordions look good here, but they're still broken in the link for the other PR.

Issue #3047, opened 2018-07-11
Audit "Other revenue streams" section


brentryanjohnson:
I noticed the `federal revenue by company` link in this section (on state pages) defaults to 2015, which isn't the most recent year for that data. Update that path and review this content for accuracy.

This section is revealed by selecting the "Revenue details by phase" tab.

![other revenue streams section](https://user-images.githubusercontent.com/32855580/42580435-6106a72a-84df-11e8-9243-d93b34794a95.png)

PR #3224, opened 2018-09-08
Updates company revenue (UDR) data to 2017

### LINKED ISSUE #2804: Update federal revenue by company data to 2017
We have [2016 federal revenue by company data](https://revenuedata.doi.gov/how-it-works/federal-revenue-by-company/2016/). We should update this data to 2017.

#2805 covers moving this data out of [How it works](https://revenuedata.doi.gov/how-it-works/#revenues) to an explore data section.

brentryanjohnson:
Fixes #2804 and #3047

[:bird: How it works - revenue PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/company-revenue/how-it-works/#revenues)

[:dollar: Federal revenue by company, 2017 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/company-revenue/how-it-works/federal-revenue-by-company/2017/)

[:open_file_folder: Data download, revenue by company PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/company-revenue/downloads/federal-revenue-by-company/)

- Adds federal revenue by company for 2017
- Adds download file for federal revenue by company, 2013-2017
- Updates links, redirects, file paths, and references for federal revenue by company


brentryanjohnson:
@jennmalcolm The redirect is caused by the link going to the generic `federal-revenue-by-company`. There isn't a page there. since the federal revenue by company data is structured by year.

I can avoid the redirect by linking directly to `federal-revenue-by-company/2017/`; we'll just need to update those links every year when we update the data. The redirect allows us to update the link in one place (the redirect), instead of each link. But we still need to remember to update it there.

jennmalcolm:
Ok, if it's easier to update when we keep it like this, that's fine.

brentryanjohnson:
@jennmalcolm Perhaps we can make the links less disruptive when we migrate to the explore data page. That flash isn't ideal.

mentastc:
In doing my review I'm noticing some small revenue numbers and I believe everything under 100k is supposed to be rolled up, I asked Data retrieval to take a look, sorry I didn't catch this in my earlier data review.  I'll keep you posted for what they say.

brentryanjohnson:
I wondered about that, @mentastc, because there is a row titled: COMPANIES BELOW $100,000 THRESHOLD. At first, I thought the intention was to roll up under that threshold, but it certainly shows listings for companies with volumes that are less than that (for 2016, too). 

There is also this also a row for ONE PRODUCT AND ONE MINE (TOTAL OF WITHHELD), but there appear to be listings for companies that may meet this category, too. I do think these rows are supposed to be the rollups, and they are sorted according to amount, like the others.

_"Subheading" is in quotes because these titles are styled and appear the same as the companies themselves, which is confusing in its own right._

mentastc:
Here's DR's response... I told them rolling to oil and gas for the below 100k would be fine.

Hi Chris,

I just spoke to Andrew and we figured out what happened with the UDR as far as the "under $100,000 threshhold".  When Nathan and I fixed the file a couple of weeks ago, we didn't do the step to roll the revenue up.  Andrew is fixing that now.  As far as the Oil, Gas, and Oil & Gas issue, it would be easier and faster to roll the past years up to "Oil and Gas" than to break 2017 into "Oil" and "Gas".  Would it be ok to change the past years or would you prefer to keep the past years the way they are?

brentryanjohnson:
@mentastc I think rolling up to oil & gas is fine, since they're so closely related, and the emphasis is on the companies more than the commodities. As far as past years, oil & gas are split out, so if we want the numbers to be comparable year on year, we may want to consider updating those, too. Perhaps we can talk more about that at standup.

mentastc:
@jennmalcolm  Not yet.

brentryanjohnson:
@mentastc I've update the data file(s) with the roll-ups. Please review and make sure all looks correct.

mentastc:
Looks good to me thanks!

brentryanjohnson:
@master12 When this finishes building, will you give it a quick review on technical merits and merge if all looks good?

PR #3238, opened 2018-09-21
Fix glossary bug

master12:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/fix-glossary-bug/)

- fix bug in displaying glossary after clicking a term
-
-


master12:
OK check out the glossary!

mcharg:
The glossary is working (except the icons are broken in IE), but the charts are missing in all the sections except Disbursements and the Nationwide menu is broken on mobile.

master12:
ugh, had the wrong setting in sourcetree so it was committing all the files properly. Added them now. Please hold...

master12:
Ok site is up again. 

IE 11 icons missing is  asecurity setting. I think our federalist site is not trusted and hence downloads of fonts might be disabled. It alos might explain other weird behavior in IE. I tested on my persnal machine and the icons work just fine

PR #3273, opened 2018-10-05
Clarifies status of Land and Water Conservation Fund

brentryanjohnson:
[:evergreen_tree: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/lwcf/how-it-works/land-and-water-conservation-fund/)

- Clarifies status of LWCF following the end of the 2018 fiscal year
- Adds pattern for update date flag

land and water conservation fund status screenshot

### DIFF
--- a/_how-it-works/lwcf.md
+++ b/_how-it-works/lwcf.md
@@ -53,7 +53,9 @@ The LWCF funds landscape conservation, outdoor recreation, and the needs of loca
 - wildlife habitat
 
 ## Authorization and funding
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
I created a second branch with an alternative design pattern featuring a more conspicuous status call-out (@mcharg and @jennmalcolm)

[:mega: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/lwcf-flag/how-it-works/land-and-water-conservation-fund/)

screenshot showing update flag with background color

mcharg:
I like the second option better. It makes it easier to see what's been updated when skimming the page.

mcharg:
Another option would be to use the same style as the archive pattern, so we're flagging status in the same way in both cases.

jennmalcolm:
The second option does draw your attention to the update date, which I like. Let's go with that.

brentryanjohnson:
Okay. I'm going to open a separate PR with the appropriate branch. I'll also create an include for the pattern, so we don't have to repeat the markup in every instance if and when we reuse it.

PR #3304, opened 2018-10-18
Adds smooth scrolling to gatsby pages

### LINKED ISSUE #3161: Add smooth scrolling to Gatsby explore data
We decided to test smooth scrolling to assist users with wayfinding on the explore data page, but the Gatsby version of `explore` doesn't appear to reference the same style sheet, which means it will have aberrant scroll behavior compared to the other pages of the site.

We should add the smooth scroll behavior to the Gatsby version of the explore data page.

brentryanjohnson:
Fixes #3161 

[:turkey: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/gatz-smooth-scroll/explore/)

- Adds smooth scrolling property to GatsbyJS portion of the site

### DIFF
--- a/gatsby-site/src/styles/_reset.scss
+++ b/gatsby-site/src/styles/_reset.scss
@@ -22,6 +22,7 @@ time, mark, audio, video {
   font-size: 100%;
   font: inherit;
   vertical-align: baseline;
+  scroll-behavior: smooth;
 }
 
 /* HTML5 display-role reset for older browsers */


brentryanjohnson:
I'd like to know what you think about this behavior, @mcharg. It's a little jolting on a page this long, but it also orients the user to their relative location on the page. It's a trade-off we'll want to continue to evaluate.

mcharg:
@brentryanjohnson Can we try slowing down the scrolling or making the speed relative to how much content you're scrolling through?

mcharg:
I'm fine with going with this and seeing if we get complaints from users.

brentryanjohnson:
I'm sensitive to the user responses, and it would be good to get as much feedback as possible. As I mentioned before, there is the possibility of triggering vestibular issues for some people. We're weighing the drawbacks against what we assume will be a better orienteering awareness on that long page of content. Let's assess repeatedly.

Issue #3321, opened 2018-10-22
Plan for how it works content for disbursements


jennmalcolm:
We  need to look at creating a new section on the how it works page for disbursements content.

My initial feeling is that under "Resources to Revenue," we could add a disbursements section. We need to have some type of overview content, but then we could also have sections for the LWCF and HPF content that already exists. That would make it easier to find that information too.

This issue will be done when we have a plan for what content to add and we have created follow-on issues for the creation of that content.

brentryanjohnson:
We have a scope for disbursements context in how-it-works. Here is the outline for what we're planning to provide:

- Laws and regulations that govern each disbursement
- Frequency of disbursement to each fund
- Which are subject to Congressional appropriations
- Provide layout that brings together existing disbursements contextual narrative (GOMESA, Land and Water Conservation Fund, Historic Preservation Fund)

lpgoldstein:
I was able to connect with another person re the HPF content and she will hopefully get back to us next week as she is on leave/work travel this week.

PR #3346, opened 2018-11-02
Adds how-it-works content and layout for disbursements

### LINKED ISSUE #3227: Expand detail in disbursements documentation
We lack sufficient information about how disbursements are apportioned. For example, we list how much ONRR disbursed to states (and each state), but we don't have documentation about the formula that determines how much a given state receives.

We also lack sufficient documentation describing why and how portions of geothermal revenues are disbursed directly to counties.

We should consider reviewing our documentation and determine where it may lack sufficient information to understand the data.

brentryanjohnson:
Fixes #3227, #3321

[:deer: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/understand-disbursements/how-it-works/#disbursements)

- Adds section to how-it-works layout dedicated to disbursements
- Creates new supporting content for disbursements

### DIFF
--- a/_how-it-works/default.md
+++ b/_how-it-works/default.md
@@ -168,12 +168,18 @@ selector: list
       </div>
       <h3 id="revenues" class="h3-bar">Understanding federal revenues</h3>
       <p class="landing-intro_link">When companies extract natural resources on federal lands and offshore areas, they pay revenue to the federal government. The government distributes these funds for public use in a variety of ways.</p>
-      <p><a href="{{site.baseurl}}/how-it-works/revenues/">Learn how revenues work</a>.</p>
       <div class="container landing-section">
+
+        <div>
+          <h4 class="h3 landing-heading"><a href="{{site.baseurl}}/how-it-works/revenues/">How revenue works</a></h4>
+          <p>Companies pay the federal government to extract natural resources on federal lands and waters. This revenue is collected by the Office of Natural Resources Revenue.</p>
+          <p><a href="{{site.baseurl}}/how-it-works/revenues/">Learn about how revenue works</a>.</p>
+        </div>
+      
         <div>
           <h4 class="h3 landing-heading"><a href="{{site.baseurl}}/how-it-works/federal-revenue-by-company/">Federal revenue by company</a></h4>
           <p>See federal non-tax revenues from natural resource extraction on federal land in 2017 by commodity, revenue type, and company.</p>
-          <p><a href="{{site.baseurl}}/how-it-works/federal-revenue-by-company/">See revenue by company</a></p>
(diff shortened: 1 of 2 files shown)


mcharg:
Here's my feedback:

- You might want to use the word "source" somewhere in the "disbursements and location" section, so users can tie it to that when we use the word "source" elsewhere.
- Will listing out LWCF, HPF, and GOMESA under disbursements make users think those are the only funds we disburse to?

jennmalcolm:
I think this is on the right track. I think it would make more sense to break out onshore, offshore, and Native American all separately. I.e., I would get rid of the "Onshore and offshore" section. Some of that could be used in an intro for the Disbursements and location section, but then I think onshore should have its own section.

brentryanjohnson:
Thank you both, @jennmalcolm and @mcharg. 

@mcharg I thought the same thing about listing those funds' pages, but I'm not sure what else to do with them. We're not going to produce the same level of content for every fund, so we may have to: (1) rely on users understanding the supporting content doesn't include all funds, (2) roll the LWCF, HPF, and GOMESA content under a single disbursements narrative, (3) write basic content for each fund and include it under that section, (4) something I haven't thought of, yet! It's a great point, since it's now conspicuous that we don't have supporting content for the Reclamation Fund (for example).

@jennmalcolm I added a dedicated onshore section and elaborated on the states' portion. Please review again, especially the reference to the 1% of "admin" funds. I couldn't remember exactly how we worded this before.

jennmalcolm:
@brentryanjohnson it looks good to me.

mcharg:
@brentryanjohnson What if we made the"learn how disbursements work" a paragraph instead of just a link and explain there that it explains everything.

Something like this:
**How disbursements work**
Learn about who receives disbursements, the laws governing them, and the impact location has on who gets the money.
_Learn how disbursements work_

We could do the same thing for revenue, if we want to be consistent.

brentryanjohnson:
@mcharg Breaking out those links into the section is a good idea, since those links get lost a bit. I'll try that and see how it works.

brentryanjohnson:
@mcharg and @jennmalcolm I pushed another commit; please have another look!

brentryanjohnson:
@mcharg Testing without the paragraphs, and it looks _much_ cleaner and easier to interact with. I'm going to remove them for now, as you suggested, and we'll rewrite them if it seems useful to connect the dots between the links below each subheading. 👍

mentastc:
@brentryanjohnson  we could add "other to the listed funds. these are payments to federal agencies

image

jennmalcolm:
@brentryanjohnson what do you think about adding the other funds as @mentastc suggested?

PR #3348, opened 2018-11-02
Adds hover state to icon for testing

brentryanjohnson:
[:horse: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/icon-hover/explore)

@mcharg mentioned an icon hover state might be a user-friendly design behavior to test. This simple hover uses a `fill` property on hover to change the icon's color to gray. We could also `fill-opacity` to desaturate the blue. 

We can test this and find out if it is a useful behavior for users.

icon-hover

### DIFF
--- a/gatsby-site/src/components/layouts/DataAndDocs.js
+++ b/gatsby-site/src/components/layouts/DataAndDocs.js
@@ -2,7 +2,7 @@ import React from 'react';
 import DataAndDocIcon from '-!svg-react-loader!../../img/svg/icon-data-and-docs-circle.svg';
 
 const DataAndDocs = (props) => ( 
-    <span>
+    <span className="icon-link">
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
After reviewing this as a marked-up prototype, I'm not convinced it adds much. The underline on hover (even when the hover is over the icon) indicates the link is active and extends to the icon, at least for me. Let me know if you think otherwise. 🤓

mcharg:
@brentryanjohnson I think you're right.  It doesn't add much value, since the link does underline when you hover over the icon.

jennmalcolm:
I agree that it doesn't do much.

brentryanjohnson:
Closing for now. We can revisit this concept, if it comes up in user research.

PR #3351, opened 2018-11-03
Changes 'Tribal' to 'Native American' in map legend

brentryanjohnson:
[:deer: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/content-guidance-legend/)

- Changes map legend in Jekyll and GatsbyJS portions of site to conform to our style guidance and federal law

### DIFF
--- a/_includes/maps/federal_land_ownership_legend.html
+++ b/_includes/maps/federal_land_ownership_legend.html
@@ -1,7 +1,7 @@
 <div class="legend-ownership-svg">
   <ul class="cells">
-    <li class="cell"><span class="swatch"></span>Federal{% if include.land %} land{% endif %}</li>
+    <li class="cell"><span class="swatch"></span>Federal{% if include.land %} lands and waters{% endif %}</li>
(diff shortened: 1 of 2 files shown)


mcharg:
I know this wasn't part of the change, but should we change it to "Federal lands and waters"? and "Other land (state, local, or privately owned)".  I don't know why this change made me notice more that "Other" is the only thing that doesn't say land and calling out waters would be consistent with the rest of the site and help explain the offshore zones.

jennmalcolm:
@brentryanjohnson, I think @mcharg's suggested edits could add clarity and consistency. For the other category, I do think it would make more sense to have "land" outside of the parenthesis, but I'm thinking adding "owned" isn't necessary.

mcharg:
I just noticed that federal has plural "lands and waters", but the others are singular "land".  Should we be consistent?

brentryanjohnson:
@mcharg Yes, I was thinking about that, too. We have the same issue with "revenues" versus "revenue." I need to add some guidance to the content guide for both.

PR #3374, opened 2018-11-08
Fixes styling of explore data button icon on homepage

brentryanjohnson:
[:snake: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/explore-data-button-icon/)

- Removes `root`-level `svg` selector overriding local styles
- Adds styling for chart icon

### DIFF
--- a/gatsby-site/src/components/layouts/buttons/ExploreDataButton/ExploreDataButton.module.css
+++ b/gatsby-site/src/components/layouts/buttons/ExploreDataButton/ExploreDataButton.module.css
@@ -9,13 +9,6 @@
 	padding: 10px 20px;
 }
 
-.root svg{
-	background-color: white;
-	fill: blue;
-	margin-right: 8px;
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
@master12 They are classes we've used on other icons on the site specifically so they _are_ reusable. The first applies to most are all icons, and the latter to icons that precede text. They preexist this pull request.

Do you prefer inline styling? That seems like a lot repetition in the code, since most of the icons share styling, but I'll defer to you.

brentryanjohnson:
Also, @master12, we should codify styles and practices into a linter and/or documentation.

master12:
@RYANCJORDAN Lets leave the classes for now then. Seems strange to me but we can have a discussion.

brentryanjohnson:
@RYANCJORDAN Ignore the comment ☝️. He means me. :)

PR #3413, opened 2018-11-27
added hover state to chart legend when the bar chart has a hover stat…

### LINKED ISSUE #3392: Fix remaining homepage bugs

- [x] Fixed

- [x] Tested

- [x] Fixed

- [x] Tested

- [x] Fixed

- [x] Tested

- [x] Fixed

- [x] Tested

image
image

- [x] Fixed

- [x] Tested

master12:
Fixes #3392 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/fix-chart-legend-hover//)

- Added hover state to the chart legend.  Address this issue:
"On hover, the swatches in the table should change to the hover color."
-
-

### DIFF
--- a/gatsby-site/src/components/charts/ChartLegendStandard/ChartLegendStandard.js
+++ b/gatsby-site/src/components/charts/ChartLegendStandard/ChartLegendStandard.js
@@ -6,10 +6,9 @@ import utils from '../../../js/utils';
 
 import styles from "./ChartLegendStandard.module.css"
 
-
 /**
  * Chart Legend Standard is used to display rows of data keys and values. Each data key can be 
- * associated with a format function, a color(using css) and a display name.
+ * associated with a format function, a css class and a display name.
  */
 class ChartLegendStandard extends React.Component {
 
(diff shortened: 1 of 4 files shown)


brentryanjohnson:
While I understand the intent behind this behavior (for the legend to match the hover state of the stacked bar), it comes across as "buggy" and distracting to me. Hovering over several bars in succession strobes the legend in a way that appears to be unintentional and takes my eye off the chart and associated data.

 Can we assume the order and relative opacity of the legend items will allow users to cognitively match to the bars, despite the slight saturation difference on hover?

mcharg:
We had it this way in user testing and nobody said it seemed buggy or distracting.

mcharg:
It was also this way when we were doing design reviews and nobody had a problem with it then.

jennmalcolm:
I also find it somewhat distracting, and I'm worried it's not really enhancing the meaning (for people like me who are not great at differentiating subtle color differences). Is there a reason not to use a behavior like the bar charts on the explore data page (upon hovering over a year that is not selected, make the selected one for that commodity go grey and the hover one go blue)? Then we could just use the three different blues and the color flashing in the legend could go away. I know the users liked this design and didn't have any problems, but I think it's helpful to continue making little tweaks as our team notices potential improvements. Since our user interviews tend to cover a lot of ground, it could be hard for something this this to surface during those discussions (unless we asked them specifically about this behavior--which I don't remember but could have happened).  

@mcharg, please let me know what you think about instead using the hover behavior from the explore data page instead.

mcharg:
If we wanted a quick gauge of whether it's distracting, we could show it to a few people in the office who don't know what it's supposed to do and see if they say anything about it.

jennmalcolm:
Ok @mcharg that sounds reasonable. I will try to grab a few people around the office today.

mcharg:
@jennmalcolm When you do that, just ask them what they think of the charts and what happens when you hover, so you don't bias their thinking and call attention to the things we think might be problematic.

jennmalcolm:
I asked one person and will get another later today. I also her to look at the hover on Explore Data and compare the behavior.  Can you also ask at your interviews this afternoon too @mcharg?

mcharg:
@jennmalcolm said we can go ahead an merge this based on the feedback we got from people today.

We showed it to:
* A previous study participant who saw the charts without interactivity and wanted them to function like the Explore Data ones and he said "I like it a lot, it's very interactive and self explanatory".  He noticed the numbers changing, but not the colors.
* A user with vertigo who said the hover functionality was "crazy" in a good way because you didn't have to leave the page to get details.  She didn't notice the color changing on hover.

PR #3436, opened 2018-12-04
Updating monthly production volumes

### LINKED ISSUE #3392: Fix remaining homepage bugs

jennmalcolm:
Partially Fixes #3392 
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/Update-production/)

-updates monthly production data through July 2018, most recent available


jennmalcolm:
Thanks @brentryanjohnson for catching those difference. I had left some July 2018 lines I put in there for my test PR in the data file. So, I think that was the issue. The July 2018 numbers were each off by 1. They should be fixed with the new commit.

PR #3442, opened 2018-12-05
Fix prod charts sync

### LINKED ISSUE #3392: Fix remaining homepage bugs

master12:
Fixes #3392 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/fix-prod-charts-sync/)

- Created single reducer
- Added set data selected
- refactored code


mcharg:
@master12 

Works great on:
- Chrome
- Firefox
- IE
- Edge

Doesn't work so great on:
- iPhone Safari
- iPad Safari

When I click on a bar, the last selected year or month stays selected.  The new bar gets selected and the table updates, but the other charts aren't affected.
iPhone Yearly
img_9735
iPhone Monthly
img_9736
iPad
img_1039

mcharg:
@jennmalcolm Are you okay with leaving the behavior as is on Safari?  See my comment above about the issue.

master12:
The click event is not firing. Will try to change to onMouseDown since hover is working

jennmalcolm:
@mcharg I approved before I saw your comment and had not looked on my phone. I too would like that issue to be fixed before merging.

master12:
Ok mousedown works. I noticed another issue so will fix that and update.

master12:
@mcharg I noticed if you move your mouse really fast the hover event doesnt fire enough to keep up. Which results in the legend still thinking something is hovered. Not sure how if can fix that due to it being a  browser speed issue. DO you still want to keep the hover feature updating the legend?

What I did add was an mouseout on the entire chart and that helps to catch it only if your mouse leaves the chart area. So its still an issue but thats the best I can do.

mcharg:
Yes, I think we should keep it to be consistent with the Explore Data charts.  Most users probably won't mouse really fast and if they end up in a weird state, it's easy enough for them to fix it by hovering over something else or refreshing the page.

master12:
@mcharg I think I got it fixed for the hover. Added a background rect to catch hover events and clear the hover.

jennmalcolm:
Looked at it on my iphone, and it seems like we still have the issue of the 2017 bar saying highlighted. Also, the bars for all commodities are not synced yet on my phone.

mcharg:
I'm still seeing the issue on my phone, as well.

master12:
@mcharg @jennmalcolm Could you try one more time? I cant recreate the issue any more on my macbook even with the emulator set to iphone se and ios

mcharg:
@master12 I'm still seeing it on both my phone and iPad.  Could it be something with an OnTap event?

mcharg:
And I confirmed that it's working fine on my Android phone.

master12:
Ok I dusted off my ipad and it does work but very poorly. 
1. Really hard to get your finger to hit the bar. 
2. For some reason I hit it once its highlighted and the hit it again to actually select it. 

But it did work but it was a bad experience. Please try again and see if you can get it to work even if it's terrible. 

My thought on a solution is to have a overlay that is wider then the current max. Basically it will be the calculated max width vs the 15 px that is currently set. I will also try to not make it highlight on first touch. weird...

My thought though is not to hold up this fix for that.

jennmalcolm:
It's working now on my phone. It is still a little hard to select a year, but not too bad. I'll merge this in .

mcharg:
It's working on my phone, too!

PR #3473, opened 2018-12-11
3409 update county prod label

### LINKED ISSUE #3409: Update county production label
The label above the maps should be "COUNTY PRODUCTION".
image

master12:
Fixes #3409

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3409-update-county-prod-label/explore/TX/#federal-production)

- Added default value of county to table heading
-
-


brentryanjohnson:
It looks like this fixed the county reference referred to in the issue, but a county label is also missing from the county table.

county table heading below choropleth map

mcharg:
The above issue appears to just be for states with counties.  Louisiana and Alaska have the label on the table.

brentryanjohnson:
We talked about it @master12, and it looks like you've added the commit, but we should have consistent capitalization for units (this is a problem in production as well).

Units should be lower case in all instances. They are lower case everywhere except in the expanded county production table.

inconsistent capitalization of barrels and mcf for units

PR #3528, opened 2019-01-29
Converts table image to semantic table for accessibility

### LINKED ISSUE #3067: Replace static image of table with HTML table
The [How it works-->Revenues table](https://revenuedata.doi.gov/how-it-works/revenues/#federal-lands-and-waters) has valuable information, but it's a `png` image with insufficient `alt` text. Consequently, it's content is neither accessible nor machine-readable. Let's replace it with an HTML table.

static table image

brentryanjohnson:
Fixes #3067

[:cow: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/image-to-table/how-it-works/revenues/#revenue-streams-and-rates)

- Converts image of table to semantic, structured tabular content
- Omits tax section from table (since it's outdated and doesn't relate to ONRR revenue)

### DIFF
--- a/_how-it-works/natural-resources/revenues.md
+++ b/_how-it-works/natural-resources/revenues.md
@@ -186,7 +186,7 @@ Lease holders also pay different fees to the Bureau of Land Management, Bureau o
       <td><strong>Bonus</strong> (competitive lease)<br>
           <strong>$0.25</strong> per acre for acquisition fee (uncompetitive lease)</td>
       <td><strong>$3</strong> annual rent per acre</td>
-      <td><strong>2%</strong> of anticipated value of wind energy produced in operating fee, unless otherwise specified or waived
+      <td><strong>2%</strong> (unless otherwise specified or waived) of anticipated value of wind energy produced in operating fee
       </td>
     </tr>
   </tbody>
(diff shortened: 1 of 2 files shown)


mcharg:
@brentryanjohnson Can you make the underlines under the Phase header green to match the style in the old style guide? The last example has a spanning header.
https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/aml-spark/styleguide/components/detail/basic-table.html

The icons also feel weird in black.  I think they're always green, blue or gray elsewhere.  Would it make sense to match the style for both the header and icon used in the revenue table?
image

brentryanjohnson:
@mcharg The green from the style guide is tied to the `<th>` element, which is why it shows up under the table header in the preview. I can play with the color of the phase border, too, but I'll have to write some custom markup to do so.

The icon is an include, so it's the same markup used most everywhere else for the same icon. When I was working on icons last year, I wrote some CSS whereby the icon inherits the color from its surrounding text (`fill: currentColor`). To make it green, I'll need to duplicate the icon includes and write some custom markup.

Because of the "phase" part of this table, we don't have a pattern that we can apply to the whole table element. I'm using as many existing styles as I can and trying to limit custom markup for this one table.

mcharg:
@brentryanjohnson Probably not worth a lot of custom work to make the icon color change, but I think changing the color on the Phase line is worth a little bit of custom work.

jennmalcolm:
Could we make the column widths for the different phases more consistent between the commodities?

mcharg:
@brentryanjohnson Looks so much better with the green line and that also makes the black icons look better.  I think the black lines were throwing all the visuals off.

I agree with @jennmalcolm about the column widths.

mcharg:
@brentryanjohnson Gotcha.  I think it's okay to leave as is until we get to Gatsby.

jennmalcolm:
Ok, makes sense to me @brentryanjohnson.

PR #3611, opened 2019-02-11
Corrects reference to PRA

### LINKED ISSUE #3610: Correct reference to PRA in blog post
18F pointed out that the PRA doesn't actually prohibit recording user research.

mcharg:
Fixes #3610

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/correct-blog-post/blog/usability-testing-training/)

- Changes reference to PRA to point out that it's logistics that prevent us from recording user research
- Adds a Corrections section to the end of the post pointing out the correction


mcharg:
@brentryanjohnson We should decide on a pattern for posting corrections to blog posts.

jennmalcolm:
Should we do this directly to master so we can get it out there today? If so, you'll need to set up a new PR, and we should increment the version too.

mcharg:
Yeah, it would be good to get it out there today.  I'll see if I can figure out how to branch from master.  I don't know anything about incrementing versions.

mcharg:
I don't appear to have access to branch from Master.

jennmalcolm:
I could do the version increment. I guess we can just do a quick release with dev to master today. I think there is only one other change in there, about the blog favicon. Go ahead and keep it here, and I'll do the increment version for the release PR.

brentryanjohnson:
@mcharg The reason we patch directly to `master` is because we haven't properly vetted everything else in `dev` at that point. But since we just released last week, we don't have a lot in `dev` that hasn't been released. So I agree with @jennmalcolm; the normal workflow through `dev` should be fine for this patch.

brentryanjohnson:
The only change I recommend is to style the correction differently than the rest of the content. A style that departs from the rest of the content will make it visually clear on scanning the post that the correction was added later and is categorically different than the preceding content.

Perhaps something like this?

_**Correction**: This blog post originally stated that the Paperwork Reduction Act prohibited us from recording user research sessions, but we have since learned that the Act doesn’t actually prohibit us from doing that. (Yay! Something we can look into doing in the future.)_

mcharg:
I like it!

PR #3712, opened 2019-03-01
3556 howitworks revenues

### LINKED ISSUE #3556: Build How Revenue Works page in Gatsby
https://revenuedata.doi.gov/how-it-works/revenues/

master12:
Fixes #3556

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3556-howitworks-revenues/how-it-works/revenues/)

- Added revenues page to how it works
-
-


master12:
@mcharg Regarding the nav. My code automatically creates the nav. I can change the names of the labels but hesitate to match their current nav completely since it doesn't match the pattern. I can exclude and change names but would like to stay consistent with the hierarchy that matches the nav on all other pages. Does that make sense? Basically H2 is a main label and H3 is a child.

mcharg:
@brentryanjohnson Needs to weigh in on the nav issue.

master12:
@mcharg @brentryanjohnson I probably could include a flag to move the h3 to the main and vice versa, so I leave it up to you what you want to do

brentryanjohnson:
In the previous version of the site, the nav was manually created and maintained. Ideally, we'd have an auto-generating in-page nav, but also be able to manually override the text. This page provides a great demonstration as to why: the subheadings are verbose, which doesn't translate well to a nav menu.

For now, I think it's fine to use the subheading text for the nav listings, but if there's a way to override it with custom text, that would be better.

mcharg:
Created #3726 for overriding the nav text.  Other updates look good.

master12:
@brentryanjohnson @mcharg There is a way to override the text already. There is also a way to exclude headings. However I did not include a way to override the hierarchy but I can. I think since the H2 was excluded in favor of the h3 to be the main link then I will go ahead and include this override as well. So in summary we can override the creation of the menu in multiple ways:

1. Change the text (add attribute displayText attribute)
2. Exclude by adding shouldDisplayTitle = false
3. Change place in a hierarchy by using displayAs attribute (= h3 or h2)

I'll make the change for the 3rd and check it in to this branch, you can delete the extra ticket for changing the label text that is already available.

mcharg:
Right nav looks good.

PR #3750, opened 2019-03-06
3747 downloads reconciliation

### LINKED ISSUE #3747: Build reconciliation page for downloads
https://revenuedata.doi.gov/downloads/reconciliation/

master12:
Fixes #3747

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3747-downloads-reconciliation/downloads/reconciliation/)

- added reconciliation page
-
-


brentryanjohnson:
@mentastc The MSG link on this page is broken, which I suppose makes sense, since the MSG no longer exists. Are the meeting minutes archived somewhere? I suppose we should change the link to go to an archive, if one exists.

jennmalcolm:
I noticed that too @brentryanjohnson but I was thinking that since we note that the content is archived and no longer updated, maybe we shouldn't do anything with it. I asked @mentastc about the MSG site, and he said there's no way for the public to get to it anymore. He can still access the archive. So, maybe we just remove that link all together?

mentastc:
@brentryanjohnson The page isn't publicly available anymore, we should just remove the MSG link.

brentryanjohnson:
Since the page is archived, I think we can just leave it alone. I suppose there's value to keeping the link there, conveying that we at one time had a page dedicated to that group. It's probably more transparent (can something be _more_ transparent?) to keep it there, even if it's broken.

usEITI:
I would encourage you to leave the link in this case because DOI infers that information is not available anymore as a possible explanation.  Note:  There are multiple links in the 2015 Executive Summary to the MSG website and we cannot remove those.

jennmalcolm:
Ok, it's a deal. We'll leave the link.

PR #3789, opened 2019-03-13
Dev-->Master

brentryanjohnson:
[:monkey: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/dev/)

- Updates monthly production and revenue data
- Publishes a new blog post about a unique path to digital product management
- Updates the status of the Land and Water Conservation Fund after new legislation was signed into law
- Updates Coal Excise Tax rates and replaced a chart image with tabular data for accessibility


brentryanjohnson:
@mcharg It's trivial to bold the totals, but I didn't want to write custom markup for just this table for the alignment. I need to track down a table style with that pattern. We should also get the pattern library updated, so we can review and apply a pattern with the desired style.

mcharg:
The revenue table design has the desired pattern:
https://etoy5p.axshare.com/#g=1&p=revenue

The reconciliation page has a table with the right alignment:
https://revenuedata.doi.gov/how-it-works/reconciliation/2016/

brentryanjohnson:
@mcharg We can't use CSS styles from Axure. And the Reconciliation table has an `article_table-last_right` style, which right aligns only the last column. That works for Reconciliation, because dollar figures are only in the last column. In this case, we have numbers in three columns.

mcharg:
The old pattern library shows this table with right alignment in multiple columns, but I was unable to find it on a page:

brentryanjohnson:
Great find, @mcharg. I think there is a class in there that might work (`article_table-numbers`).

PR #3844, opened 2019-04-01
Question icon to replace glossary

### LINKED ISSUE #3693: Change glossary icon from book to question mark
See design here: https://etoy5p.axshare.com/#g=1&p=revenue

Tasks:
- [ ] Change icon
- [ ] Finalize copy changes for header in the drawer and site header
- [ ] Implement copy changes

This change is needed before we implement the monthly disbursements filterable table (#3187) and new revenue graphic (#3692).

brentryanjohnson:
Fixes #3693

[:dragon: PREVIEW](https://federalist-proxy-staging.app.cloud.gov/preview/onrr/doi-extractives-data/question-icon/)

- Attempts to replace book icon with question mark icon


brentryanjohnson:
There are a couple things I could use your help looking out for, @mcharg.

The book icon was a font icon. Consequently, it would size itself according to the surrounding font size. I've replaced it with an `svg` in this branch, so...

1. The icon size needs to be declared for each context in which it appears. For instance, I'm using a different width and height for paragraph text than the caption text for instances such as bar chart captions. We'll need to watch for new or existing instances with unique font size requirements.

2. There are some instances in which all `svg`s are given styling properties that are meant to style large `svg`s, (not icons). For the one instance of this I found (`.ribbon`), I added a rule that omits `svg`s with the `.icon` class from being targeted:
```
.ribbon {
  svg {
  svg:not(.icon) {
```
We'll need to watch out for other, similar instances.

3. There is unique styling for the question icon that accompanies the glossary toggle (larger version). If we have uses for a larger version, we should try to reuse this one.
```
.glossary-click .icon-question {
  width: 30px;
  height: 30px;
}
```

mcharg:
Can we keep the icon from wrapping to the next line, like in this example on the Native American Production page?
image

mcharg:
Does the icon in the archive banner need to be smaller?
image

brentryanjohnson:
I noticed that one, too, @mcharg. That're related to the sizing challenges I pointed out above. Since this is a temporary notice for in-place archiving, I left it alone, because I don't want to write a CSS rule for every instance (it would get really messy and difficult to maintain).

mcharg:
@brentryanjohnson I'm not seeing any other places where it looks missized.

Case Studies still has the old icon.  Should we update it there or just leave it as is since it's an archive?

brentryanjohnson:
I noticed that, too, @mcharg. It's because case studies wasn't ported over to Gatsby, and I only made the change in Gatsby. Changing the icon in case studies is a bigger lift, because the icon is placed via a Ruby plugin, and not in the CSS. Since we're archiving case studies, I've left it alone, for now at least.

mcharg:
I think that's fine.

PR #3853, opened 2019-04-04
Removes blog blockquote border

brentryanjohnson:
[:dragon: PREVIEW](https://federalist-proxy-staging.app.cloud.gov/preview/onrr/doi-extractives-data/blog-quote-border/blog)

- Removes blog blockquote border

## How it _should_ look
block quote border too close to text, text is just example text and not meaningful

## How it often looks
block quote without border

## How it will look now

border removed from block quote, so just intro and block quote


brentryanjohnson:
I'm going to resubmit when merged into `dev` to avoid conflicts and package problems.

PR #3871, opened 2019-04-09
Edits sequestration line in monthly disbursements docs

brentryanjohnson:
[:ghost: PREVIEW](https://federalist-proxy-staging.app.cloud.gov/preview/onrr/doi-extractives-data/docs-seq/)

- Edits line for "sequestration" in accordance with SME guidance


jennmalcolm:
On this preview, the monthly disbursements tab doesn't work. This may have been branch off before Jeff made the change to fix that link reference issue?

brentryanjohnson:
Yes, it's missing the preview path `/preview/onrr/doi-extractives-data/docs-seq/`. It should resolve in production, but we should probably confirm with @master12.

Issue #3937, opened 2019-04-15
Fix drop-down alignment in Firefox


mcharg:
Organize by is below the others.
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/2fa78960-8e57-4609-b21f-3aa7ba422104)

brentryanjohnson:
@mcharg and @master12 This is due to the collision of and element with `overflow: hidden` and a subsequent `inline-block` element. Changing `hidden` to `visible` fixes this problem in Firefox, but it results in the year list overflowing the select list (when many years are selected). We'll have to find an alternative solution.

mcharg:
@brentryanjohnson Is this done?

brentryanjohnson:
Yes.

PR #3956, opened 2019-04-16
Revenue table firefox fixes

### LINKED ISSUE #3935: Add horizontal row markers to first column on filterable tables in Firefox

brentryanjohnson:
Fixes #3935 and #3937 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/revenue-table-firefox/explore/revenue)

- Fixes border styling for first column in Firefox
- Fixes alignment of select lists in Firefox

### DIFF
--- a/src/components/tables/FilterTable/FilterTable.module.scss
+++ b/src/components/tables/FilterTable/FilterTable.module.scss
@@ -47,14 +47,10 @@
 	tr {
 		th:first-of-type {
 			position: sticky;
+			background:white; 
+			background-clip: padding-box;
 			left: 0px;
 			z-index: 10;
-			span {
-				position: sticky;
-				background-color: white;
-				left: 0px;
-				z-index: 10;
-			}
 		}
 		th:first-of-type:after {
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
Not quite there, yet. Removing `background-color` restores the table borders, but it results in overlap of the first column and subsequent columns on scroll. Still working on this.

image

brentryanjohnson:
Okay. I think I solved the overlap issue with `background-clip`. I had replced `inline-block` with `inline-grid`, which fixed the drop-down alignment, but it also (for some reason) canceled out `overflow: hidden`, with the result that the years select list would grow according to the number of selections. I used `float` to deal with that issue, and added a media query to display as blocks at 600px.

You'll also notice I eliminated many of the borders. In my view, they're unnecessary and take attention away from the content. I also eliminated the bottom row background color , as (1) it was the same color as the borders, which meant the borders no longer showed up and (2) bolding effectively delineates it from the preceding rows.

Issue #3954, opened 2019-04-16
Build updated filterable table design


mcharg:
https://t81vx1.axshare.com/#g=1&p=revenue_table_-_stats_plus

Differences from current state:
- Add drop-downs to select land category and location
- Land category rolls up the current land owner and source columns
- "organize by" changes to "first column"
- "add columns" changes to "second column" and users can only add one second column
- Second column that's added is broken out
- Shading on total rows

mcharg:
@jennmalcolm Should this go into the current sprint or wait until next sprint?

jennmalcolm:
I think it will need to wait until next sprint in looking at the workload. I did remove two filterable table issues from this sprint that I think we may need to discuss more before implementing. If @master12 has any extra time in this sprint, we could bring this issue in.

Issue #3923, opened 2019-04-12
Set Revenue type as the default "organize by" in the revenue filterable table


mcharg:
This will make it closer to what was displayed on the stats page.
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/7960bc14-dac6-4e49-9408-1b6353c45530)

PR #3974, opened 2019-04-19
Reorders 'organize by' drop down to place default at top

brentryanjohnson:
[:leftwards_arrow_with_hook: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/rev-table-index/explore/revenue)

- Resets default to 'commodity'

I'm not sure if defaulting to revenue type was intentional, but this PR reverts back to commodity if not. If the change was intentional, I think we should reorder the list (so the default option is at the top of the list).

### DIFF
--- a/src/pages/explore/revenue/index.js
+++ b/src/pages/explore/revenue/index.js
@@ -39,7 +39,7 @@ const TOGGLE_VALUES = {
   Month: 'month'
 }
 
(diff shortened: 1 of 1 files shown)


mcharg:
See #3923 and the order of the items in the drop downs will change as part of #3954.

brentryanjohnson:
No longer relevant, since the table format is changing.

PR #4040, opened 2019-05-05
Add Native American data to FY production file

### LINKED ISSUE #3981: Get Native American production data into FY homepage graphs
We don't have an annual data set for Native American production yet, and I know that the annual numbers on the homepage graphs are coming from those data sets. We can wait to get that data set and then incorporate the Native American numbers into the FY graphs. Alternatively we could sum the monthly numbers and use those, but I don't like this solution too much. We need to be consistent on what data we're using for which graphs. We have been using the annual data sets for the yearly graphs and the monthly data sets for the monthly graphs. I also created an issue to make this clear somehow on the home page (#3980).

brentryanjohnson:
Fixes #3981 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/na-fy-prod/)

- Concatenates FY Native American production data to file


brentryanjohnson:
The units are not uniform. For example, in federal production the product is listed as `Coal (tons)` and in the Native American production data, it's listed as `Coal (ton)`. That alone doesn't explain why the data isn't driving the charts, since the `Gas` and `Oil` units appear to be uniform. Let's look into it, @master12.

brentryanjohnson:
@master12 trimmed out some stubborn, adversarial white space. Ready to review.

mcharg:
@brentryanjohnson Is it just the homepage production charts that we need to look at?

brentryanjohnson:
Yes, @mcharg; just the homepage. For now, anyway.

Issue #3911, opened 2019-04-12
Clean up treatment of blank items in revenue filterable table


mcharg:
I'm not sure which pieces of this need to happen in the data file vs. the build.

No data for a year:
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/b33a2fa8-e784-43a5-8e70-b8f72b5cc339)

No commodity assigned and organized by commodity (displays last when organized by commodity):
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/e0588456-6f32-454d-b3fb-369495fd2295)

Location withheld and location not tied to a lease (displays last when organized by location):
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/043ad8e7-9760-40a8-8cf9-37859daae9ad)

Not tied to a lease when organized by source:
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/b62ce691-e971-4880-8195-3c0d7caf104d)

mcharg:
See #4051 for the issue around when there's "no data".

Issue #4051, opened 2019-05-07
Decide on treatment for fields in the filterable table when there's no value


mcharg:
The table currently puts in 0s, but displaying something like "no data" would be better.

mcharg:
We should also look at withheld data as part of this.  There's one idea in the production filterable table prototype.
https://wm3k5u.axshare.com/#g=1&p=production_table_1

mcharg:
What we decided:
- Only display rows when there is data for at least one year the user has selected.
- If there are years with no data in the user's selection, but numbers for other visible year, the years with no data get a dash.
- For production, don't display rows with only withheld data for the visible years, but have a checkbox that allows the user to opt to display rows with only withheld data.

Issue #3934, opened 2019-04-15
Make filterable table work in IE and Edge


mcharg:
They won't even open in IE and Edge.
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/043ec0e0-70c7-450d-b218-d6b57dd5d666)

mcharg:
The table doesn't even display in IE and Edge after #3954
![image.png](https://images.zenhubusercontent.com/5c589fe5e180620812ea22aa/9cd50a7e-6d03-4daf-9476-6985a44e1825)

mcharg:
No longer an issue.

PR #4052, opened 2019-05-07
3954 update filterable table

### LINKED ISSUE #3954: Build updated filterable table design

master12:
Fixes #3954 #3911 

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3954-update-filterable-table/explore/revenue/)

- Updated table to fix blanks
- Updated table to include grouping
- Updated table to use and apply button
- Updated filterable and group by options


mcharg:
I've created separate issues for tweaks, but the table isn't showing up at all in IE, Edge, (#3934) and mobile Safari (#4059).  We probably should fix that before releasing.

master12:
@mcharg It appears there is a loading issue with react-dom-min.js file. Refresh your browser a couple times to see if it loads. I will try to see if there is a solution.

mcharg:
@master12 I think the logic around not tied to a location is off.  It's showing up when I picked a single location.
image

master12:
@mcharg oh could catch on location. The data actually pulls from multiple columns for location so I shoudl change that. Need to check all columns or make a single source which is what I think i will do.

jennmalcolm:
@mentastc @Maroyafaied @lpgoldstein could 2 or all of you check some data?

Maroyafaied:
@jennmalcolm yes, I'll review it now.

Maroyafaied:
I'm seeing a difference in numbers for coal for Bonus's  2016 -2018
image
image

Maroyafaied:
Oil & Gas (Non-Royalty) is listed twice under Rents, the first row is accurate but not sure what the other row is pulling. 
image

master12:
@Maroyafaied You also have to include the Native American file. Which has $1 million in 2018 so that is the difference I think you are seeing.

master12:
@Maroyafaied The oil and Gas is from the native american file. there must be an extra space or something. i will get the numebr to merge. thanks

lpgoldstein:
Noticing that a several numbers are slightly off from the current version of the table. Is this normal? 
Example: other revenues 2018: new table $147,193,229 vs $147,193,234 in current table

Maroyafaied:
@master12 the coal numbers are fine after including the Native American file. Also, I'm not seeing offshore inspection fees on the table.

Maroyafaied:
The revenue type "other revenues" does not show when you select a specific state.

mentastc:
Looks like Inspection fees are not included in the filterable table making the bottom line totals off by around 40 million each year
image

mentastc:
Other Revenues not showing up when filtering by State and Revenue Type (also I don't believe that they show up when you're just looking at state)

image

master12:
> Looks like Inspection fees are not included in the filterable table making the bottom line totals off by around 40 million each year
> image

I see inspection fees?

image

Maroyafaied:
The inspection fees show on the general table but don't show when you select an offshore property. The same for other revenues, they show on the general table but not when you select a specific state.

jennmalcolm:
@Maroyafaied and @mentastc where the revenues are not linked to a specific state or onshore/offshore, they will not show up when sorted by that category. In the data, there are no locations linked to inspection fees. 

@master12 any way you could work in the treatment of blanks #3911 so that could be part of this iteration? Since even on our team we have trouble understanding the current treatment of blanks, this will likely be an issue for users without as much experience with the data.

master12:
@jennmalcolm The treatment of blanks is a part of this

mcharg:
@jennmalcolm There's one outstanding item with blanks, which is the cells with no data that are showing up as 0s, which Jeff said was a big fix, so we created #4051 to handle it later.

jennmalcolm:
Ok, @master12 sorry for not seeing that. @Maroyafaied and @mentastc  I don't think that the inspection fees should show up when you select a specific location since they are not tied to a specific location. Let us know if you see any remaining issues with other revenues for locations after Jeff fixed the logic for that.

Thanks, @mcharg that makes sense.

mentastc:
@jennmalcolm  when a sorting field is blank we could have it return that line as "not tied to [field name]" I like this because than our bottom line totals would always be the same.  But I do get the logic at if it doesn't have a location then it won't show up.

mcharg:
@mentastc That's essentially what we're doing with "not tied to a commodity" and "not tied to a location".  The difference is between sorting and filtering.  Sorting shouldn't change the total, but filtering will winnow the data down to just the items that match the filter selections.

mcharg:
I'm okay with approving this, but I can't find the approved button on my phone.  

@master12 There is a layout issue that would be nice to fix noted in a comment on the issue for the Native American note, but it shouldn't hold up release.

PR #4074, opened 2019-05-13
3892 revenue trends

### LINKED ISSUE #3892: Build revenue trends section on home page
Build revenue trends and move well/dollar graphic on home page

https://bghlq2.axshare.com/#g=1&p=a__replace_land_category_with_revenue_type

master12:
Fixes #3892 3827

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/3892-revenue-trends//)

- Added revenue trends to homepage
- Created a sparkline component
-


brentryanjohnson:
☝️ Commit applies [content guidance](https://github.com/ONRR/doi-extractives-data/wiki/Content-guide) to revenue trends.

mcharg:
@master12 @brentryanjohnson Can we bold the "Total revenues" and "$4.20 billion" text for the total line?

mcharg:
@master12 This is what it looks like in IE.
image

mcharg:
@master12 Edge is also weird.
image

mcharg:
Chrome, Firefox, iPhone Safari, and Android all look good.

brentryanjohnson:
Shouldn't the total add up to $4.3 billion, since even adding just bonuses to the royalties figure would push the rounding up beyond $4.2?

master12:
Everything is rounded but the rounded numbers arent added so hence the weirdness

master12:
We are not supporting IE so let stop testing it

mcharg:
Then we should put in the message that says we're not supporting IE.

mcharg:
This build is doing funky things to the disbursements chart on Explore Data.
image

mcharg:
The revenue table also isn't showing up in this build.
image

master12:
revenue table is another PR

mcharg:
The Edge fix appears to have fixed both Edge and IE.

mcharg:
I'm good with limiting to one decimal.

mcharg:
We might also want to clean up the alignment and either align them all on the left of right of the column.  I'm not sure what the aligning principle is right now (looks like the header might be center, the dollar amount left and the change right?).

image

mcharg:
@master12 The tabs and graphs aren't showing up in this latest build.  I refreshed a ton of times and that didn't help.

image

master12:
@mcharg Send me the log

mentastc:
For fy19 I have 6.2 billion so far, the site shows 4.2

image

mentastc:
Is this including civil Penalties and Inspection fees? this wouldn't make up the 2 billion difference but I'm curious where that revenue is, maybe in other revenues would work.

mcharg:
@mentastc I think it makes sense to roll up everything that's not included in the other categories into "other revenues".

jennmalcolm:
In looking at the numbers, there is a problem. Here are the numbers I got. 
image
I think the issue is that October is not being included. @master12 are you starting the fiscal year calculations with October and going through the most recent (March for now)? When I compared to the numbers without October, they seem to match.

I also think the trend lines are not accurate. Revenues have gone up in recent years. They were the lowest in 2016 but then have gone back up. Is the year range you're using 2009-2018 (the most recent full year).

When I did the spark lines in Excel for fiscal years 2009-2018, here is what they looked like:
image

mentastc:
@master12  have you had a chance to look at difference that I'm seeing?

master12:
@mentastc @jennmalcolm I was not including October. I was using greater than 9 for the month but actually the months start with 0 not 1 so I was starting with november. I will upload that fix

master12:
@mentastc @jennmalcolm Ok, found the issue with the sparkline. Was only including the filtered 2018 data which was only Oct - March. Data is now copied so the filter wont effect the sparkline full year data. It should match Jennifer's sparkline now.

mentastc:
@master12  your other revenues are lower than Jennifer's is are you missing inspection fees and civil penalties?

master12:
@mentastc I can do that but I seriously thought that was a joke. But I will make the change

mcharg:
@brentryanjohnson Do we need the negative sign after the down triangle?  It feels like a double negative.

brentryanjohnson:
I think we need it @mcharg, because screenreaders don't know the semantic meaning of the down arrow. They need to read out the minus. I'm glad you pointed it out, because it's an issue for the positive arrows, too.

brentryanjohnson:
FYI...I added a `title` attribute to the arrow `SVG`s to surface their meaning to screenreaders ("increase" and "decrease", respectively).

Without the attribute, their meaning can only be determined visually.

brentryanjohnson:
@mcharg Even though I've added an attribute for accessibility, I think the minus symbol is more explicit for now, perhaps lending itself to quicker scanning and understanding of the numbers. I want to test it with a screenreader, too, to determine if the double negative that results is ambiguous when read aloud. I think we can ship this and then revisit over time. 👍

PR #4077, opened 2019-05-13
Heavy, chunky font

brentryanjohnson:
Fixes #<PUT ISSUE NUM HERE>

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/heavy-ugly-font/)

-
-
-

### DIFF
--- a/src/css-global/colors.scss
+++ b/src/css-global/colors.scss
@@ -9,7 +9,7 @@
 	grayDarker: #5c737f;
 	grayDarkest: #435159;
 	blackLight: #323c42;
-	black: #1c2326;
+	black: black;
 	grayBlueDark: #4e4e4e;
 	grayBlueMid: #a9a9a9;
 	grayBlueMidLight: #e0e0e0;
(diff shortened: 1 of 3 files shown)


mcharg:
@brentryanjohnson You're right, bold does become hard to differentiate.
image

mcharg:
@brentryanjohnson I'm not finding bold text in any other location on the site.  Maybe we just address the few spots where we use it by using other means to give the content more weight?

brentryanjohnson:
I'm still working on this branch, @mcharg, but there will be more instances where the problems emerge. I'll let you know when I've increased the weight of all instances of "light" to "regular".

jennmalcolm:
How difficult will it be for you to put together a branch to preview changing from almost black to black to see if maybe that will address the concerns without having to go to such heaving text?

brentryanjohnson:
Color tweak (only) is in #4079, @jennmalcolm.

brentryanjohnson:
I've changed the default body font weight from `500` in the previous iteration to `400` to determine if that helps preserve emphasis contrast.

mcharg:
Looks a little better.

PR #4100, opened 2019-05-16
Changes revenue trends arrow style and colors to limit editorializing the values

brentryanjohnson:
[ ↕️  PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/neutral-arrows/)

I think the red and green arrows are a little too heavy-handed. In other words, the colors imply more than what I think we should be saying (green=good, red=bad). 

While I don't think we should be implying "good" or "bad" on the site, I also think they might be misleading. For instance, implying that rents are "in the red" might look bad to someone who thinks we should be constantly increasing revenue, but it might simply mean that land moved into a royalty-producing status more quickly than the previous year. In other words, a lesser value compared to the previous year doesn't necessarily mean "bad," but the color red does imply "bad."

I'm exploring alternatives in this branch that will still differentiate the two, but not imply "good/bad" values in the same way.


mcharg:
My only concern is that these may now look clickable, since we use blue for links and buttons.

brentryanjohnson:
That's a good point, @mcharg. Let me shift them back to a black ramp and see what you think.

mcharg:
I wonder if we need to explore different arrow styles.  Without the color these are very similar to arrows used on many sites for accordions and sorting.

brentryanjohnson:
FYI...the branch is building now with alternative arrow styles.

mcharg:
Can we try aligning the arrow line with the top of the sparkline, so it gets better tied to the dollar amount above?  Right now they're evenly spaced with the row above and below, making it difficult to tell which is tied with which.

Current build:
image

Proposed:
image

brentryanjohnson:
I also need to fix the numbers dropping down on mobile. I'm not sure why this is happening.

Revenue trends table with arrows indicating up and down values compared to last year, with some numbers dropping down

mcharg:
@brentryanjohnson, @master12 is also working on some fixes to the trends section (#4107), so you may want to coordinate to avoid any merge conflicts.

master12:
closing this since this has been merged with other revenue trend changes

PR #4113, opened 2019-05-22
Lighthouse accessibility audit fixes

### LINKED ISSUE #4067: Fix accessibility issues from Lighthouse audit
## Lighthouse results

- [ ] [id] attributes on the page are not unique - The value of an id attribute must be unique to prevent other instances from being overlooked by assistive technologies.

```
Failing Elements
--
rect#backgroundRect
g#maxExtent
g#bars
g#groups
```

- [x] Lists do not contain only `<li>` elements and script supporting elements (`<script>` and `<template>`). Screen readers have a specific way of announcing lists. Ensuring proper list structure aids screen reader output.

```
Failing Elements
--
ul.header-nav_top
```

- [x] `<html>` element does not have a `[lang]` attribute - If a page doesn't specify a lang attribute, a screen reader assumes that the page is in the default language that the user chose when setting up the screen reader. If the page isn't actually in the default language, then the screen reader might not announce the page's text correctly.
```
Failing Elements
--
html
```

- [ ] Form elements do not have associated labels - Labels ensure that form controls are announced properly by assistive technologies, like screen readers.

```
Failing Elements
--
select
select
<select><option value="calendar_year">Calendar year</option><option selected="" value="fiscal_year">Fiscal year</option></select>
```

brentryanjohnson:
Partially Fixes #4067

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/lighthouse-accessibility-audit/)

- Checklist in #4067 

I'll need to cowork with @master12 to fix a couple of these React-prop-generated problems.

### DIFF
--- a/src/components/layouts/DefaultLayout/DefaultLayout.js
+++ b/src/components/layouts/DefaultLayout/DefaultLayout.js
@@ -35,6 +35,7 @@ const DefaultLayout = ({ children }) => {
   return (
     <div>
       <Helmet
+        htmlAttributes={{ lang: 'en' }}
         meta={[
           { name: 'google-site-verification', content: 'OxyG3U-Vtui-uK6wHUeOw83OgdfcfxvsWWZcb5x7aZ0' },
           // Mobile Specific Metas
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
@jennmalcolm Since this PR has been sitting for awhile, I suggest we merge this one (with review) and then move on to the other issues.

jennmalcolm:
Ok @brentryanjohnson should we have @master12 review this?

brentryanjohnson:
Couldn't hurt, but there shouldn't be anything in this one that requires his review, @jennmalcolm. It's just a `meta` tag and the removal of a `span`.

PR #4116, opened 2019-05-24
4107 change revenue trends

### LINKED ISSUE #4107: Add date revenue trends was last updated
So it's obvious what months the current year so far includes.

master12:
Fixes #4107 4097 4100

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/4107-change-revenue-trends/)

- Updated sparkline to go 10 years from latest date
- Updated text
- Updated styling to shrink table
- Removed duplicate sparkline


jennmalcolm:
Shouldn't the most recent month be April 2019? That's what we have the data for in the file and in the revenue monthly graph? 
image

master12:
Good catch @jennmalcolm . The data and code were all up to date, The issue was a "timezone". Federalist runs apparently on the west coast so when your browser on the east coast loaded the page and converted the timestamp to locale it was converted 4 hours earlier. We now use timezone UTC for the conversion so it will be universal.

PR #4167, opened 2019-06-12
Monthly Production and Rev Update 2

mentastc:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/monthly_update_revenue_production_june_2/)

-updates rev charts and downloads
-updates production charts and downloads
-


jennmalcolm:
@lpgoldstein reviewed all monthly charts and the downloads files. I'm adding @Maroyafaied as well. @Maroyafaied, could you also check that the monthly trend numbers are correct based on the new data?

PR #4178, opened 2019-06-17
Adds 'by year' to annual download page titles

brentryanjohnson:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/downloads-structure/downloads)

- Adds 'by Year' to annual Revenue and Disbursements page titles to match Production


jennmalcolm:
This is good, but now I noticed that the order of year/month is different in disbursements than in revenue and production.

brentryanjohnson:
Good catch, @jennmalcolm. Reversed in this commit ☝️.

PR #4217, opened 2019-07-03
Adds search field and results index

### LINKED ISSUE #3849: Implement search
After Gatsby switch over, we will need to implement a new search functionality

brentryanjohnson:
Fixes #3849

[🔍 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/search-lunr/)

- Reimplements search
- Adds indexing package to Gatsby
- Styles search
- Creates search results page

This is an MVP of search. Next steps:

- Write intro to convey scope of search
- Add "no results" line of text if search produces no result(s)
- Look into adding description from front matter to result list
- Further style results


jennmalcolm:
This works in Chrome, Firefox, and Safari on my phone, but I'm not getting any search results in Edge or IE. I tried a few terms that all returned results in the other browsers and always got blank results.

image

brentryanjohnson:
@jennmalcolm When this build finishes, could you test again in Edge? As long as it works in Edge, we should be good to ship.

jennmalcolm:
Yes, it is working in Edge (and IE) now.

PR #4257, opened 2019-07-16
Adds copy for no search results

### LINKED ISSUE #4256: Add copy indicating that no results were found in search
It's a bit jarring for there to be nothing there.

brentryanjohnson:
Fixes #4256

[🔍 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/search-no-results/)

- Adds conditional to produce copy indicating no search results yielded
- Adds a link to glossary for secondary search

This PR adds an explanation for no search results, but it also adds the search term as a link to open the glossary with that term. This is not necessary, but it could be useful in such cases as "BLM," which currently doesn't produce any search results, but for which we have a glossary term entry:

We didn't find any results for your search. You might want to try searching for blm in our glossary, with open glossary panel with blm defined as The Bureau of Land Management (BLM) is part of the U.S. Department of the Interior, and manages exploration, development, and production of natural resources on federal lands

Alternatively, we could simply produce the line:

> We didn't find any results for `{ queryString }`.

which, in this case, would resolve to:

> We didn't find any results for blm.

### DIFF
--- a/src/pages/search-results/index.js
+++ b/src/pages/search-results/index.js
@@ -56,7 +56,7 @@ const SearchResults = () => {
                     results.map((item, index) => {
                       return <li key={ index }><Link to={ item.path }>{ item.title }</Link></li>
                     }
-                    ) : <p><strong>Sorry, we didn't find any results for your search.</strong></p>
+                    ) : <p><strong>We didn't find any results for your search.</strong> You might want to try searching for <GlossaryTerm termKey={queryString}>{queryString}</GlossaryTerm> in our glossary.</p>
                   }
                 </ul>
               </article>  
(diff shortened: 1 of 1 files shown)


brentryanjohnson:
I'm going to look into only surfacing the glossary line if the term is in the glossary. Hold off merging, please, @jennmalcolm.

brentryanjohnson:
I'm probably not going to get to the add-on this sprint after all, @jennmalcolm, so if you want to merge this, I can work on the extras next sprint.

jennmalcolm:
Ok @brentryanjohnson, will do.

PR #4401, opened 2019-07-31
Adds glossary filter to search terms to conditionally print term result

brentryanjohnson:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/search-gloss-filter/)

- Adds filter to only print glossary copy if search term is in glossary

### DIFF
--- a/src/components/utils/Glossary.js
+++ b/src/components/utils/Glossary.js
@@ -105,7 +105,7 @@ class Glossary extends React.Component {
 
 function filterGlossaryTerms (glossaryTerm) {
   let numOfTermsToShow = 0
-  if (glossaryTerm !== undefined && glossaryTerm !== '') {
+  if (glossaryTerm !== undefined && glossaryTerm !== null && glossaryTerm !== '') {
     GLOSSARY_TERMS.forEach(term => {
       term.show = false
       if (term.name.toLowerCase().includes(glossaryTerm.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


jennmalcolm:
Did you know this build failed @brentryanjohnson?

brentryanjohnson:
No, @jennmalcolm. I'll take a look.

brentryanjohnson:
I fixed the build problem, @jennmalcolm (the CircleCI failure is separate, related to the unit tests). You should be able to preview now.

There is a one minor problem with a flash of text that I'll try to fix down the road.

mcharg:
@brentryanjohnson Now that we know the term is in the glossary, would it make more sense to say something like "BLM is in our glossary." instead of "you might try searching..." to make it clear we've already done the work and there's no question that it's in the glossary?

jennmalcolm:
This is looking good, and keeps getting better. One last thing that I think would be nice (if there's a method that you could do it easily) is to still print the glossary link for acronyms like NGL and EIS, which are in the glossary as parentheticals with their terms.  I don't think we should use a partial match like the glossary serach does, but perhaps, you could make manual exceptions for these? Here is the list of acronyms that this situation applies to in case you have extra capacity to work on this more when you're dealing with the flash: CY, EIS, FY, IA, MSG, NAICS, NGL, RAC

This is not a must have, so I am going to approve. It is already a big improvement.

PR #4453, opened 2019-08-15
Removes revenue section from dives, links to other revenue rates

brentryanjohnson:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/removes-fees-section/how-it-works/minerals/)

- Removes rates, revenues section from all dive-in pages.
- Rationale: We maintain rates and fees content in three separate places, and the rates change with some frequency. This approach removes one instance on an otherwise static, contextual page, and links to another place in `how-it-works` where we recently updated the content.


lpgoldstein:
@brentryanjohnson rates are still showing under onshore oil and gas https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/removes-fees-section/how-it-works/onshore-oil-gas/

brentryanjohnson:
Good catch, @lpgoldstein. Should be fixed with this commit ☝️.

PR #4565, opened 2019-09-10
Update federal revenue by company text

lpgoldstein:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/BRANCH_NAME/)

- revenue by company text is supposed to be in italics but there was a mistake in the html

image


brentryanjohnson:
Looks good!

I'm not sure why these had to be refactored from markdown to HTML, but it looks like it was done during the Gatsby migration. We'll probably find a few more of these at some point.

PR #4669, opened 2019-10-08
Update FY Revenue Data Set

### LINKED ISSUE #4645: Picky about rounding
The grand totals (depending on grouping) on the filterable table vary by about $20. Could we do the rounding later in the process or something to get these closer? Also, the numbers I looked at when compared to Excel all seemed to be rounded down even when up was closer, so we should just review our rounding protocol for that table.

mentastc:
Fixes #4645 
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/fy_2019_rev_file/)

-Updates FY homepage graph
-Updates FY Rev file (xlsx & CSV)
-Updates download language

Data to check
Revenue Fiscal Year (fiscal_year_revenue)	
- [x] Homepage: FY bar charts
- [x] How it works: Native American revenue table 
- [x] Explore data: Filterable revenue data table 
- [x] Download: xls and csv files updated with correct dates and file size


master12:
Ok it now rounds but it also rounds negative numbers so you will still get some variance

mentastc:
@master12 with the rounding its still adding up the rounded subtotal numbers and then totaling those for the grand total at the bottom, this makes the grand total change based the sort.
What the grand totals should be for 2017-2019:
image
What they are for Rev/commodity sort:
image

What they are fro a location/commodity sort:
image

mentastc:
@master12, that looks to have done the trick

PR #4675, opened 2019-10-10
State page intro

### LINKED ISSUE #3277: Restructure state summary content
Related to #3038 and #3123 

Observing our data publication schedule, I'd like to start reformatting the state-page intros to eliminate the dependence on GDP data. We're planning to eliminate the Economic Impact section and associated data when we publish FY data in November. The economic data isn't specific to federal lands and waters, and thus doesn't align with our updated [product vision](https://github.com/ONRR/doi-extractives-data/wiki/Product-framing#our-vision).

To prepare for the removal of that data, we need to replace content that is driven by it. One obvious example of this is the state-page summary, in which GDP data drives state-level content.

I've worked on a possible prototype for this. This is fairly straight forward for Wyoming, but we'll need to draft and evaluate an example for states for which there is no production, revenue, or disbursements.

We should also consider a summary that shows the portion of state production from federal land versus all other categories of land ownership.

State page summary content

mojobnichols:
Fixes #3277

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/state-page-intro/)

Summarizes data without GDP and other external data

-
-


brentryanjohnson:
I think this works overall, but sometimes there are some awkward collisions. For example, the [Minnesota](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/state-page-intro/explore/MN/) intro lists no production, but does have both revenue and disbursements for the year. That may be confusing for some users, but we probably can't do much about that.

mcharg:
How does Minnesota have revenue without production? Is the data withheld or is it carryover from a previous year's production? Can we explain it for the user?

mcharg:
For states with disbursements to the county are both the state and county level disbursements included in the amount we say went to the state? We might want to clarify that it went to either the state or counties, if the counties are included.

mcharg:
Is the 4 included in the 11 or in addition to it?
image

brentryanjohnson:
I believe the four are included in the total number of commodities, @mcharg. The volumes are withheld, but the commodities are listed. That's one of the other lines we need to tweak.

jennmalcolm:
@mcharg Minesota's revenue was the result of Minimum or estimated royalties, settlements, and interest payments. Not all revenue requires production (e.g. bonuses, rents). That does mean that perhaps we should re-phrase the revenue bullett to state "Leasing or  production..." I don't know if that's too complicated, and maybe just "production" works since companies wouldn't be leasing without the intention to produce and permit fees and penalties are all related to production as well.

@mojobnichols and @brentryanjohnson, it doesn't seem like the number of commodities for 2018 is working quite right. Take a look at PA. It only has gas in 2018, but it says it has 2 commodities in 2018.

mcharg:
I think it does make sense to call out leasing or production to better indicate where revenue comes from.

brentryanjohnson:
I think the number of commodities issue is due to not constraining the commodity list to the current year. @mojobnichols 

```
const StateProductionSummary = props => {
const commodityCount = Object.keys(props.production).length
const currentYear = props.productionYears[props.productionYears.length - 1]
const withhelds = Object.keys(props.production).filter(commodity => props.production[commodity].volume[currentYear] === 0)
```

PR #4684, opened 2019-10-15
Land and Water Conservation Fund authorization content updates

brentryanjohnson:
[🌲 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/lwcf-authorize/how-it-works/#disbursements)

- Removes 'update' flags from LWCF content
- Removes references to 'when authorized,' since the fund is now permanently authorized

### DIFF
--- a/src/data/terms.yml
+++ b/src/data/terms.yml
@@ -12,6 +12,8 @@
   definition: "A yearly maintenance fee for maintaining a claim."
 - name: "APD"
   definition: "Application for permit to drill"
+- name: "Appropriation"
+  definition: "There are two main congressional actions that result in federal spending: authorization and appropriation. A fund or recipient may be authorized to receive money during the federal budget process, but Congress must still designate a specific amount to be distributed to the fund or recipient. This process is called "appropriation"."  
 - name: "Authorization"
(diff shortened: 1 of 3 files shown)


jennmalcolm:
Your changes look good, but when reading through it, I thought we could be a bit clearer about appropriation. Authorization allows the fund to receive funds, but fund need to be appropriated to be spent (through grants, etc.). Also, why do we have authorize in the glossary but not appropriate? Could we add appropriate? I know it will take some work to get that clear, so I could create a new issue for that and merge this PR. Let me know what you think @brentryanjohnson.

brentryanjohnson:
Okay, @jennmalcolm. I've added "appropriation" to the glossary, and glossarized the term "appropriate" in the _appropriate_ paragraph. 😉 

This reveals the challenge of balancing technical terms and plain language, but hopefully this comes close:

> There are two main congressional actions that result in federal spending: authorization and appropriation. A fund or recipient may be authorized to receive money during the federal budget process, but Congress must still designate a specific amount to be distributed to the fund or recipient. This process is called "appropriation".

jennmalcolm:
The appropriation addition to the glossary did not build. Error log:

Generating JavaScript bundles failed

can not read an implicit mapping pair; a colon is missed at line 16, column 351:
     ... ss is called "appropriation"."

Could you fix @brentryanjohnson?

brentryanjohnson:
@jennmalcolm The straight quotes around "appropriation" were breaking the YAML syntax. It should be good now.

PR #4696, opened 2019-10-17
Revert "Update FY Revenue Data Set"

jennmalcolm:
Reverts ONRR/doi-extractives-data#4669

So that we can do a release with the blog post and other items but not the revenue updates.


jennmalcolm:
Closing this since we will likely release these numbers tomorrow.

PR #4714, opened 2019-10-21
Revert "Update Monthly Production and Revenue October 2019"

jennmalcolm:
Reverts ONRR/doi-extractives-data#4661

So that we can do a release with the blog post and other items but not the revenue updates.


jennmalcolm:
Never mind. Now it looking like we might release this all tomorrow.

PR #4728, opened 2019-10-21
Add "Not tied to lease" category to homepage revenue graph

### LINKED ISSUE #3929: Consider adding "Not tied to a location" category to revenue homepage graph
We should consider whether we need to add a new category to the revenue home page graph so that we do not confuse people when the totals do not equal the totals on the filterable table or the downloads page.  For the revenues that are blank in the onshore/offshore category, should we use "not tied to a lease federal" or something similar to what we will use on the filterable table?

I made the decision to leave non-royalty, bonus or rent revenues out of the graph rather than stick them into one of the categories we have, but if we can add a category so that the total is accurate, I think that would be helpful. 

There doesn't seem to be a totals problem for the disbursements graph, but I'm not sure what methodology is used to break those numbers down fully when presumably the source data for those categories are also coming from the royalty reports. I guess we could explore whether a similar methodology would allow us to fully break down revenues into the given categories.

image

Alternatively, we have heard the suggestion that it might be nice to breakdown the revenues by revenue type. I'm not sure if we should go down this road since we would lose the consistency with the other graphs, but I'm open to others' opinions.

jlouis-onrr:
Fixes #3929

[preview](https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/onrr/doi-extractives-data/3929-ConsiderAddingNotTiedToALocation/)

- Adds "Not tied to a lease" data row to homepage Total Revenue Table
- Adjusts color range for each source

### DIFF
--- a/src/components/charts/ChartLegend_README.md
+++ b/src/components/charts/ChartLegend_README.md
@@ -12,7 +12,7 @@ The Chart Legend is legendary. Ryan and/or Shannon please add a description.
 
 ## How to use
 
-The legend can be used with any chart to display the values of each fo the data poitns with a calculated sum for the total. 
+The legend can be used with any chart to display the values of each of the data points with a calculated sum for the total. 
 
 ### Chart Data:
 This is an array of objects that will be used to populate the legend
--- a/src/components/layouts/charts/StackedBarChartLayout/StackedBarChartLayout.js
+++ b/src/components/layouts/charts/StackedBarChartLayout/StackedBarChartLayout.js
@@ -1,93 +1,119 @@
-import React from 'react'
-import PropTypes from 'prop-types'
-import MediaQuery from 'react-responsive'
+import React from "react";
+import PropTypes from "prop-types";
+import MediaQuery from "react-responsive";
 
-import styles from './StackedBarChartLayout.module.scss'
+import styles from "./StackedBarChartLayout.module.scss";
 
-import { ChartTitle } from '../../../charts/ChartTitle'
-import { StackedBarChart } from '../../../charts/StackedBarChart'
-import { ChartLegendStandard } from '../../../charts/ChartLegendStandard'
-import { Accordion } from '../../Accordion'
+import { ChartTitle } from "../../../charts/ChartTitle";
+import { StackedBarChart } from "../../../charts/StackedBarChart";
+import { ChartLegendStandard } from "../../../charts/ChartLegendStandard";
+import { Accordion } from "../../Accordion";
 
-import utils from '../../../../js/utils'
+import utils from "../../../../js/utils";
 
-const MAX_CHART_BAR_SIZE = 15
+const MAX_CHART_BAR_SIZE = 15;
 
 class StackedBarChartLayout extends React.Component {
   state = {
     dataSet: this.props.dataSet,
-    barHovered: this.props.barHovered,
+    barHovered: this.props.barHovered
+  };
+
+  componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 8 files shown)


mcharg:
@jlouis-onrr This looks good. We should change the colors on Production and Disbursements, as well, so we're always using the same color for federal onshore, federal offshore, and Native American.

jlouis-onrr:
@mcharg great point, I think that was overlooked.  Ill update those as well.

brentryanjohnson:
I understand why we're adding this revenue line item to the table (so the revenue total is a "true" total), but I don't understand why we're adding it to the stacked charts. The value is so low relative to the others and the total, there's nothing to chart. There's no data to visualize (unless I'm missing something).

mcharg:
@brentryanjohnson It feels like we should have a color for everything that's in the table, so that if it does spike to a level that it does chart, we have a way to show it. 

@jlouis-onrr The hover colors also need to be updated to reflect the new shades.

jennmalcolm:
@mentastc @brentryanjohnson and @mcharg please review this PR ASAP to see if we can get it out with the release of the numbers, which will likely go tomorrow.

jlouis-onrr:
@mcharg it appears that currently the same colors are being used for the hover as the selected values, but the opacity drops a a couple of points from 1 down to .8, how would you like to see the hover event changed reflected?

federal onshore: #032435
federal offshore: #086996
native american: #79bbda
not tied to a lease: #c6e9ff

mcharg:
@jlouis-onrr The federal onshore looks like it's dropping to a lighter shade than the others, but if it's doing what it's supposed to do, it's fine.

jlouis-onrr:
@mentastc - reordering of the list would require rearranging the colors in order bottom to top or dark to lightest.  I think since the "not tied to a lease" is at the top and doesnt have enough value to add to the chart, I would think it could be confusing to the end user to place it somewhere other than the top or bottom.

Check out @brentryanjohnson previous comment:
> I wonder if we might cause confusion by showing in the legend a value that isn't charted, and likely won't be because of the context. It's just 0.02% of the total, and I suspect that line item's relative share of revenue is unlikely to be significant enough to merit visualizing at that scale very often.
> 
> I approve merging, but I'd like to get user feedback about this at some point (compared to listing the line item and value in the table with no corresponding legend swatch, or another solution).

brentryanjohnson:
I showed this to a few ONRR folks today, and they weren't sure at first whether the light blue on the stacked chart was the Native American value or the "not tied to a lease" value, somewhat confirming my concerns. 

It's not a formal setting or representative sample, but it was clear from the limited testing that showing an item in the legend without a corresponding charted value caused some initial confusion.

jennmalcolm:
Could we just take the swatch off of that one but keep the number in the table?

mcharg:
And if we put it last it'll be grouped with the other federal items.

jennmalcolm:
@jlouis-onrr what do you think about taking off the swatch and moving the Federal- not tied to a lease line down to the bottom? @mcharg I don't think the not tied to lease numberwill likely ever get high enough to appear on the chart.

jlouis-onrr:
I think moving to the bottom could work, but more importantly the lack of a swatch or no color on the chart makes it confusing to the end user.  Maybe an asterisk or tool tip that could explain to the end user that the line item doesn't have enough data to show on chart, but is still part of the data-set could be helpful?

jennmalcolm:
I will go ahead and merge since we want the totals to match when we highlight the FY 19 data. I will create another issue to address potential confusion about having a swatch for something that will never be visible on the graph.

PR #4751, opened 2019-10-25
Fy disbursements again

jennmalcolm:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/FY-disbursements-again/)

- replaced xlss FY disbursements file - only copied and pasted 2019 into old file
- changed file size
-
This seems to be only a partial solution as the homepage bar chart and state pages did not update to 2019, but the explore data page did update and the file goes through 2019. @lpgoldstein and @mentasti, please check the data in the explore data disbursments chart and the file in case this is the best we can do this morning. 

Disbursements Fiscal Year (disbursements)

- [x]  Explore data: by recipient table

- [x]  Download: xls updated with correct dates and file size"


jennmalcolm:
@mentastc and formatting changes you made to the 2003-2018 data won't be in here since I only added 2019 to the file.

PR #4774, opened 2019-10-31
Refactors state page intros

### LINKED ISSUE #3277: Restructure state summary content

brentryanjohnson:
Fixes #3277

[🐌 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/state-intro-again/explore/PA)

- Refactors state page intros (again)

### DIFF
--- a/src/components/locations/SectionOverview.js
+++ b/src/components/locations/SectionOverview.js
@@ -11,13 +11,50 @@ import FederalLandOwnershipLegend from '../maps/FederalLandOwnershipLegend'
 import FederalLandOwnershipSvg from '../maps/FederalLandOwnershipSvg'
 import utils from '../../js/utils'
 import styles from './SectionOverview.module.scss'
+import GlossaryTerm from '../utils/glossary-term.js'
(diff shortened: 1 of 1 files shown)


brentryanjohnson:
The file still contains the faulty query. I might just need to rewrite the code.

PR #4797, opened 2019-11-13
Refactors state-page summaries

brentryanjohnson:
[🐢 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/state-intro-again/explore/PA)

- Refactors state page intro summary to remove deprecated GDP data and replace with state summary of production, revenue, and disbursements

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -404,6 +404,7 @@ const createStatePages = (createPage, graphql) => {
             }
           }
         }
+
       }`
 	      ).then(result => {
 	        if (result.errors) {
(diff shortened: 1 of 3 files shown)


mcharg:
It feels kind of weird that production and revenue are 2018 and disbursements are 2019.  Do we need to better explain how production in 2018 results in disbursements in 2019?
image

brentryanjohnson:
I know that's a bit awkward, @mcharg, but we'll have a similar issue with the data in detail, since we get it at different times. I'm not sure the best way to handle it...the purpose isn't to try create a relationship for a single year, but rather a relationship between the different phases (production leads to revenue, revenue leads to disbursements). We should discuss more and iterate.

jennmalcolm:
Since disbursements are only in fiscal year and the other two are in calendar year, that does create a problem when trying to tie them together. We will have CY 2019 disbursements soon (January), so we could use those in the intro, but then those wouldn't be tied to the disbursements section on the page. Also, we'll be changing the explore data relatively soon, so it's probably not worth much more effort on this for now.

PR #4805, opened 2019-11-15
"Fixes" double production units on homepage tab

### LINKED ISSUE #4733: Fix duplicate units on production
We have duplicate units on production values on the homepage.

Units for production with duplicate labels, bbl, bbl, mcf, mcf, tons, tons

brentryanjohnson:
Fixes #4733

[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/double-vision-units/)

- Fixes duplicate production units

### DIFF
--- a/src/components/layouts/charts/StackedBarChartLayout/StackedBarChartLayout.js
+++ b/src/components/layouts/charts/StackedBarChartLayout/StackedBarChartLayout.js
@@ -106,14 +106,11 @@ class StackedBarChartLayout extends React.Component {
       longUnits,
       xAxisLabels
     } = this.state.dataSet;
-    let titleUnits = "";
-    if (showUnits) {
-      titleUnits = " (" + longUnits + ")";
-    }
+
     return (
       <div className={styles.root}>
         <ChartTitle>
-          {title} {titleUnits}
+          {title} {" (" + longUnits + ")"}
         </ChartTitle>
 
         {this.state.dataSet && (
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
This is working for the Fiscal Year/Calendar Year switch, but not for the Yearly/Monthly switch. More to do...

brentryanjohnson:
This should be working now.

jennmalcolm:
Now the units are missing for the monthly charts. Was that the intention?

brentryanjohnson:
Definitely not, @jennmalcolm. They show up for me...I'll check to see if that's just cache.

Production values for oil, gas, and coal by month for a 10-year period with units showing

jennmalcolm:
Ok, Ithe duplicate issue does seem "fixed," but the units above the graphs are missing for monthly. Also, is there a reason that on the live site we spell "barrels" out for yearly graphs and use "bbl" for the monthly graphs. 
image

brentryanjohnson:
Ah, I see @jennmalcolm. I'll look into the missing top labels.

There are two forms for the units, referred to in the code as `longUnits` and `shortUnits`. I'll see if I can work the long units into the top label.

brentryanjohnson:
Okay, @jennmalcolm, I think I've fixed the missing title unit. I'm not sure why it shows long units on one view, and short units on the other, so I can look into that more. Also, I'm not sure that we need the units in three places on this chart, so the title units can probably be omitted (thoughts?).

jennmalcolm:
Yes, the missing units are back now. I don't think it's necessary to have the units in three places, the units with the max dotted line are also inconsistent on the oil graph (barrels with annual and bbl with monthly) I prefer using barrels here for people who may not know the bbl abbreviation. I wonder if it would also make sense to spell out "thousand cubic feet" on the gas line?

lpgoldstein:
@jennmalcolm @brentryanjohnson I think it would make sense to spell out the units as well. Someone recently brought up that she/he didn't understand mcf. I think it was in a user interview.

PR #4809, opened 2019-11-18
fixed sparkline for revenues, #4489

### LINKED ISSUE #4489: Define what data the revenue trend line should use
It's currently using a mix of FY and monthly data.  We should decide whether we want it to use either all FY or all monthly.

master12:
Fixes #4489

[:sunglasses: PREVIEW](https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/onrr/doi-extractives-data/4489-revenue-trend-line-fix/)

- Changed the trend line to be the previous 10 years unless the current year is complete
-
-

### DIFF
--- a/plugins/custom-excel-transformer/data-transformers/revenues-transformer.js
+++ b/plugins/custom-excel-transformer/data-transformers/revenues-transformer.js
@@ -93,7 +93,7 @@ const createRevenueNode = (revenueData, type) => {
   let year = revenueNode.CalendarYear || revenueNode.FiscalYear
   let month = (revenueNode.Month) ? getMonthFromString(revenueNode.Month) : 0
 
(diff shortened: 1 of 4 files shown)


jennmalcolm:
This still has 2020. Shouldn’t it end with 2019?

master12:
This is what I see locally. Will try again:
image

master12:
Ok I think I found the issue. Since the date was set to the first of the month at midnight, on the server which is a different time zone from my laptop, it was sending back Sept 30th vs Oct 1 so that was the issue. I changed the date to use the 15th of the month since we don't care about the day and this should resolve the issue.

master12:
@jennmalcolm Ok this looks good now.

PR #4818, opened 2019-11-21
Adds explanation for missing 2006 HPF disbursement

### LINKED ISSUE #4779: Historic Preservation Fund Disbursement missing in 2006
The excel download has a dash for 2006 for the Hisitoric Preservation Fund, and it shows up as zero in the explore data page and the query tool. I'm not sure if this has been verified that there was no disbursement or if this data is missing?

image
image

brentryanjohnson:
Fixes #4779

[💸 PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/hpf-2006/explore/#federal-disbursements)

- Adds explanation for 2006 HPF to `yaml` file


brentryanjohnson:
FYI @master12 and @jennmalcolm, I had to remove the Historic Preservation Fund row for 2006 in the spreadsheet for this to work, presumably because the cell wasn't empty (it had a `-`). We might want to make the query more resilient, if trivial to do so, if we can't get more consistent data files based on more consisted data practices.

The only other year in which we're using this explanation text is 2016, and in that case the row was absent in the data file.

lpgoldstein:
@brentryanjohnson Thanks for taking care of that. It looks good to me.

PR #4848, opened 2019-12-05
NPM package update (for the brave)

brentryanjohnson:
[:sunglasses: PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/npm-update-maybe/)

- No content or date changes here
- I had issues with my local dev builds, so I dumped my mods, ran `npm install`, ran `npm audit fix`, and here we are.


mojobnichols:
I pulled and compiled successfully.   This mainly updates a number of the libraries that we use.   I believe your environment was just a too old version of react (or other library or node).   I don't see anything wrong with this update though and keeping used libriaries up to date is good.  Also I noted that the LTS (Long term support for node is now 12.2 not 10.16.  We run node 10.16 in circleci config this could be change to reflect that although I'm not sure it matters that much since the builds always fail here.   That said if federalist is using 10.16 then perhaps this is a good time to make federalist update its node version as well.

PR #4860, opened 2019-12-10
[WIP] Reclamation Fund, how it works

### LINKED ISSUE #4852: Write how-it-works content for Reclamation Fund
Our colleague pointed out that we lack supporting content for the Reclamation Fund, while we do have supporting content for other disbursement funds, including LWCF and HPF. We should consider developing content for the Reclamation Fund...it's a content gap.

brentryanjohnson:
Fixes #4852

[✍️ PREVIEW](https://federalist-proxy.app.cloud.gov/preview/onrr/doi-extractives-data/hiw-reclamation-fund/how-it-works/reclamation-fund/)

- Adds how-it-works content for Reclamation Fund
- Adds link and summary to how-it-works landing page

### DIFF
--- a/src/markdown/how-it-works/default.md
+++ b/src/markdown/how-it-works/default.md
@@ -150,6 +150,11 @@ redirect_from: /how-it-works/production/
             <p><custom-link to="/how-it-works/historic-preservation-fund">Learn about the Historic Preservation Fund</custom-link></p>
           </div>
           <div>
+            <h4 class="h3 landing-heading"><custom-link to="/how-it-works/reclamation-fund">Reclamation Fund</custom-link></h4>
+            <p>The Reclamation Fund receives 40% of revenue from onshore natural resource extraction. The fund supports the construction, operation, and maintenance of irrigation and hydropower projects overseen by the U.S. Bureau of Reclamation.</p>
+            <p><custom-link to="/how-it-works/historic-preservation-fund">Learn about the Reclamation Fund</custom-link></p>
+          </div>
+          <div>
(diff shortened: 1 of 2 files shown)


brentryanjohnson:
@Maroyafaied Feel free to copy this into a Word document and edit.

brentryanjohnson:
There are some specific data points in this content. I attempted to avoid them, so we could avoid the accompanying maintenance, but I found it necessary to use some data points to convey the discrepancy between revenue and appropriations, along with the balance.

If anyone has ideas about how to convey the disposition of the balance and incoming/outgoing funds _without_ curating individual data points, I welcome them!

brentryanjohnson:
☝️ The reading grade level was just over our goal at 11th grade, so I made a few gentle edits to bring it down to 10.

jennmalcolm:
Should we include bonuses and rentals in this part? 

image

Also, the original section 35 of the act specifies 52.5% to Reclamation fund (unless I'm not getting something), and it really seems to be the 1976 amendment (sec 317/page 110) that specifies 40%. Should we change the reference to section 317?

image

jennmalcolm:
Do we need the last two since they are also within the text? 
image

brentryanjohnson:
> Should we include bonuses and rentals in this part?
> 
> image
> 
> Also, the original section 35 of the act specifies 52.5% to Reclamation fund (unless I'm not getting something), and it really seems to be the 1976 amendment (sec 317/page 110) that specifies 40%. Should we change the reference to section 317?
> 
> image

I included section 35 because it appears to be where Alaska is exempted. The parenthetical is meant to reference "except in Alaska," not the revenue amount. But I can take it out if it's more confusing than helpful.

jennmalcolm:
@brentryanjohnson should we include that law in the end reference section then?

